PR #417, opened 2014-10-30
Change RGB/Hex conversions to not use eval.

naskooskov:
When using the library in an environment with strict Content Security Policy configuration calls to "eval()" are forbidden. This patch eliminates two of the existing calls to ensure custom coloring can be used in such configurations.


AlexDM0:
@naskooskov Could you update your pull so we can merge it?

Cheers!

AlexDM0:
Due to inactivity, I'll merge this in manually. Thank you.

PR #546, opened 2015-01-09
CircularImage shape

brendon1982:
This adds support for a circular image shape to network nodes. The image that is specified for a node is clipped into the shape of a circle and the normal circle shape is drawn behind it.

I also found and fixed an issue where if a node is updated its imgObject property was being set to undefined.

### DIFF
--- a/examples/network/34_circular_images.html
+++ b/examples/network/34_circular_images.html
@@ -1,0 +1,87 @@
+<!doctype html>
+<html>
+<head>
+  <title>Network | Scalable images</title>
+
+  <style type="text/css">
+    body {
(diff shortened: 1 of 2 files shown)


lib/network/Images.js
```diff
@@ -23,10 +23,11 @@ Images.prototype.setOnloadCallback = function(callback) {
  * @return {Image} img          The image object
  */
 Images.prototype.load = function(url, brokenUrl) {
```
brendon1982: Not at all, my example was the one I was using to test while I dev'ed ... and as you can see it's pretty much a modified version the scalable images example :D

Issue #450, opened 2014-11-18
Timeline Issue:  Sometimes the redraw cycle can become stuck in an infinite loop


brichbe:
I've been using the timeline product (v3.6.4 - v3.7.0) for a few weeks now and so far it's been _**great**_ for supporting my use case.  My application provides a tree structure where the user can see all the data that's available to them, and then once they select a node of that tree, that sub-set of the user's data is displayed on the timeline.

The data being displayed on the timeline is relatively simple right now -- only a few groups, background items, and range items -- that span across 2-10 days.  Each time the user chooses a new tree node to display, I populate the timeline with that new sub-set of data by completely re-creating my item and group `DataSets`.  This is probably not the most elegant way to re-populate the timeline with different data (as opposed to something like a filter), but for my use case it's been working perfectly fine the majority of the time.

Unfortunately, the browser will sometimes completely freeze after giving a new sub-set of the data to the timeline for display.  The freeze is caused by an infinite (or very long running) loop in the redraw cycle of the various timeline components.

When I give my user's data sub-set to the timeline, I do so in a few distinct back-to-back steps:
- First I create the necessary group(s) in a new `DataSet`
- Next I create the necessary background and range item(s) in another new `DataSet`
- I set these two `DataSets` into the timeline
- Finally I set the timeline window to span the entire range of those background/range items
- (I really don't believe there are any issues with these steps, because like I said above, most of the time this all works perfectly fine)

After each of those steps occur, the timeline redraws itself via the Core redraw function
- This Core function causes each of the timeline's top-level components to also redraw themselves as necessary
- This Core function also ends up calling itself again if it's determined that one of the components was resized

Obviously, the Core redraw function calling itself should behave in a safe recursive manner so that sequence of calls stops relatively quickly and efficiently.  Indeed, it normally does stop after a few redraw cycles.  

However, I've found there are also times when that redraw cycle does not stop as it should.  When this infinite looping issue occurs, so far I've been able to trace that the following is happening:
- Each and every time the Core redraw function triggers a redraw of its components, the `TimeAxis` and `ItemSet` components continuously return true (identifying they've been resized and therefore the entire redraw cycle should repeat)
- For the `TimeAxis` I've found the reason it always returns true is because it looks to see if its current width/height match the previous width/height
  - The `props.width` value is set from the `frame.offsetWidth`
[...]

PR #559, opened 2015-01-13
No longer throw Error when MAX_REDRAWS count has been exceeded.

brichbe:
Throwing this error was introduced recently, but appears to have been unintentionally committed (since the error message isn't helpful/relevant, and because the unexpected Error will unnecessarily disrupt the display of the timeline if not caught).

### DIFF
--- a/lib/timeline/Core.js
+++ b/lib/timeline/Core.js
@@ -614,8 +614,7 @@ Core.prototype.redraw = function() {
       this.redraw();
     }
     else {
-      console.log('WARNING: infinite loop in redraw?')
-      throw new Error("bla")
+      console.log('WARNING: infinite loop in redraw?');
     }
     this.redrawCount = 0;
   }


AlexDM0:
Could you share the code that made this required? We're trying to find the bug that causes big redraws. Until it is found I'd rather not remove this.

brichbe:
I still don't have precise code for reproducing the redraw issue.  As I wrote in issue #450 , I cannot reproduce it consistently and therefore cannot trace the issue to one particular type of data, nor any particular component size/style.

The reason I'm proposing this change is because in the discussion of #450 we seemed to agree that detecting the condition, stopping the loop, and logging to the console was a reasonable intermediate fix until the true cause/solution is found.

Then a change was recently introduced to *also* throw an error in this case.  I'm not aware of any benefit in throwing the error now in addition to what was being done before.  Instead I find the thrown error to introduce yet another problem because it causes the timeline refresh/redraw to halt unexpectedly under those problematic circumstances.

AlexDM0:
Hi,

My apologies, it seems to have slipped through. Thanks for reminding us of this. I'll accept the merge.

Cheers!

PR #560, opened 2015-01-13
Change _getNodeAt/_getEdgeAt to return the first matched object

pavlos256:
There's a small issue on a busy network with many edges. If hover is enabled and the edges have titles, hover is picking one edge while the popup is shown for another edge.

The reason is that the `_getNodeAt()` and `_getEdgeAt()` methods return the *last* of all the matches objects, while the code in the `_checkShowPopup()` method picks the first match.

This changes `_getNodeAt()` and `_getEdgeAt()` to return the first match.


AlexDM0:
Hi,

I would actually argue that the Popup should be changed to the last index. As stated in the comment above the line you altered, the last one is the one on top. 

Additionally, we cannot accept merges to the master branch, could you change this and merge to develop?

Regards,

Alex

pavlos256:
OK, will do, thanks!

PR #562, opened 2015-01-13
Select the last (top) overlapping object when showing the popup

pavlos256:
Follow up to #560.

When there are many overlapping objects, pick the last one which
is drawn at the top of the others. This is the same way the
_getNodeAt() / _getEdgeAt() methods of the SelectionMixin work.


AlexDM0:
Hi Pavlos,

Thank you for making these changes!

Regards,

Alex

PR #566, opened 2015-01-14
Added Label Alignment

T-rav:
I have added a feature to allow the user to specify how labels on edges behave.
This new behavior aligns the label to the line, either above or on it.
There is a new configuration property labelAlignment for edges. 
It has two values : line-center or line-above.
If no value is given the current behavior is used.

I have added a working example 35_label_alignment.html showing all three cases of labels on edges.


AlexDM0:
Hi Trav,

Could you add documentation for this?

Regards,

Alex

AlexDM0:
I'm not sure I can tell what exactly is new in the example you provided.. 
http://htmlpreview.github.io/?https://github.com/T-rav/vis/blob/master/examples/network/35_label_alignment.html

The orientation of the labels is still absolute horizontal

T-rav:
Looks like a merge issue. I do not see my work. Will check it out.

AlexDM0:
Hi T-rav,

Could you also simplify the example? Since this is about labels, we do not need the (rounded) images. We have another example for that. Just the same node setup as the very basic example 1, but with edge labels in all alignments would be perfect.

Regards,

Alex

T-rav:
Hi Alex,

I have made the example simpler as you requested.
I have also committed the dist/vs.js and dist/vs.min.js files. 
The example was not working due to the fact these where not committed.

Cheers,
Trav

AlexDM0:
Hi Trav,

Could you merge it with the latest develop? There have been changes to the code you're using lately so I'd rather you merge with that beforehand. Could you add documentation? Finally, it seems if you're offering line-center and line-above, that line-below would be a natural addition. Is that possible?

Regards,

Alex

T-rav:
I will merge with the latest develop.
And then add a line-below option as well. 
As for documentation, where do I need to add it?

AlexDM0:
Hi Trav,

There's an HTML file in /docs/network.html. Those are the docs we use on the website. If you could add the option there in the appropriate place that would be great! 

Your picture seems to be orphaned now that you changed the example. Could you remove it from the merge? We have had some problems with merging when the dist folder was also in the merge. The problem is that if we merge manually, git does not always pick up on that the pull req was performed. I'd like it to show as accepted if possible, to your credit. If the dist folder is not present, everything is a bit easier :)

Thank you very much for the contribution!

Regards,

Alex

T-rav:
Hi Alex,

I have merged from your develop branch.
I have removed the picture of me.
I have also added a line-below option and included it in the example.
And I have updated the network.html docs to reflect this option in Edge configuration.

Cheers,
Trav

lib/network/Edge.js
```diff
@@ -580,33 +583,122 @@ Edge.prototype._label = function (ctx, text, x, y) {
       this.labelDimensions = {top:top,left:left,width:width,height:height,yLine:yLine};
     }
 
+	var labelAlignment = this.options.labelAlignment;
+	var yLine = this.labelDimensions.yLine;
+	
+	ctx.save();
+	
+<<<<<<< HEAD
```
AlexDM0: this has not merged correctly it seems...

lib/network/Edge.js
```diff
@@ -580,33 +583,122 @@ Edge.prototype._label = function (ctx, text, x, y) {
...
+ * Draws the label rectangle 
+ * @param {CanvasRenderingContext2D} ctx
+ * @param {String} labelAlignment
+ * @private
+ */
+Edge.prototype._drawLabelRect = function(ctx, labelAlignment) {
+	if (this.options.fontFill !== undefined && this.options.fontFill !== null && this.options.fontFill !== "none") {
+		ctx.fillStyle = this.options.fontFill;
+		
+<<<<<<< HEAD
```
AlexDM0: merge issues

AlexDM0:
Hi Trav,

Sorry to make this so complicated.. It seems something is wrong with the merge you performed. I added some notes to it. Could you check it?

Regards,

Alex

lib/network/Edge.js
```diff
@@ -580,33 +583,122 @@ Edge.prototype._label = function (ctx, text, x, y) {
...
+ * @private
+ */
+Edge.prototype._drawLabelText = function(ctx, x, yLine, lines, lineCount, fontSize, labelAlignment) {
+
+	// draw text
+	ctx.fillStyle = this.options.fontColor || "black";
+	ctx.textAlign = "center";
+	if(labelAlignment == 'line-above'){
+		ctx.textBaseline =  "bottom";
+<<<<<<< HEAD
```
AlexDM0: additional merge issues

T-rav:
Should be fixed now. Sorry about that.

AlexDM0:
Hi Travis,

It does not seem possible to use this with the smoothCurves which we have enabled by default. I'll attempt a merge and see what we get :).

Regards,

Alex

T-rav:
I am a bit confused about the smoothCurves issue....
If smoothCurves is on by default it should have caused issues when I developed the feature.
When moving a node edge will bend and curve but the text alignment follows a sensible flow.
Let me know if you need anything else done after the merge.

AlexDM0:
Hi Travis,

It seems to work, the smoothCurves seem to sort themselves out. My bad. I'm tweaking it a little to be more stable across browsers. Specifically: http://www.w3schools.com/tags/canvas_textbaseline.asp

the different browsers leave more open to interpretation using top and bottom, so I prefer alphabetic and hanging.

I've made the default entree be 'horizontal' as instead of undefined. I'd prefer all options to be defined for clarity.

It seems the labelStrokes have been broken. I'll fix them.

I have merged everything is it's working. Thank you for your contribution!

Regards,

Alex

T-rav:
Thanks for the patch ups and merging the feature in.

PR #572, opened 2015-01-15
More work on circular images

brendon1982:
Did a bit more work on circular images.

- Updated the circularImage node shape to render a circle when images are loading or have failed to load. Previously it would render nothing at all in these situations.
- Fixed a bug where when the image and brokenImage failed to load it would infinitely try load the brokenImage.
- Fixed an issue where multi line labels for image and circularImage nodes were overlapping the image.


lib/network/Images.js
```diff
@@ -52,6 +60,13 @@ Images.prototype.load = function(url, brokenUrl) {
           me.callback(this);
         }
       }
+      else if (this.src === me._resolveRelativeUrl(brokenUrl)){
```
AlexDM0: I dont really agree with this method.. But the problem is relevant. I'll push the fix below to develop:

```
else if (me.imageBroken[url] === true) {
        console.error("Could not load brokenImage:", brokenUrl);
        delete this.src;
        if (me.callback) {
          me.callback(this);
        }
      }
      else {
        this.src = brokenUrl;
        me.imageBroken[url] = true;
      }
```

lib/network/Images.js
```diff
@@ -7,6 +7,14 @@ function Images() {
   this.callback = undefined;
 }
 
+Images.prototype._resolveRelativeUrl = function(url){
```
AlexDM0: read my comment below: specifically this part I don't think is required.

AlexDM0:
Hi Brendon,

Could you remove the images.js file from your commit? See my comments for explaination. I'll merge it in after that. It will be just in time for the release today :). Thanks for the contribution!

Regards,

Alex

brendon1982:
Hi Alex,

Thanks, I also prefer your way of dealing with it :D 

I've reverted the changes to the Images.js

AlexDM0:
Thanks!

AlexDM0:
Hi Brendon,

It's not rendering a circle now when it has no image. I'll fix it.

Regards,

Alex

AlexDM0:
Sorry, didn't rebuild. 

Cheers!

brendon1982:
:D

PR #625, opened 2015-02-10
Interactions with minor labels

y-lohse:
Hi, I'm not sure you're interested in this, but since we needed it… :-)

This commit adds touch/click events on **minor** labels. The use case for this being recentering the view on the clicked label or maybe zooming into it (ie. clicking on "Sun. 23" shows the whole day).

The implementation mirrors the way Item selection is designed so I don't think I went too much astray there; The actual changes are fairly little, so not sure why the dif on the minified files is so large. The event contains a single property called `label` which is a Date object.

It'd be quick to implement that on major labels too althought I'm not sure if there is a use case.

Oh, and the example is a bit messy right now but it tests support for both orientations & on the fly orientation change.


josdejong:
Thanks for your PR @y-lohse . The timeline still lacks support for events like click, double-click, hover, etc. So you can add your own behavior. Clicking a label is one of these customizations. 

I haven't yet thought about a clear API for these event listeners. An API like `.on('label', ...)` may be limiting if we would want to support say double-clicking on a label too. Maybe we could have a generic `.on('click', callback)`, which provides the clicked thing as argument, that could be an item, label, group, or empty space.

y-lohse:
You're right, this approach is very limited and the one you described seems to be closer to what's going on in the rest of the vis.js. I'll close this PR, the code is still there if someone needs a headstart for the labels at some point :-)

PR #638, opened 2015-02-16
Add ability to use icon fonts for nodes

Dude9177:
Add the ability to use icon fonts for nodes such as FontAwesome

# Example
Define nodes to show users and usergroups

``` javascript
var nodes = [
        {
            id: 1,
            label: 'User',
            shape: 'icon',
            iconFontFace: 'FontAwesome',
            icon: '\uf007', // unicode for the 'user' icon from FontAwesome
            iconSize: 50,
            iconColor: '#1ba1e2'
        },
        {
            id: 2, 
            label: 'Usergroup', 
            shape: 'icon',
            iconFontFace: 'FontAwesome',
            icon: '\uf0c0', // unicode for the 'users' icon from FontAwesome
            iconSize: 50,
            iconColor: '#009966'
        }
    ];
´´´

### DIFF
--- a/docs/network.html
+++ b/docs/network.html
@@ -1009,7 +1009,6 @@ mySize = minSize + diff * scale;</pre>
         <td>'white'</td>
         <td>The color of the label stroke.</td>
     </tr>
-
 <tr>
     <td  class="greenField">shape</td>
     <td>string</td>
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

**Comment by AlexDM0** [comment]
Hi,

Thanks for your contribution! Could you also add an example and documentation?

Regards

**Comment by Dude9177** [comment]
I've added examples.

Greetings

**Comment by AlexDM0** [comment]
Thanks for the contribution!

**Comment by AlexDM0** [comment]
Err. I just checked the example and I dont seen anything other than the default nodes.. did I do something wrong?

**Comment by AlexDM0** [comment]
Nevermind, its Monday. I forgot to rebuild.

**Comment by AlexDM0** [comment]
Hi,

It seems the node is very hard to actually click due to this part:

```
Node.prototype._resizeIcon = function (ctx) {
  if (!this.width) {
    var margin = 5;
    var textSize =
    {
      width: 1,
      height: Number(this.options.iconSize) + 4
    };
    this.width = textSize.width + 2 * margin;
    this.height = textSize.height + 2 * margin;
```

The width is very small compared to the height (11 vs 64 in your example). The icons itself look like they are roughly square though. Could you explain why the width is like this?

Cheers

Dude9177:
Sorry, it was a mistake.

I made a new pull request #641 which should fix this problem.

Greetings

PR #641, opened 2015-02-17
Fixed the width problem for icons wich made it difficult to click on them

Dude9177:
Fixed the problem with the width of an icon that made it difficult to click on it because the width was very small compared to the height.

### DIFF
--- a/lib/network/Node.js
+++ b/lib/network/Node.js
@@ -1015,19 +1015,19 @@ Node.prototype._drawText = function (ctx) {
 Node.prototype._resizeIcon = function (ctx) {
   if (!this.width) {
     var margin = 5;
-    var textSize =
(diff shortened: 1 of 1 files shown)


AlexDM0:
Hi, 

Could you explain the + 4 in the height? I would like the magic numbers to be commented or in a named variable because (as I've encountered in my own code numerous times) I forget what they mean and why they are there. 

Thanks for the quick reply!

Dude9177:
Hi,

I used this +4 to make a spacing between the icon and the label - but as you found out this was the wrong place and not well done too :)

I commited a maybe slightly better solution for the spacing.

Greetings.

AlexDM0:
Great! Thanks!

PR #646, opened 2015-02-19
Movable Current Time bar

aytech:
Hello,
I'm currently working on an enterprise project where one of the requirements is the timeline implementation. Since vis.js allows for high degree of customization, we agreed to try to implement it into our environment. 
However, since the requirements for the timeline itself reach beyond the current vis.js capabilities, I would like to join the project and offer my contribution to it.
Current requirement is the ability to dynamically change the current time setting, i.e. move the current time vertical bar the same way the custom bar can be moved. So the list of my changes is as follows: 
1. Feature is enabled via moveCurrentTime attribute, which defaults to false.
2. If enabled, the draggable current time vertical bar is added asynchronously (since CurrentTime setOptions method is asynchronous and options are constructed after the _create method is called)
3. Dynamic update of the current time bar (as time progresses) is paused during drag, resumed on release.
4. Added "cursor: move" CSS style to the moveable current time bar (disabled if moveCurrentTime is false);
5. Added Hammer instance to the CurrentTime object

Overall, the functionality was mostly copied from CustomTime object and adjusted to the CurrentTime object with as minimal code change as possible.

Thank you,
Oleg

### DIFF
--- a/dist/vis.js
+++ b/dist/vis.js
@@ -19325,6 +19325,7 @@ return /******/ (function(modules) { // webpackBootstrap
 /* 39 */
 /***/ function(module, exports, __webpack_require__) {
 
+  var Hammer = __webpack_require__(19);
   var util = __webpack_require__(1);
   var Component = __webpack_require__(23);
   var moment = __webpack_require__(2);
@@ -19339,11 +19340,13 @@ return /******/ (function(modules) { // webpackBootstrap
    * @extends Component
    */
   function CurrentTime (body, options) {
+
     this.body = body;
 
     // default options
     this.defaultOptions = {
       showCurrentTime: true,
+      moveCurrentTime: false,
 
       locales: locales,
       locale: 'en'
@@ -19352,6 +19355,8 @@ return /******/ (function(modules) { // webpackBootstrap
     this.offset = 0;
 
     this._create();
+    this.currentTime = new Date();
+    this.eventParams = {};
 
     this.setOptions(options);
   }
@@ -19372,6 +19377,26 @@ return /******/ (function(modules) { // webpackBootstrap
     this.bar = bar;
   };
 
+  CurrentTime.prototype._enableMove = function() {
+    var drag = document.createElement('div');
+    drag.style.position = 'relative';
+    drag.style.top = '0px';
+    drag.style.left = '-10px';
+    drag.style.height = '100%';
+    drag.style.width = '20px';
+    drag.style.cursor = 'move';
+    
+    this.bar.appendChild(drag);
+
+    // attach event listeners
+    this.hammer = Hammer(this.bar, {
+      prevent_default: true
+    });
+    this.hammer.on('dragstart', this._onDragStart.bind(this));
+    this.hammer.on('drag', this._onDrag.bind(this));
+    this.hammer.on('dragend', this._onDragEnd.bind(this));
+  }
+
   /**
    * Destroy the CurrentTime bar
    */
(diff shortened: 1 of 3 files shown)


AlexDM0:
Hi Oleg,

First of all, thank you for contributing! What I don't quite understand is: why would you want to move the currentTime bar? Why not just hide it and use the customTime bar? Would it not be easier to update the customTime with an increment?

I'd be really interested in hearing the usecase :)

Regards,

Alex

aytech:
Hi Alex,
Well, our use case is like this: we have a dynamic map of nodes that is listening to the timeline events and updates in near real-time. And this map supports view for time ranges, say you have a list of nodes that were created during a certain time period and if you set the range on the timeline, then only these nodes will be shown. 
So, in order for this to work correctly, we need 2 time stamps, one for the start, another for the end of the time range, and both have to be draggable. I was thinking at first to implement the optional second customTime bar, but it looked to me that enabling the move of the currentTime bar will require less change in the codebase. 

This is one of the requirements that, in order to meet, I need to hack into the core of the library, and I thought that this might help others who would have similar requirements.

And, of course, I'd love to hear your feedback on the changes or suggestions on how this could be implemented better

With regards,
Oleg

AlexDM0:
Hi Oleg,

I think it would be better to go with adding a method to insert and remove (draggable) customTime markers. These can already be moved to match the current time using the method. I don't think there is a big difference in implementation difficulty but then again, I might be biased :).

I discussed with @josdejong and we think that, while functional, your approach is not in line with the way we would like to do this. This means that we will not merge it into the codebase as it stands now. 

We feel the time bar is really ment as a indication of the current time. The solution we would opt for is to generalize the customTime bar and have methods to add/remove/style them.

Now naturally we don't expect you to redo your work just because we feel a different approach would be better suited for vis.js and if your solution works for you then that's great :) It why the project is opensource in the first place. 

If you'd like we can treat this as a feature request? If you are feeling up to changing your implementation, we'd happily merge it in.

Regards,

Alex

aytech:
Hi Alex,
I'm a bit constrained with a deadline, so I guess my best bet is to stick to my implementation, for now.
That being said, I believe your solution is great and makes more sense, and I would love to work on that. I feel like I should have discussed that before I started working on my implementation, so it's my bad, but I wasn't even sure I can pull off an update to such an extensive library. 

To summarize, let's keep this as a feature request, and I'll update my implementation to fit your description. To make it more clear, you think it'll be OK if an option to create second customTime bar will be introduced, say, as an enableCustomRange boolean switch ? If enabled, 2 custom bars will appear, separated by some predefined time period. 

Regards,
Oleg

AlexDM0:
Hi Oleg,

I'd prefer a method like addCustomTimeBar(time) which will return an ID (like 0 for default, 1, 2, etc). The new customTime bar will spawn at the time given to the constructor. You would then supply the ID to all functions regarding the customTime bar (do we have more than 1?) with a default ID to 0 so if you do not supply one, the default bar is moved. Finally a removeCustomTimeBar(barId) that will only delete the user added ones. If you want to delete all of them, just turn the visibility off in the options.

Would it be OK for you to make a new pull req once you've finished a new implementation? For questions you can always post an issue or ask here.

Regards,

Alex

AlexDM0:
To add, your enthusiasm is greatly appreciated :) thanks for wanting to help out!

aytech:
Alex,
Got it, will do as you suggest. 
You're welcome, its quite exciting for me to work on such a project, so thank you for guidance.

Regards,
Oleg

PR #649, opened 2015-02-20
Option to add custom vertical bars implemented

aytech:
Hello,
As discussed in #646, I've added the option to spawn more than one custom vertical bars. Each new bar gets an id, and this id is emitted together with time during timechange and timechanged events. 
Also I've added an option to remove the custom vertical bars. 
Example added to 34_add_custom_timebar.html

Regards,
Oleg

PR #651, opened 2015-02-21
Multiple customTime vertical bars

aytech:
Hello,
This is another take on the correct feature implementation, previous being #649, where I did it all wrong ...
The update is simple: enable a user to add multiple customTime bars, each assigned with a unique ID. This ID is returned by the method, also the ID is added to the object emitted on timechange and timechanged events. 
Also the method to remove these custom lines was added. 
This commit includes update to the documentation

Regards,
Oleg

### DIFF
--- a/docs/timeline.html
+++ b/docs/timeline.html
@@ -910,9 +910,9 @@ timeline.clear({options: true}); // clear options only
   </tr>
 
   <tr>
-    <td>getCustomTime()</td>
+    <td>getCustomTime([id])</td>
     <td>Date</td>
-    <td>Retrieve the custom time. Only applicable when the option <code>showCustomTime</code> is true.
+    <td>Retrieve the custom time. Only applicable when the option <code>showCustomTime</code> is true. If parameter <code>id</code> is provided, time of the custom time bar under that ID is returned.
     </td>
   </tr>
 
@@ -966,6 +966,14 @@ timeline.clear({options: true}); // clear options only
   </tr>
 
   <tr>
+    <td>removeCustomTime(id)</td>
+    <td>none</td>
+    <td>
+      Remove vertical bars previously added to the timeline via <code>addCustomTime</code> method. Parameter <code>id</code> is the ID of the custom vertical bar returned by <code>addCustomTime</code> method.
+    </td>
+  </tr>
+
+  <tr>
     <td>setCurrentTime(time)</td>
     <td>none</td>
     <td>Set a current time. This can be used for example to ensure that a client's time is synchronized with a shared server time.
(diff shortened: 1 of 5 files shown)


AlexDM0:
Hi Oleg,

I have a few points of comment (yes, again :)). You did not document the removeCustomTime function and I don't think you've updated the setCustomTime function to be able to change the customTime afterwards or the getCustomTime function to be able to supply an ID. All those functions need to be included and their documentation extended. If no ID is supplied, the default ID (0?) should be assumed.

Thanks for the effort :)

Regards

aytech:
Hi Alex,
You're right, I completely forgot about the utility functions, that they should be adapted to the new behavior. Will do that, and documentation too :)
If I'll have more than one commit, and you decide to merge, can all these commits be merged under one pull request? 
Sorry for the newby questions, I'm kinda new to contributing on GitHub... :)

Oleg

AlexDM0:
Hi Oleg,

Yes you can stack the commits. This is automatic if you commit to your develop branch. 

Cheers

aytech:
Alex,
I've pushed the updates, please see if it is OK now.

Thanks,
Oleg

AlexDM0:
Hi Oleg,

Looks good! I have just one more thing, could the user supply their manually chosen ID when creating? I could imagine it would be useful if you could name them something like "triggerTime" or whatever to use in an application. (Downside is that we have to check for duplicates)

Let me know what you think!

Cheers

aytech:
Alex,
Why not, it will probably make more sense than just a bunch of numeric IDs. I'll have to redo the id tracking and generation, since I filter it as number in couple places. Good catch, update is on the way :)

P.S. Strange thing happened to me, looks like if you compile the sources on Windows, the custom bar doesn't work at all. I checked out the nearest commit before my updates to make sure I didn't break anything, same thing. If I compile on Debian or Mac, everything works as expected. Is it a known thing?

Oleg

AlexDM0:
> Hi Oleg,

I've always compiled on windows so as far as I know it works. That said, the new gulp seems to skip from time to time. Try restarting the npm run watch-dev 

Regards

aytech:
Alex,
Done, the ID for custom bar can be passed to the method now. So addCustomTime can now return either Number or String, depending on what was passed. If nothing was passed, the ID is auto generated, and is returned as Number. 
I keep all the IDs in one array and check for duplicates when a new bar is being added. If a duplicate is found, the Error is thrown. I've updated the docs too.

Thanks,
Oleg

aytech:
I've tried the compile on a different Windows machine and it works OK, probably there was something wrong with the setup :) I don't see the watch-dev task in gulpfile...

Oleg

AlexDM0:
Hi Oleg, 

It's hidden in the package.json, it only bundles, not minifies and is much faster :). I'll merge the request, looks good! Thanks for all the work :)

Regards,

Alex

aytech:
Hi Alex,
Thanks for the tip, bundling on watch is definitely slow and is frustrating when fixing small things :) Will use the dev from now on. 
Great news that it will be merged, I have a couple more changes in mind, I'll go the official way this time and create a feature request. 

Thanks for your support, this way of adding custom lines already proved to be exactly what we need in our environment, amazing that you had a better vision even though you haven't seen our app :) 

Thanks!
Oleg

PR #678, opened 2015-03-02
Added possibility to rename fields on a DataSet.get()

spatialillusions:
Added functionality to extend the fields array, to a fields object, similar to the type declaration. This makes it possible rename properties when you get them from a dataset or data view.

This would make it possible to easily plot a Timeline dataset to a Graph2d and the other way around. You can as before set the options to {fields:["start","value"]}, but it would also be possible to set it to {fields:[start:"x",value:"y"]} to rename start to x and value to y. What do you think?

### DIFF
--- a/docs/dataset.html
+++ b/docs/dataset.html
@@ -740,9 +740,10 @@ DataSet.map(callback [, options]);
 
   <tr>
     <td>fields</td>
-    <td>String[&nbsp;]</td>
+    <td>String[&nbsp;] | Object.&lt;String,&nbsp;String&gt;</td>
     <td>
-      An array with field names.
+      An array with field names, or an object with current field name and 
+      new field name that the field is returned as.
       By default, all properties of the items are emitted.
       When <code>fields</code> is defined, only the properties
       whose name is specified in <code>fields</code> will be included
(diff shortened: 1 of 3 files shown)


spatialillusions:
Another use case would be that you have a dataset with several different value fields, and you want to plot the dataset to several different graphs, then you could simply create multiple views of the data. (Of course it would be nice if you could create several views and plot them to the same graph, but then you would have to add a group id to the different views in some way and merge them to a single dataset, and I can't see a simple way to do that.)

josdejong:
Thanks, nice idea to be able to transform the field names. A few thoughts:

- we should implement the same mapping functionality for DataViews, with the same API. 
- maybe for performance you could do the check whether `fields` is an Array or Object outside of the for loop (and thus write the for loop two times)?
- practical points for your PR:
  - can you please do the PR to the `develop` branch?
  - can you please update the relevant documentation with this new feature?
  - can you use 2-space indentation?

spatialillusions:
Hi, 

I will change so that the check is outside of the for loop, good idea, change to two space indentation and update the documentation. Then I'll do a new PR to the develop branch. 
DataViews actually already works, since they internally calls the same function.

josdejong:
> DataViews actually already works, since they internally calls the same function.

ah indeed, even better :)

thanks beforehand!

spatialillusions:
Added a new PR with updated code and docs.

PR #686, opened 2015-03-04
Added possibility to rename fields on a DataSet.get() or in a DataView.

spatialillusions:
Adds the possibility to rename fields when you get your DataSet or creates a DataView. Updated code and docs.


josdejong:
Thanks Måns!

PR #708, opened 2015-03-11
Fixes bug where new Timeline Items are created in the wrong group...

markusos:
...when client is scrolled.

The ItemSet.prototype.groupFromTarget function did not take into
account the if the client view was scrolled when calculating where the
user clicks.

How to reproduce:
- Open http://visjs.org/examples/timeline/09_order_groups.html
- Add a couple of events in the first group at the same time, make sure they stack.
- Resize the window so that only group one is visible
- Scroll down to the third group and try to add an item.
- The Item will be added to either the first or the second group, depending on your scroll offset.

### DIFF
--- a/lib/timeline/component/ItemSet.js
+++ b/lib/timeline/component/ItemSet.js
@@ -1599,6 +1599,8 @@ ItemSet.prototype.itemFromTarget = function(event) {
  */
 ItemSet.prototype.groupFromTarget = function(event) {
   var clientY = event.gesture ? event.gesture.center.clientY : event.clientY;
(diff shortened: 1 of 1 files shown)


josdejong:
Thanks, good find. I think this is a broader issue, let me have a thorough look at this.

josdejong:
I found the underlying issue: usage of pageY instead of clientY. Using pageY we can safely use the position in combination with `util.getAbsoluteTop` and `util.getAbsoluteLeft`.

josdejong:
I've pushed a fix to the `develop` branch, replacing usage of clientY with pageY and the same with clientX.

Issue #331, opened 2014-09-18
Mouse events on the timeline


Travellerme:
Hello.

I could paste events with html tags and add some event listeners for this html elements.
But now when I tried to add event listeners to icon on the timeline (for example:" $(document).on('click', '.myClass',function); "), this would not work.
In firefox I can add "mouseup" and "mousedown" events, but in IE 10 it will not work.
I have tried to use "clickToUse" option. I thought that it would enable mouse events on the timeline area, but it didn't work.

Thank You.

PR #802, opened 2015-04-24
Fix Problem with click events on timeline items

toxix:
The click events for timeline items do not fire. (Seems to
be catched by Hammer) So manualy call the click event on the item
wich had been clicked.

Seems related to https://github.com/almende/vis/issues/331

Not sure if this is the best solution or bring other unwanted side effects.

### DIFF
--- a/lib/timeline/component/ItemSet.js
+++ b/lib/timeline/component/ItemSet.js
@@ -1371,6 +1371,7 @@ ItemSet.prototype._onDragEnd = function (event) {
  * @private
  */
 ItemSet.prototype._onSelectItem = function (event) {
+  event.target.click();
   if (!this.options.selectable) return;
 
   var ctrlKey  = event.gesture.srcEvent && event.gesture.srcEvent.ctrlKey;


josdejong:
Thanks. This solution feels a bit hacky though. The underlying cause has to do with the Hammer instances being configured with `preventDefault: true`, which was needed to solve some other issues in the past. Clicking buttons and links works fine on desktop but not on touch devices.

Note that this issue has been solved in the `v4` branch, it may be ok to just wait for the release of v4?

toxix:
Seen the comment at Hammer, so decided to make the hack like this. But if you find better solutions in ```v4``` that's also great.

> Clicking buttons and links works fine on desktop but not on touch devices.

html links work fine. But bounded click events do not fire even on the desktops (crome and firefox on ubuntu).

> Note that this issue has been solved in the v4 branch, it may be ok to just wait for the release of v4?

Okay, will try ```v4``` branch. Is this branch save to update or would you advise to use the patched ```master``` until the release of ```v4```? (I'm currently only using the dataset and timeline)

josdejong:
v4 is not yet stable but will be released within one or two weeks.

toxix:
Just done a quick test on ```v4```. Binding click events seems to work for me in a quick test. So waiting for the release. Seems like there are some nice changes, and some work to do on ```v4```. So lets close this issue/pull for now.

AlexDM0:
Hi, 

If you encountered bugs in the v4 branch, could you make an issue? As far as I know it should be relatively stable by now. 

Most of the remaining work should be docs. 

Cheers!

frittola:
it's very strange, but i work with last release and this problem hasn't been fixed yet.
maybe I did something wrong.

I use a template html into an item but when i click on custom button it don't work

PR #832, opened 2015-05-13
Added the ability to customize the format of the Data axis labels

ewjoachim:
This pull requests would allow passing a function in the options that would let the developer chose their preferd way of displaying the data (mainly if they want greater control on the numerical display, or add an unit or anything)

(and thanks @SmartImpulse for sponsoring this commit)

### DIFF
--- a/docs/graph2d.html
+++ b/docs/graph2d.html
@@ -506,7 +506,7 @@ The options colored in green can also be used as options for the groups. All opt
   <td>undefined</td>
   <td>Set the number of decimal points used on the the left axis. If set, this will fix the number of decimal places
     displayed after the decimal point.
-    If undefined, trailing zeros will be removed.</td>
+    If undefined, trailing zeros will be removed. If dataAxis.format.left.formatter is defined, decimals will be unused.</td>
 </tr>
 <tr>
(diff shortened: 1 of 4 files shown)


AlexDM0:
Hi,

Thanks for the contribution! Thing is, I've actually also added this to the v4 branch last week. We plan to launch v4 next week. 

From your documentation entree I see that it's exactly the same idea, input in, string out. This function completely replaces the decimals input we had before. 

Nevertheless, thank you for the effort. 

Cheers!

ewjoachim:
... Well I'll try to coordinate better next time. But then I don't need to be the one who did it, as long as it's implemented, so if it means it will be out sooner, I'm glad anyway ! And then I had the opportunity to dive into vis' source and process, so all the better :) Thanks for your quick response !

(PS : if I missed something and I had a way to know it was already being taken care of, please let me know how, so I'll know for next time !)

AlexDM0:
Hi Joachim,

It has been quite hectic with the v4 development since we changed a lot of things. You couldn't have seen what was being worked on. 

For next time, you could first make an issue with a feature request and mention that you would like to implement it yourself. We can then help out by pointing you to the pieces of code that are relevant. 

Regards,

Alex

PR #856, opened 2015-05-23
fixed colors for selected and hover states

killerDJO:
Configuration options for highlighted and hovered edges didn't work. This should fix the issue.


AlexDM0:
good catch! thanks!

PR #906, opened 2015-06-01
Ignore folders unneeded by Bower

btmills:
We're using vis.js via Bower as part of a Django project. [Django's staticfiles app](https://docs.djangoproject.com/en/1.8/ref/contrib/staticfiles/#collectstatic) was erroring out because it was unable to follow some `url` links within `docs`, `examples`, and `lib`. Hopefully these are actually unnecessary and can be removed entirely from the Bower distribution. For now I've referenced my fork as a git dependency in `bower.json` and things are working fine, but that requires Bower to clone the entire large repository. We'd much rather reference the standard distribution if this change is accepted!


josdejong:
Yes these folders are important, if you install `vis` it gives you the examples and docs locally, so you can easily get up and running in a local environment.

Looking at the docs you refered to it looks like you can ignore patterns of files using `--ignore <pattern>`, so you could just ignore these folders `vis/docs`, `vis/examples` and `vis/lib` in you Django project, or maybe you can ignore all `*.html` files in assets like vis.js.

btmills:
That rationale makes complete sense, and I somewhat expected that would be the case. I'd like to avoid special-casing this command every time it's run, so I've modified the PR as follows:

- You mentioned wanting to include both docs and examples in the distribution. `lib` is now the only folder ignored.
- `bootstrap.css` was referencing a webfont that wasn't included in the docs. I've copied Bootstrap's `fonts` directory into `docs`.
- `examples.css` included a link to an image inside `images`, which I've corrected to match the actual name, `img`.

I hope this change is more acceptable, and thank you for taking the time to look at this PR - I'm well aware that it provides very little direct benefit to vis.js itself aside from a trivial fix to a broken background image link.

josdejong:
Thanks. Indeed `./lib` isn't really necessary. Thanks for adding the missing bootstrap fonts. As for the `examples.css` file: I see it's not in use anymore so it can be completely removed, I will do that.

But does leaving out `./lib` this solve your issue then?

btmills:
> But does leaving out ./lib this solve your issue then?

Yep! The image links in the CSS couldn't be followed from the development sources in `./lib` because they're relative to the concatenated CSS in `./dist`. The other option would've been to set up [URL rebasing](https://github.com/jakubpawlowicz/clean-css#how-to-rebase-relative-image-urls) in the CSS minifier so the links would be valid in both places, but I considered that overkill because `./lib` isn't necessary in the distribution anyway.

Is there somewhere I can watch (a mailing list, issue, etc.) to be notified on the next release so I can update our `bower.json` from my fork to the official distribution?

josdejong:
Ah indeed.

There is no mailing list or something like that, but maybe you can regularly run a tool like [bower-update](https://www.npmjs.com/package/bower-update) for your project.

PR #961, opened 2015-06-11
Remove moot `version` property from bower.json

kkirsche:
Per bower/bower.json-spec@a325da3

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -1,6 +1,5 @@
 {
   "name": "vis",
-  "version": "4.2.0",
   "main": ["dist/vis.min.js", "dist/vis.min.css"],
   "description": "A dynamic, browser-based visualization library.",
   "homepage": "http://visjs.org/",


josdejong:
Thanks. I understand that the version number is not used by bower, but I think it's a bad advice to leave the version number out. I think it's important to be able to (easily and irrefutable) find the actual version number of the library you have actually on your disk.

kkirsche:
Considering it may be moving to NPM anyway I disagree with you pretty strongly. bower/bower#1520 is just one of the discussions. Recommending users even use bower is just bad IMO

kkirsche:
Also their maintainer says they probably won't ever use it: http://stackoverflow.com/questions/24844901/bowers-bower-json-file-version-property

josdejong:
> Considering it may be moving to NPM anyway

I'm not sure what having bower support at all has to do with leaving a version number inside `bower.json` (the issue at hand). But I very strongly hope that everybody will move to npm. Bower did fill a need, but it has a number of fundamental flaws, and having two package managers for (full stack) web development is very unpractical. I would be very happy when Bower will retire.

PR #966, opened 2015-06-11
Core.prototype.fit and Core.prototype.getDataRange fixes

mschallar:
Hi,

I stumbled upon some errors after upgrading to vis-4.2.0. Traced the bug down to these two functions:

    Core.prototype.getDataRange
    Core.prototype.fit

1.) `fit` accessed the wrong property of the variable `range`. Instead of `min` and `max` it did access `start` and `end`.

2.) `getDataRange` returned always `null` as `start` and `end`.

3.) Renamed the properties of `getDataRange` from `start` and `end` to `min` and `max` as the other functions were expecting such properties, and the documentation states it as well.

Regards

### DIFF
--- a/lib/timeline/Core.js
+++ b/lib/timeline/Core.js
@@ -459,7 +459,7 @@ Core.prototype.getVisibleItems = function() {
 Core.prototype.fit = function(options) {
   var range = this.getDataRange();
 
-  // skip range set if there is no start and end date
+  // skip range set if there is no min and max date
   if (range.min === null && range.max === null) {
     return;
   }
--- a/lib/timeline/Core.js
+++ b/lib/timeline/Core.js
@@ -496,8 +496,8 @@ Core.prototype.getDataRange = function() {
   }
 
   return {
-    start: start,
-    end: end
+    min: start,
+    max: end
   }
 };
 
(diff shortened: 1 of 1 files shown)


AdamNowaczyk:
Yep, just found the same issue. Is it going to be merge into main branch?

mschallar:
Yes I'd like to merge it to their develop branch. I'm waiting for a feedback.

josdejong:
Thanks Manuel, the issue in the method `fit` was already fixed in eef693113e54c7acb94dd9db187a31be8e305c47, but I hadn't seen error in `getDataRange`.

josdejong:
Turned out the code in `Core.prototype.getDataRange` was redundant and should actually return `{start: null, end: null}`: it's the abstract version of the library which is overridden by both `Timeline` and `Graph2d`.

mschallar:
Thanks for info. It was just a bit confusing and I thought it was an error. Maybe add some documentation and make the function do nothing - and let the implementing methods from the subclasses implement something useful?

josdejong:
maybe the abstract implementation should throw an error, so you know it should be implemented by a subclass.

Issue #529, opened 2014-12-30
Allow templating of groups in timeline


ccollie:
Similar to how templating works for timeline items, allow an optional groupTemplate option.

josdejong:
Thanks for the suggestion, makes sense.

AlexDM0:
Hi,

Over the last year a lot of feature requests have been made. We have just introduced our new website which has a list of the requested features. We have placed this request on that list. 

The list can be found here:
http://visjs.org/featureRequests.html

An explaination of the new system can be found here:
http://visjs.org/blog.html#New\ website\ for\ vis\.js\!

I would like to stress that this does not mean we abandon this request. Discussion here will continue if needed on this feature but we will close it to keep our Github issue page more of a bug-todo list.

Future feature requests will still be made here and then added to the website by us.

Regards,

Alex

josdejong:
This has been implmeneted by @hansmaulwurf23 and will be available in the first next release.

ccollie:
Thanks for this. I'll be needing this in a sprint or three.

PR #996, opened 2015-06-19
added templating for groups

hansmaulwurf23:
New option is called groupTemplate (like suggested in #529), changed docs, too.
Not sure whether I am accessing the timeline options the right way, though.

### DIFF
--- a/docs/timeline/index.html
+++ b/docs/timeline/index.html
@@ -581,6 +581,13 @@ function (option, path) {
     </tr>
 
     <tr>
+      <td>groupTemplate</td>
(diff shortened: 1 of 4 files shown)


josdejong:
Thanks Martin! Sorry for the late reply, I've been on vacation.

hansmaulwurf23:
Hi Manuel,

Isn't what you want to be found here: https://github.com/almende/vis/blob/master/examples/graph2d/09_external_legend.html? The groups of Graph2D are not drawn the same way they are in the Timeline... Or did I get you wrong?

PR #998, opened 2015-06-19
Physics pause button for navigation icons

odranoel75:
Hi, this is my first ever open source contrib, so please be kind if I've screwed up anything. 

 This is a pull request to implement a new toggle button in the network navigation to freeze or unfreeze the animations (the Icons I drew are a Play or Pause buttons).  This is required for one of our projects that is using Vis to show the relationship between threads in a server.  There are often hundreds or thousands of threads, and having the ability to pause the simulation is critical for a good user experience.


AlexDM0:
Hi, 

thank you for the pull request! I'm not sure if this should be enabled by default though. The navigation buttons are for navigation, not for controlling the simulation (though I agree it could be a good thing for certain cases).

Do you think you could make it an option? Also the example 
http://visjs.org/examples/network/other/navigation.html

should be update to show the new button with the description on how to enable it in the example and in the documentation.

Regards

PR #1004, opened 2015-06-22
Don't change backgroundColor if not requested on setOptions

ozydingo:
When backgroundColor is not specified in setOptions, the existing method was setting the backgroundColor to undefined, which defaults to white, regardless of current value. Other options in this method are assumed to not be changed if not specified, which is a reasonable mode for backgroundColor as well.


josdejong:
Thanks Andrew!

PR #1014, opened 2015-06-23
added getDataSet to network manipulation edge add and edit functions …

ericvandever:
…to match existing remove and node equivalents


AlexDM0:
Thanks :)

Issue #1018, opened 2015-06-24
Draw points only if they have an itemlabel


mschallar:
Hello,

It would be nice if there would be an option like:

    var options = {
      drawPoints: "onlyIfLabelExists" // [true, false, onlyIfLabelExists]
    };

which is availabel for the Graph2D as well as for the Groups.

I'm adding a label to my points every e.g., month and it's becoming unreadable if I zoom out

![image](https://cloud.githubusercontent.com/assets/6470225/8325736/0df1a3aa-1a5c-11e5-90d0-3215c908f5f4.png)

(I don't wanne use X and Y offsets in this example)

Regards,
Manuel

AlexDM0:
Thank you for your feature request. We want to redo the graph2d this year and we'll keep this in mind. 

Regards

PR #1027, opened 2015-06-25
Draw points only if they have an itemlabel

mschallar:
Hello,

I implemented my own feature request from https://github.com/almende/vis/issues/1018 

This option can be set for a Group or the Graph2d. The Group setting overwrites the Graph2d setting.
I've updated the docs and the 19_labels example as well.

Please let me know if there is anything else to do.

Regards,
Manuel

### DIFF
--- a/docs/graph2d/index.html
+++ b/docs/graph2d/index.html
@@ -557,6 +557,12 @@ function (value) {
             <td>Toggle the drawing of the datapoints.</td>
         </tr>
         <tr parent="drawPoints" class="hidden">
+            <td class="greenField indent">drawPoints.onlyIfLabelExists</td>
+            <td>Boolean</td>
+            <td>false</td>
+            <td>Datapoints will only be drawn if there is a label attached to them.</td>
+        </tr>
+        <tr parent="drawPoints" class="hidden">
             <td class="greenField indent">drawPoints.size</td>
             <td>Number</td>
             <td>6</td>
--- a/examples/graph2d/19_labels.html
+++ b/examples/graph2d/19_labels.html
@@ -22,7 +22,7 @@
 <body>
 <h2>Graph2d | Label Example</h2>
 <div style="width:700px; font-size:14px; text-align: justify;">
-  This example shows the how to add a label to each point in Graph2d. Each item can have a label object which contains the content and CSS class.In addition, xOffset and yOffset will adjust the location of the label relative to the point being labelled. 
+  This example shows the how to add a label to each point in Graph2d. Each item can have a label object which contains the content and CSS class.In addition, xOffset and yOffset will adjust the location of the label relative to the point being labelled. The label from the first item's group will only be drawn if a label exists. The label from the second item's group will always be drawn even if no label exists.
 
   
   <br /><br />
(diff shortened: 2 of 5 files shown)


AlexDM0:
Hi,

We cannot accept pull requests to master. All requests have to go into develop. 

I wonder, would it be cleaner to set an override called something like showPoints to groups and items? That way you can chose in the group to hide all points and in the specific item to show it or the other way around. I think it would be more flexible and achieve the same result. 

What do you think? These very specific options are very much bound to a single usecase. I prefer options that allow multiple usages. 

Regards

mschallar:
Sorry about the master/develop, I thought I had clicked almende:develop :-(.

Regarding your idea.
Sounds good at first glance, but:

This would mean that the server which provides the JSON data for all the items has to add the information whether to draw the point or not to **every** item which makes him somehow responsible for the view, this means:

1. The decision has to be made on the server whether to draw the point or not by setting the flag
2. The client (i.e., the options defined by the client) can not override it because the "item option" would overrule the "group option" and even the "graph2d" option.

IMHO is the client responsible for the view i.e., how the items received should be drawn. This must not be done on the server side, the server is responsible for the data. Means further that the server will attach item labels so items, and some items will not have a label. It's the client decision whether to draw those items or not, therefore the option 'onlyifLabelExists' sounds good to me.

Why to **every** item?
The server may not know have any information about the client's options. If the server should be responsible for the view (according to your idea) he must generate a valid view i.e., attaching to every  item the 'drawPoint' information. May also increase the JSON size significantly as well.

Hope to get your feedback.

Regards,
Manuel

AlexDM0:
Hi Manuel,

I agree that the server should not have to do this, but there is also the option of adding a preprocessor on the client side that checks if label exists and if yes, adds the exception to the item.

I figured something like [{x:x,y:y},{x:x,y:y},{x:x,y:y},{x:x,y:y},{x:x,y:y,label:'bla',point:true},{x:x,y:y},{x:x,y:y},{x:x,y:y}] would be clean? Another option would be to allow a callback function to be given, point could be true, false or function. the function would get information on the datapoint as an argument and returns true or false? or true false or object with information like radius and shape?

Regards

mschallar:
Hi Alex,

If you want callbacks for each individual datapoint than they can not be inside items like

    {
      x: x,
      y: y,
      label: 'bla',
      point: function (item, group, graph2d, whateverParameter, ...)  {
                  return true;
             }
    }

How should the server generate such a callback on an item? Not feasible IMHO. An alternative would be so add such a callback to the Graph2d options.

I think we (you) should even replace the `drawPoints.enabled` parameter from `true/false` to `true/false/function` instead of adding a new parameter.

    drawPoints {
      enabled: function(item) {
        // here we can access item, the group (by the item's groupid) and the graph2d (over the framework variable) and decide whether to draw the datapoint or not.
      }
    }

Yes, it's even an possibility for the function to return an object, containing more information about the drawing style of that data point like:

    return {
      style: 'square', // square, circle, according to docs
      size: 6 // the size of the data point, according to docs
    }

These 'square' and 'size' parameters could override the parameters for Group/Graph2d.

Regards,
Manuel

AlexDM0:
Hi Manuel,

Yes, I ment the callback at the graph2d or the group level, not the points. We would probably do it like so:

```
drawPoints: {enabled, style, size}  

OR 

drawPoints: true/false/function
```
where the function would return true/false/{style:x,size:x}

Would you like to implement this?

Regards

mschallar:
Yes I will. I'll do it on the Graph2d level, sounds better to me.

AlexDM0:
I think it should be over ridable in the groups though, different groups could have different requirements :)

mschallar:
You can access the group within the callback. There you can decide according to the groups policy whether to draw or not.

It's quite the same with the discussion Group vs. Graph2d option, as it was back an hour ago.
The groups could be provided by the server together with the items.

The only thing not provided by the server is the Graph2d's options. I think it's safest to update the options there. If there would be a separation between the actual groups, and the group options, your solution would work as well.

AlexDM0:
This is fully dependent on the architecture of your system. In your case the group settings may come form the server, in a lot of my cases I get pure x y data from servers and put them in groups with a preprocessor on the client side. 

Hence if the callback can be defined in groups as well as graph2d, it is the best of both worlds. In your case you can configure everything in the graph2d, in my case I can make callbacks for different groups.

Regards

mschallar:
Ok, I'll do it your proposed way. Will take a few days.

mschallar:
Few days are over ;)
New pull request: https://github.com/almende/vis/pull/1030

PR #1038, opened 2015-06-27
Add the missing group reference identification to exception.

skinkie:


AlexDM0:
Hi,

Thanks for the pull! We cannot merge into master unfortunately as this is completely overwritten every release. Could you make another pull req into develop?

Cheers

skinkie:
Will do so later today.

skinkie:
See Pull #1045

PR #1052, opened 2015-07-01
Use hasOwnProperty in Network.prototype._isMoving

janrotter:


AlexDM0:
Any elaboration as to why this is better?

janrotter:
Hi Alex,
I've encountered a code where some additional properties had been added to the Object.prototype and they were considered (wrongly) as node id's in the _isMoving function.

I've noticed, that in the rest of the Network.js the hasOwnProperty check is commonly used, so such situations won't occur. My commit changes the _isMoving function to follow that convention.

I failed to find any other checks for nodes[id] !== undefined or a reason to perform that check.

Best regards,
Jan

AlexDM0:
Hi Jan,

We actually use the undefined check a lot. HasOwnProperty is a lot more expensive computationally and I don't really want to change it for this one case of bad JavaScript practice. Don't touch what's not yours ;). 

Is it required to molest the Object?

Regards

janrotter:
Obviously not breaking the Object in the first place is the best solution :)

I've seen the hasOwnProperty being used in the _discreteStepNodes few lines later and it seemed the right way to go. I wasn't aware of the performance impact.

By the way, could you please be so kind and explain why the _discreteStepNodes uses hasOwnProperty and the _isMoving function checks for undefined? Shouldn't the behavior be the same in both cases?

Regards,
Jan

AlexDM0:
Hi,

Sorry you're right. I had a lot of other checks in mind, not specifically after a for in. 

I can accept this merge but we don't really support the v3 branch anymore. Is there a specific reason you don't want to switch the v4? 

Regards

janrotter:
No, I just didn't have much time to dive into the differences between the v3 and v4.

Is the v3 already a dead branch? For how long will it be alive?

Best regards,
Jan

AlexDM0:
Hi,

The v3 is just there for reference or very significant bugfixes. There will be no more releases on v3 and no new features. All work that is being done on vis is on v4. 

I'll accept your merge and rebuild the dist. 

Regards,

Alex

PR #1069, opened 2015-07-07
Udpated groupsPerformance.html

Rikkitp:
Changed count items in link to proper item count.


AlexDM0:
thanks :) seems like a good idea!

PR #1070, opened 2015-07-07
Missed "apply" call on DOMtoCanvas method

sc8696:
Missed "apply" call on DOMtoCanvas method


AlexDM0:
Hi,

Good catch! We cannot accept pulls to master I'm afraid though, we override master with develop on release. Could you make a new pull to the develop branch?

Regards,

Alex

PR #1072, opened 2015-07-07
Added feature that allows the border for a circularImage shape to be dashed

zukomgwili:
Added the following:
- dashes property to nodes in options.js
- dashes property with value false to defaultOptions in NodesHandler.js
- support for Arrays to selectiveNotDeepExtend function in util.js
- functionality to CircleImageBase.js that allows circularImage nodes to have dashed borders
- example demonstrating how to apply the dashed border to a circularImage node
- documentation explaining the usage of the dashes property on a node


AlexDM0:
Hi,

Thanks a lot for your contribution! I had a design spec in mind for these kind of properties and I hope you'd be willing to discuss adhering to it. 

There have been a few issues that require specific options for individual node shapes and this also applies to that.

My proposal is as follows: introduce a single new option called shapeProperties. This object would contain all configurations for specific shapes. Your specific contribution could be called borderDashes and should be functional for all shapes that have borders. If this works nicely it will give us a category of options that can be used to customise the specific node shapes. 

To illustrate other options: border radius for the box shape, apply size to images or use their absolute sizes and I'm sure more wil follow in the future. 

What do you think?

Regards,

Alex

zukomgwili:
Thank you Alex. I like your suggestion I will look into it's implementation
as soon as possible.

Kind Regards,
Zuko

zukomgwili:
Hi Alex,

I've implemented your recommendation. Is it any close to what you had in mind?

Kind Regards,
Zuko

docs/network/nodes.html
```diff
@@ -592,6 +591,21 @@
...
+        <tr class='toggle collapsible' onclick="toggleTable('optionTable','shapeProperties', this);">
+            <td><span parent="shapeProperties" class="right-caret"></span> shapeProperties</td>
+            <td>Object</td>
+            <td><code>Object</code></td>
+            <td>This object contains configuration for specific shapes
+            </td>
+        </tr>
+        <tr parent="shapeProperties" class="hidden">
+            <td class="indent">shapeProperties.borderDashes</td>
+            <td>Object or Boolean</td>
```
AlexDM0: this seems to be Array or Boolean going by the description?

examples/network/nodeStyles/circularImageWithDashedBorder.html
```diff
@@ -0,0 +1,99 @@
+<!doctype html>
```
AlexDM0: Since the new setting should apply to all shapes I'd propose the example to be closer to something like the shapes example. We can pick 5 or 6 shapes to demo it.

AlexDM0:
Hi,

I just reviewed the code (which was tricky due to all changes in spaces but thats fine) and it seems to me that the ellipse, database etc, all shapes with text inside of them do not have this new feature? The shapebase is only used by the shaped with text below them.

Other then that, looks good! I have added some comments in certain parts for review.

Regards

zukomgwili:
Thank you Alex for your feedback. I have made the requested changes

Regards

AlexDM0:
Hi,

Looks good! I'll merge it in! Thanks you for the contribution.

Regards,

Alex

AlexDM0:
(small tip for future use, you can use npm run watch-dev to only build the vis.js file, not the minified one. It's much quicker!)

zukomgwili:
Ah, cool. thanks :-)

AlexDM0:
Hi,

I just got issue #1112, and I'd like to warn you that the code you submitted will crash vis on firefox on unix systems. I have pushed a fix for this to develop. If you're deploying your code anywhere, make sure to add the fix!

Regards,

Alex

zukomgwili:
Thank you for the heads up, Alex, much appreciated.

Kind Regards,
Zuko

PR #1082, opened 2015-07-09
Added option zoomKey

hansmaulwurf23:
The option `zoomKey` can hold one of the wheel events properties to check whether alt, ctrl or meta was down ('pressed'). If set, the specified key must be down (evaluate to true) to apply zooming.
I have one question though: how are the available values for this option validated? I tried setting this in https://github.com/hansmaulwurf23/vis/blob/zoom_key/lib/timeline/optionsTimeline.js#L201 but when specifying 'bla' it does not show any error when instantiating the timeline. Is `optionsTimeline` the wrong file for this or is there no validation for available values for this kind of option?

Cheers
Martin

### DIFF
--- a/docs/graph2d/index.html
+++ b/docs/graph2d/index.html
@@ -1026,6 +1026,26 @@ function (option, path) {
             <td>The width of the timeline in pixels or as a percentage.</td>
         </tr>
 
+		<tr>
+      		<td>zoomable</td>
+      		<td>boolean</td>
+      		<td><code>true</code></td>
+      		<td>
+        		Specifies whether the Timeline can be zoomed by pinching or scrolling in the window.
+        		Only applicable when option <code>moveable</code> is set <code><code>true</code></code>.
+      		</td>
+    	</tr>
+
+		<tr>
+		    <td>zoomKey</td>
+		    <td>String</td>
+		    <td><code>''</code></td>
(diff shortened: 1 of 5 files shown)


josdejong:
Nice! We could apply the same for Graph2d, are you willing to apply the `zoomKey` option there too?

@AlexDM0 can you have a look at the issue with [`optionsTimeline`](https://github.com/hansmaulwurf23/vis/blob/zoom_key/lib/timeline/optionsTimeline.js#L201)? To me it looks like it should work like this.

AlexDM0:
Hi,

As for the option validation, there are two parts.

The configure options are for the configuration interface, which is where you have supplied the list of possibilities. In the actual validator part (allOptions) you only say that is has to be string. You can look at https://github.com/almende/vis/blob/master/lib/network/options.js#L82

for information on how to specify it.

Regards and thanks for the contribution!

hansmaulwurf23:
Hi!

@josdejong added the option to Graph2d and changed the docs accordingly.
@AlexDM0 thanks for the clarification. So if I don't want an option to appear in the configuration interface, I simply exclude it from the `configureOptions` map?

Cheers
Martin

josdejong:
Thanks for updating your PR.

> @AlexDM0 thanks for the clarification. So if I don't want an option to appear in the configuration interface, I simply exclude it from the configureOptions map?

That's correct.

Issue #909, opened 2015-06-02
Visjs.org Usability Feedback


felixhayashi:
Hi guys,

I really like the new structuring of the (network) docs at vis.js! Also many thanks for investing a big amount of time in the docs!

Just wanted to give you some feedback (after I did quite a bit of option-lookup for v4):

**Navigation**
When navigating to network "modules" like "interaction" or "groups" there is no link (or breadcrumbs) back to the overview (http://visjs.org/docs/network/). I need to click on Modules (http://visjs.org/index.html#modules) and Network -> view docs again.

**Search**
Before 4.0 all config was on one page. While this often made it hard to find a setting. It was also easy to "globally" search for options (ctrl+f). Maybe it is possible in the future to implement a search over all options and link to the page where a particular option is described.

**Page Titles**
When I open some of the option pages as tabs, they all have the same window title ("vis.js - A dynamic, browser based visualization library") which makes it hard to distinguish which tab contains which infomation.

-Felix

AlexDM0:
Hi Felix,

Thank you for the feedback! Good point about the breadcrums, I'll put a big button at the top.

We want to do some sort of search but we're not sure what the best way would be.

I'll update the titles as well, good catch!

Regards,

Alex

felixhayashi:
Thanks! :+1:

AlexDM0:
Hi Felix,

Since you got the hang of the pull reqs to the website, feel like picking this up? 

;) Worth a shot

Cheers

felixhayashi:
Hi Alex,

I'll have a look at it :)

-Felix

felixhayashi:
I added a pull for the breadcrumbs now. Once I find the time, I'll also have a look at the search ("tipue" looks promising for this).

felixhayashi:
Hi Alex,

To complete this I integrated a searchbar this morning.

It is only implemented for the network/layout section and only scans the network section.

At the moment tipue does not show an extract of the result are like google but I had a look at their issue list and they plan to implement it in the next release (https://github.com/Tipue/Tipue-Search/issues/20)

I designed it with bootstrap elements and use a panel for the results. The resultspanel is contained in the navbar so it doesn't visually mix with the content.

A question: In case you guys like it, where should I do the pull request?  The devolp branch only contains the docs section. Should I do two pulls one for gh_pages and one for develop?

Demo (go to network/layout module): http://bit.ly/vis_site_demo

Hope you like it

One Issue is that fork me at github overlaps with certain screen sizes.

[...]

PR #1084, opened 2015-07-10
yay! new search for visjs.org

felixhayashi:
see #909


AlexDM0:
Hi Felix,

It looks great! I have a few questions on the usability though, if you feel you're done I completely understand though :)

Is it possible to live-update the search results without having to press search? 

Could we make scrolling to the first occurence the default behaviour? If needed we can unfold everything when a search is happening?

Regards,

Alex

felixhayashi:
Hi Alex,

> Is it possible to live-update the search results without having to press search?

you mean like google does it? No, tipue does not allow that. I requires a page reload.

> Could we make scrolling to the first occurence the default behaviour? 

The reason I did it this way is because in case a keyword is hidden (for any reason) then the user thinks "I searched something and now nothing has happened" because he gets not feedback. I forgot that I also wanted to add the number of hits/matches to the info message so the user knows how many occurences exist. But if you don't want all this, I can directly scroll and remove the message.

> If needed we can unfold everything when a search is happening?

I did not really understand this question, when keywords exist in the url, I programmatically click on "full options". But of course it would be nice if this were not necessary.

-Felix

AlexDM0:
Hi Felix, 

>I did not really understand this question, when keywords exist in the url, I programmatically click on "full options". But of course it would be nice if this were not necessary.

For instance, if we search for colors there are only hits in the explaination of one of the options but those are folded (hidden). We could try to unfold (so show all hidden divs) before going to the first hit.

We could also leave the bar on top as it is now but try to scroll to the first occurence? We could also add the text to the bar "If you can't see the highlighted text, it may be in a folded option description." for clarity? (or something similar).

If I search for color in the network docs, it also sees graph3d docs and gives me those as first hit. Should we bind the search region to the module? Maybe only the module and the datasets?

Regards

felixhayashi:
> For instance, if we search for colors there are only hits in the explaination of one of the options but those are folded (hidden). We could try to unfold (so show all hidden divs) before going to the first hit.

Ahh, now I got it!! I now unfold all table elements when keywords are active. I updated the demo.

> We could also leave the bar on top as it is now but try to scroll to the first occurence?

Hmm, one last try to convince you here. I now added the number of search hits. It makes sense to show this information to the user before jumping to a first result because he/she might not know that there are other results as well somewhere (maybe far more down) in the page.

selection_404

Convinced or not convinced?

> "If you can't see the highlighted text, it may be in a folded option description." for clarity? (or something similar).

Not necessary anymore as I unfold everything in search mode.

> If I search for color in the network docs, it also sees graph3d docs and gives me those as first hit. Should we bind the search region to the module? Maybe only the module and the datasets?

This makes sense but it would make everything more complex. In this case we cannot store all sites centrally in a config js file (`docs/js/tipuesearch.config.js`) but would need an exclusive file for every module. Also the user maybe wants to search something related to the dataset while he is in the network module -- in this case: no chance unless the search is global. So I do not recommend it.

-Felix

AlexDM0:
Hi Felix,

Good points, you've convinced me. The click here to jump doesnt seem to work anymore though. I searched for color in the main network docs and clicked on the nodes page.

Regardss

felixhayashi:
> Good points, you've convinced me. 

* relief *

> The click here to jump doesnt seem to work anymore though.

Now that was a weird bug!!! Took me about an hour to figure that out. At start, the prettyprinting plugin recreates the spans (the ones used for prettyprinting the json) but keeps the highlighting. While on the surface everything seems ok, this makes the elements I stored in the variable before the click on the "click here button" parentless (=invisible). The solution is to call the select for the first highlighted element inside the click handler.

I further improved the script to always show the first visible hit at click time -- in contrast to the first hit (which may be hidden in another tab).

-Felix

AlexDM0:
Sounds good :) it's trickier then you'd think beforehand!

Is everything ready now for merging in?

Regards

felixhayashi:
Hi Alex

>  it's trickier then you'd think beforehand!

Yes it was. Didn't expext that. But now it's solved.

>  Is everything ready now for merging in?

Yes, this PR is up-to-date and may be merged. We'll see how it works on an every-day basis. In any case: it looks cool ;)

-Felix

AlexDM0:
Hi Felix,

Thanks for all the effort! Is does look great :). I'm gonna check if I can't force the search to do live searching with a little hacking :). 

Cheers

felixhayashi:
> Thanks for all the effort! Is does look great :). 

Thanks, I am glad you like it.

>  I'm gonna check if I can't force the search to do live searching with a little hacking :). 

Go for it! :)

Ehem... if you are inside `main.js`... maybe you could change 

https://github.com/almende/vis/blob/develop/docs/js/main.js#L120

to `if(firstHit.length) {`. I blindly expected jQuery to return null if first() is called on an empty resultset but it returns an object..

-Felix

AlexDM0:
Hi Felix,

I've added liveupdate for search and fixed the entree in main.

Regards

felixhayashi:
Hi Alex,

haha you really did it! :)

But now it prevents the user from completing the keywords because the request takes long and renders the input field inactive. Try to write "hello world its me" (medium speed) and see which letters make it into the input field. Makes it impossible to continue entering new keywords :(

-Felix

AlexDM0:
Hi Felix,

I can't seem to reproduce this, what browser are you using? It works on chrome here..

Regards

felixhayashi:
I am working with linux mint and I can barely enter text using either Chrome or FF (both latest browser versions).

AlexDM0:
I see, reproduced in firefox now. I'll see if I can tidy it up a bit :)

Cheers

AlexDM0:
Hi Felix,

I've updated it again. Its now REALLY fast :)

Regards

felixhayashi:
Hi Alex!

Yes, works fine now. Really nice! Seems like the main issue was that you accidentally directly executed the function in the timeout instead of passing a reference.

Anyways, glad your hacking was successful :) I can't wait to search the docs next time on visjs.org!! Now we only have to wait for the next tipue release in which the searc-hresult context is also displayed in the results...

-Felix

AlexDM0:
Hi Felix,

Actually, the thing was that I didnt cache the loaded pages. Tipue search is made for collect once on load, then search. Since I repeated it over and over it tried to get the pages over and over. That, and the timeout helps :).

Regards

felixhayashi:
Ah that makes sense. Now I also understand the reasons for the latest modifications in the tipue.js

felixhayashi:
Hi Alex

seems like the arrow next to "just search it" (https://github.com/almende/vis/blob/master/docs/img/arrow.svg) is not properly referenced in the css of the live version of visjs.org...

-Felix

felixhayashi:
Also the search doesn't work. Probably also a path issue.

I think it's because of the relative links used in `tipuesearch.config.js`.

AlexDM0:
Hi Felix, it seems the files werent added to git.
I've pushed them now, could you check?

felixhayashi:
Now it works :)

AlexDM0:
sorry for the oversight :)

PR #1091, opened 2015-07-14
remove subgroup from subgroups list if there is no more items in the …

ChenMachluf:
I found a bug that  happens cuz my prev commit.
I forget to delete the subgroup if there is no more items in it , that cause the orderSubgroups function to throw cuz it go to items[0].
Im also reduce the subgroupIndex member and call orderSubgroups after remove item from the group


AlexDM0:
good catch! thanks!

PR #1099, opened 2015-07-16
onMoved callback, ready event, item and group data accessors and partial stacking

Rodbourn:
Timeline.prototype.getItems() and Timeline.prototype.getGroups() were added so that the dataset can be pulled on demand, such as after a range is moved to update the models from which the data is derived.

The onMoved callback was created so that data as altered by the move can be accessed just after the drop has completed.

The 'ready' event was added so that dom manipulation of the items can occur after the timeline is 'ready'.  In my case it was setting coloring and z-index values based on the item data (putting items behind others when not stacking).

'Partial stacking' support was added by simply checking for item.data.stack when computing the vertical offset for stacking.


josdejong:
Thanks for your pull request, we will have a look at it next Monday (it's a bit busy today).

Rodbourn:
Thank you, hopefully it's an easy one to review :)

josdejong:
@Rodbourn I had a look at your code, a few remarks here:

- I don't think it's a good idea to expose the internal groups and items via `getItems` and `getGroups`, they are not directly suitable for that. Can you explain a bit more on why you needed this? I don't understand what you mean with "dataset can be pulled on demand".
- I don't fully understand what `onMoved` does: it's called like `onMoved(null, null)` so doesn't contain any data. Are you aware of the possibility to subscribe on changes in a DataSet (see example [editingItems.html](http://visjs.org/examples/timeline/editing/editingItems.html))? 
- The `ready` event is fired immediately and synchronously at the end of the Timeline constructor, I don't see what value this has: directly after you call `new vis.Timeline(...)`, the Timeline *is* ready, right?
- Can you explain more about the reason you need partial stacking? It sounds quite specific to me. At least it's not a good idea to require a property `stack: true` on every value by default which is the case in your PR if I'm not mistaken. It should be optional and if not provided use the global setting.

Rodbourn:
* Exposing the data was needed so that I could go through it and update any entities that would have new Start/End/Group values.  Should there not be some symmetry in that if you can set the data you can also get the data?  A copy of the data could be better, though.
* I wasn't, that makes this redundant.  Incidentally, if I make changes to the DataSet, will the timeline redraw?  I've been assigning new DataSets when there is updated data.
* I was using it in an angular directive to 'trigger' some post processing of the timeline items (dom manipulation of items)
* Agreed, I should have it as an optional value.  Of the PR this seems like the one that might actually be worthwhile if I update it to make it optional?

josdejong:
> I make changes to the DataSet, will the timeline redraw?

Yes, that's correct :D I suppose that resolves the first two points?

I think there is no reason to add this `ready` event to the Timeline. If you need it in your Angular app, you can emit it yourself after having created the Timeline.

The partial stacking could be interesting if we can come up with a general, broad enough use case for it, but I can't come up with a use case for this myself. Can you elaborate?

Rodbourn:
> Yes, that's correct :D I suppose that resolves the first two points?

*Very* nice, perhaps add a note to the affect in the docs?  I went through it a bit and didn't catch that.

> I think there is no reason to add this ready event to the Timeline. If you need it in your Angular app, you can emit it yourself after having created the Timeline.

Agreed

> The partial stacking could be interesting if we can come up with a general, broad enough use case for it, but I can't come up with a use case for this myself. Can you elaborate?

Certainly, image 'putting' items in a time window where there can be many duplicates occurring at the same time and you want to 'collapse' the duplicates only.

josdejong:
I've mentioned the 2 way data binding of the DataSet in the docs, see https://github.com/almende/vis/commit/fd6b70c1cb4a6b1e41a1ad019b4755d3e5cc5fc3. If you see more places/ways to clarify the docs please let us know.

As for the stacking: the logic to deal with "duplicate" items may be better outside of the Timeline, at the point where your prepare your data. I don't think it's something that should belong inside of the Timeline.

Rodbourn:
Thank you for the doc update (that is helpful).

For the item level control of stacking, I do think that is useful, perhaps not in that contrived example.  I do use it to 'expand'/'collapse' a group's items with stacking.  I have many groups, so the page become too tall if they are all stacked.

Similarly control over whether an item can be deleted would be nice (similar to how you can control whether an item can be edited).

Rodbourn:
By the way, without the getItems to get the inner item data which has the dom available, how would you get the element for an item in a DataSet?

josdejong:
Making groups expandable/collapsable sounds like an interesting idea, maybe worth it's own issue. But using an individual `stack` property feels a bit like a hack then... 

Normally you should not mess with the DOM of the rendered items, it's controlled by the Timeline and items are rendered/moved/removed continuously when dragging the Timeline. You could access them of course via some querySelector if needed. But why do you need access to the DOM of the items?

Rodbourn:
>  Making groups expandable/collapsable sounds like an interesting idea, maybe worth it's own issue. But using an individual stack property feels a bit like a hack then...

I'm certainly using it a bit as a hack to get just some of my items to stack (some I always want to be in the background as I am moving others 'into' them, a container of sort).

Is the ability to set item level properties a hack though?  Stackable/editable/deletable/etc?  Seems like it's just more precise control.

>  Normally you should not mess with the DOM of the rendered items, it's controlled by the Timeline and items are rendered/moved/removed continuously when dragging the Timeline. You could access them of course via some querySelector if needed. But why do you need access to the DOM of the items?

I need to update the dom of the element that I insert into the item range which is the result of the 'content' from the DataSet.  Updating the doms for the visible items is much faster than updating the DataSet's content, at least I have found.  Unfortunately a querySelector doesn't fit the angularjs directive pattern I am using around the timeline.

josdejong:
Maybe you can use background areas for items that should stay in the backgrounds? See example: http://visjs.org/examples/timeline/items/backgroundAreas.html

Sure more precise control for behavior and representation of individual items is good, but for this option to make them individually stackable I would like to see a good use case first. 

I don't know what you need to change to the DOM but I can imagine there are cases where that's needed.

Rodbourn:
> Maybe you can use background areas for items that should stay in the backgrounds? See example: http://visjs.org/examples/timeline/items/backgroundAreas.html

That would be perfect except the background items can also be edited just like the other items in my case.

> I don't know what you need to change to the DOM but I can imagine there are cases where that's needed.

Maybe a method to get the dom for an item by id?

josdejong:
> That would be perfect except the background items can also be edited just like the other items in my case.

You can mark individual items as editable/readonly :)

> Maybe a method to get the dom for an item by id?

You can provide HTML code or use [templates](http://visjs.org/docs/timeline/#Templates) to build the DOM for an item. In this HTML or template you can put attributes or identifiers or whatever to be able to find the item again via the DOM.

I'm not in favor of adding an API to get the DOM of an item, this invites people to mess around with the DOM, which can easily give problems. I think in many cases where people *think* they need to mess with the DOM, there is an other solution for the underlying problem (an XY problem).

mojoaxel:
@Rodbourn Sorry for the long delay  :hourglass_flowing_sand: As you can see in #1781 this project is seriously understaffed. 
Are you still interested on getting this merged?

Rodbourn:
It's been well over a year, I think the conflicts will be a bit high at this point.  If I need to apply these changes to the current branch for a client I'll create a new pull request (not expecting this to happen)

mojoaxel:
@Rodbourn Every help is welcome. Thanks again. Closeing (for now) ...

Issue #435, opened 2014-11-10
Dragging subgroups (up/down) to change the order.


nsconde:
It would be great if it would be possible to not only drag events, but also groups and sub groups.  Right now, I have a button the user presses on each lane that selects the children and drags all the events at the same time. However, I think it would be a useful feature to be able to drag a subgroup, whether it be a background or a sub group, and drag its children.  I hope that makes sense.

AlexDM0:
Hi,

Over the last year a lot of feature requests have been made. We have just introduced our new website which has a list of the requested features. We have placed this request on that list. 

The list can be found here:
http://visjs.org/featureRequests.html

An explaination of the new system can be found here:
http://visjs.org/blog.html#New\ website\ for\ vis\.js\!

I would like to stress that this does not mean we abandon this request. Discussion here will continue if needed on this feature but we will close it to keep our Github issue page more of a bug-todo list.

Future feature requests will still be made here and then added to the website by us.

Regards,

Alex

mojoaxel:
Reopening as Feature-Request issue (see #2114). 
Everybody: Please feel free to implement this!

yotamberk:
With #1109 it is possible to drag and reorder groups. To reorder items, you can use the multiselect and drag multiple items between groups. It doesn't make sense to drag all the content from group to group within the timeline. It should be implemented by you.

PR #1109, opened 2015-07-19
Group Dragging in Timeline

hansmaulwurf23:
Hi!

I started to implement the dragging of groups as suggested in #435. This feature is activated by the option `groupsDraggable`. Event handling is done by a second instance of hammer `groupHammer` which does not enforce the group to be selected first. The dragging code follows your item dragging concept with `_onDragStart`, `_onDrag` and `_onDragEnd`. To achieve changing the groups appearance I had to provide a default implementation for the `groupOrder` function, which relies on a new attribute `order` in the groups data.

This is a first idea, please help me with some problems that arise with this:
* In case of a user provided `groupOrder` function, the `groupsDraggable` option cannot be used, since the user takes care of the groups appearance. In my opinion `groupsDraggable` and `groupOrder` need some sort of special validation when the options are set.
* The name of the `order` attribute in the groups data should be configurable, so I suggest something like a `groupOrderAttribute` option to avoid conflicts with user defined attributes.
* For now the user has to set the `order` values when the DataSet for the groups is created. This could also be done in the DataSet event handler for `add` (`_onAddGroups`).
* The default implementation of `groupOrder` relies on `order` being a number. Maybe this should be a little more generic (for example using `String.localeCompare()` and such)?

Regards,
Martin


josdejong:
Thanks for picking this one up Martin!

First, the ability to drag groups is one of the features we need to fully manipulate groups: in the end we may want three things: (a) create new groups, (b) delete existing groups, (c) change the order of groups by dragging them. It's quite similar to the `editable` option of items, and maybe we should structure this similarly for groups: make an option `groupsEditable: {add: boolean, remove: boolean: order: boolean}` (with a shortcut to define `groupsEditable` as a boolean, to set all three options to either true or false).

The field which determines the groupOrder is already customizable, you can define `groupOrder` as string. This will indeed not work when `groupOrder` is a function. In that case the Timeline lacks information to know how to change the order. 

Just thinking aloud here: maybe we shouldn't dictate that the ordering is determined by one specific field, maybe we can just request a callback function from the user which will change the groups order for us. All we need is a function `groupOrder` to order the groups (already there), and a function `moveBefore` to move one group before another group. The implementation of `moveBefore` should take care of actually changing the order of groups, probably by changing values of some `order` field on the groups. We could provide and use a default implementation of `moveBefore` which makes use of a field `order` on the items, so a user is not forced to implement this function but can override it when needed.

hansmaulwurf23:
Hi Jos,

Thanks for the quick reply!

The thing with `groupsEditable` is a good idea, but isn't the timeline initialized like this
```javascript
var timeline = new vis.Timeline(container, items, groups, options);
```
with groups being a DataSet, which can be edited from outside? The event handlers (subscriptions) for the `DataSet` cannot cancel an insert, update or delete, right? How do you want to enforce the `groupsEditable` then? By resetting the old values in the subscriptions from within the timeline code?

I like the idea with `moveBefore`... Will give that a try later. For my usage of timeline, I will set up a `groupOrder` holding the name of the `order` field of the `groups`. My implementation of `moveBefore` is simply swapping the values of that `order` fields. Simple and nice! :-)

josdejong:
Currently groups can only be edited from outside, but that's exactly what we would like to change right. In the end groups will be completely editable by the user from within the Timeline, similar to how the user can edit/create/remove items inside the Timeline. 

I'm not sure what you mean with the issues round event handlers for the DataSet. Suppose we have a configuration option `groupsEditable: {remove: true}`, then we can create a "Delete" button on the group labels  when hovering them. If the user clicks the delete button, the Timeline will simply call `DataSet.remove(group.id)` on the DataSet of the groups.

btw maybe `groupEditable` (singular) would be a better name to be consistent with options like `groupOrder` and `groupTemplate`.

hansmaulwurf23:
Ok, I will implement the `groupEditable` option in a way that the `order` part of it will be enforced when `_onGroupDragStart` is executed. Removing and adding UI elements can be added later on.

Concerning the event handlers for the `DataSet`: I thought you want to enforce the `groupEditable` options, so that `add` and `remove` will fail on the `DataSet` if these options are set to false. If these options simply add or hide UI elements that enable the user to add or remove groups forget about my thoughts ;-)

About the dragging itself: I changed my code to use a moveBefore function to change the order. Obiously this function must increase/decrease a value (or multiple) that define the order of the groups (via the `groupOrder` option). The problem is though, that the dragging events are fired multiple times and invoking some kind of changing operation will increase/decrease way to many times (or are not invoked at all if moved too fast). Also, if the order is changed within `_onDrag` (over and over again) some sort of toggling can happen, since the need to change the order is defined by the event positions underlying group. 

The alternative implementation I tried is simply showing some kind of hover effect on the group that has the position which the dragged group will take. The actual changing of order is done in `_onGroupDragEnd`. 

Long story short: either dragging of groups is implemented live while dragging or at the very end (on mouse up). The first way has the effect of seeing live what will happen (within the items container) and the second one has not so many error prone side effects.
What is the preferred way for you here?

josdejong:
Sounds like a nice challenge :D 

> Concerning the event handlers for the DataSet

No, the DataSet operations should not be blocked, it's only the UI actions done by the user which must be enabled/disabled depending on the configuration of `groupEditable`.

Getting this working may indeed not be so trivial. I've had troubles with the toggling you describe in the past, when moving items from one group to another when adding/removing this item changes the height of the group. Can be tricky to get it right.

The dragging should not repeately fire the moveBefore, only if the target of moveBefore changes, right? Then it should replace the index value and shift the values of all lower down groups.

hansmaulwurf23:
Hi Jos!

It is done :-)

I used a function `groupOrderSwap` that swaps the order of two groups. `groupOrderSwap` is part of the options and has a default implementation. This is not a `moveBefore` but avoids error prone side effects we discussed in this thread.
In every execution of `_onGroupDrag` I check whether the order of the groups was messed up with misfires of the dragging event and fix the positions accordingly ([Line 1480](https://github.com/almende/vis/pull/1109/files#diff-d3c0773c9b1eaa102ae263e2f400774dR1480)). The `groupOrderSwap` is quite handy here since the value(s) of any order field(s) does not need recalculation whatsoever.
I found a way to avoid toggling as well. Before actually swapping the order of the dragged group and the underlying group I check whether any new order results in a stable constellation ([Line 1445](https://github.com/almende/vis/pull/1109/files#diff-d3c0773c9b1eaa102ae263e2f400774dR1445)). In other words when changing the order the `clientY` must be within the dragged group to create a stable constellation.
The best thing is: the code works with the option `order` being a fieldname or a function.
The new option is also incorporated in `Group` to add the classname `draggable` which sets `cursor: pointer` in [`labelset.css`](https://github.com/almende/vis/pull/1109/files#diff-57a28c5491c4d1c9631f32c99adfde2dR24).

Tell me whether you are fine with this concept and the naming of `groupOrderSwap` and the new classname `draggable` and I will change the docs, too.

I hope you like it :-)

josdejong:
Thanks, looks nice!

A few remarks: 
- I'm not sure about the `groupOrderSwap`. I suppose this only works for swapping two subsequent groups? This is normally the case as mouse movements are triggered a lot with normally just a few pixels change in position. But in general, it could be that you drag the group at position 5 on top of the group at position 2, then group 2 should get the index of group 3, group 3 that of group 4, group 4 that of group 5 and group 5 the index of group 2 right? There is a shift in the index. That was what I had in mind with the `moveBefore` function. But if you can guarantee that this can never happen I guess it's ok (and it's simpler than the whole index shifting thing).
- Right now the `groupOrderSwap` function accepts `Groups`, which is an internal data structure. Would't it be more logical for the end user when the function is invoked with the *data* of these groups, the exact same data that the user has added to his DataSet with groups? 
- I don't have a strong opinion about the name for the event `groupDragged`. Mabye it's good to keep it similar to the way you edit items ([see docs](http://visjs.org/docs/timeline/#Editing_Items)): you can provide event handlers for `onMove`, `onAdd`, `onRemove`, so we could implement this for groups like `onMoveGroup`, `onAddGroup`, and `onRemoveGroup`?
- It would be great if you could update the docs too :)

hansmaulwurf23:
Hi!

`groupOrderSwap` is not only working for subsequent groups, I guarantee ;-). If you dragged a group at position 5 on top of the group at position 2, then they change positions, thats true (and therefore wrong :D). But after that swap I check against the original order. Since only one group can be dragged at a time everything above the new position of the dragged group should be where it was before the drag. At the dragged position must be the dragged group. Everything below the dragged group should shift their positions one step down. Therefore I check the rest of the positions with a `newOffset`. If there is anything at the wrong position I call the swapping function until the rest of the positions is where they should be. 
In case the dragged group is moved below its original position everything above the original position should be where it was before the dragging started. From the position where the drag started down to new position of the dragged group all groups should have shifted their positions one step up (therefore the `origOffset`). Then again up until the lower end of all groups every group should have their original position, so I swap everything that is wrong. (Hopefully you get what I mean :D)
IMHO this is why the `groupOrderSwap` is a better way of achieving reordering, because in swapping no absolute values that define positions are recalculated or smth (and may drift apart) but simply exchanged.

> Right now the groupOrderSwap function accepts Groups, which is an internal data structure. Would't it be more logical for the end user when the function is invoked with the data of these groups, the exact same data that the user has added to his DataSet with groups?

I'm not sure what you mean. Shall I clone the data of both groups with `util.extend` first and/or do the actual updating on return of the function call? I thought `this.groupsData` is referring to the original `DataSet` since this is whats done in `ItemSet.prototype.setGroups`, right? 

> you can provide event handlers for onMove, onAdd, onRemove, so we could implement this for groups like onMoveGroup, onAddGroup, and onRemoveGroup?

Done in the updated Pull Request. Please have a look at that. On a `callback(null)` I restore the original order, which gives the user a good way of canceling the drag with own checks afterwards. Did you want me to honor the `callback` concept here, anyway?

> It would be great if you could update the docs too :)

I will update the docs after you decide the semantics of the event handlers. I promise :-)

josdejong:
Ah I get it. Smart solution with `groupOrderSwap`, that's a nicer and simpler solution than I had in mind! Nice!

As for passing the data of the group instead of the `Group` itself: right now a group does not store a pointer to it's data, you could store that in `Group.setData` by doing `this.data = data;` and initialize `this.data = null` in the Group constructor, same as `Item` does. Then when calling `groupOrderSwap` you can call `groupOrderSwap(draggedGroup.data, targetGroup.data, this.groupsData)`. 

I would say though that it's the task of the Timeline to send the changed data to the DataSet of the groups after `groupOrderSwap`, this should not be the task of groupOrderSwap. Keep the "swapping" logic separate from the "storage" logic. So it could look like:

```js
// ...
    groupOrderSwap: function(draggedGroupData, targetGroupData, groups) {
    	var targetOrder = targetGroup.order;
    	targetGroup.order = draggedGroup.order;
    	draggedGroup.order = targetOrder;
    },

// ...
this.options.groupOrderSwap(draggedGroup, targetGroup, this.groupsData);
this.groupsData.update(draggedGroup);
this.groupsData.update(targetGroup);
// ...
```
The third parameter passed to `groupOrderSwap` is now optional, but it's always neat to pass the "context" (similar to JavaScript's `Array.forEach(callback: function(currentValue, index, array))`.

Thanks for adding the `onMoveGroup`, `onAddGroup`, and `onRemoveGroup` handlers. We still have to add buttons on the UI to actually use them but it's nice that the structure is already in place.

hansmaulwurf23:
I still don't understand the issue with passing the group. Isn't the two passed parameters exactly what the user added to the `DataSet` when initializing the Timeline, since it is fetched via `groups.get(...)`? What is the advantage of changing the Group and holding a `data` attribute when the groups data can be fetched any time with the groups id?

Everything else is done. Now updating of the group for `groupOrderSwap` is done in the Timeline. I changed the docs and added an example for the `groupEditable` usage (including `groupOrderSwap`). Hope you like it.

josdejong:
> I still don't understand the issue with passing the group. Isn't the two passed parameters exactly what the user added to the DataSet when initializing the Timeline, since it is fetched via groups.get(...)?

Ah, sorry. I think I misread your code, you already did it like that! Great :)

I will merge your PR now, thanks a lot for all your work Martin!

josdejong:
You included a nice example :)

Maybe we can also add some visual clues for the mouse cursor when hovering and dragging a group (`cursor: move;`).

Issue #861, opened 2015-05-24
brokenImage only works for one node


BorisKingma:
Dear Alex,

as in the title, if I have multiple nodes with a broken image, only the first one shows a broken image. The rest shows no image at all (I tried up to three).

below is the code example excerpt to clarify:

this.nodes = new vis.DataSet([]);
this.edges = new vis.DataSet([]);
      	
this.nodes.add({id: 2, value: 1, label:"a", group:"alfa",  shape: 'image', brokenImage: IMAGE_DIR + 'logo.png', image: 'notexisting.png'});
this.nodes.add({id: 3, value: 1, label:"b", group:"beta",  shape: 'image',brokenImage: IMAGE_DIR + 'logo.png', image: 'notexisting.png'});

regards,
Boris

AlexDM0:
Hi Boris,

Thank you for reporting this. I'll look into it. 

Cheers

AlexDM0:
Hi Boris,

I cannot seem to reproduce this. Could you make a jsbin to show it?

Cheers

BorisKingma:
Hi,

I am not familiar with jsbin, but maybe it is helpful to know that the case
I had was that all images had broken links.

Regards,
Boris

AlexDM0:
Hi Boris, 

I managed to reproduce it. I'll make sure it's fixed for our release next week. 

Regards,

Alex

AlexDM0:
Hi Boris,

This is fixed in develop. I think this bug has been there for a long long time. Thanks for catching it :)

Cheers

PR #1119, opened 2015-07-23
Fixed issue where broken images wouldn't always be rendered.

brendon1982:
Hi,

Before I get into the details of the pull request. I am aware that it is a very heavy handed one. I usually try make as small a change as is necessary to achieve what I need but in this instance I couldn't find a small simple change to solve my problem. I have thoroughly tested my changes in my production project, where I use the broken image feature extensively, as well as all the examples that use images.

Now to the issue at had. I was having a problem with fall back images where sometimes they wouldn't render. This was quite weird because the problem wasn't that the broken images weren't loading, I confirmed in the browser developer tools that they were loading just fine, I even checked that the 'onload' event was firing for the broken image URL.

After quite a bit of investigation I found that the issue was happening because the 'onerror' event was firing multiple times when the imageUrl failed to load, I confirmed that this was happening even if I only had one node in the network! This caused a problem with the current implementation because it assumed that the first time the 'onerror' event fired that it was a failure for the imageUrl and the second time it fired that it was a failure for the brokenImageUrl and then it cleared the src property of the image, which at that point was set to the brokenImageUrl. I couldn't figure out why the 'onerror' event was firing multiple times, it wasn't because the load function was being called multiple times, or because the src property of the image was being set multiple times. So I decided to update the implementation to use a different function to handle errors in loading brokenImageUrl to errors in loading imageUrl, this removed the complexity of needing to figure out which URL had failed to load, and made the event being raised multiple times a non issue.

I am aware that the new implementation doesn't use the 'id' parameter that was introduced to fix issue #861, but I have tested this scenario with the new implementation and it works perfectly.

Finally thanks for the great library and for all the effort you put into it.

### DIFF
--- a/examples/network/nodeStyles/circularImages.html
+++ b/examples/network/nodeStyles/circularImages.html
@@ -45,7 +45,8 @@
         {id: 12, shape: 'circularImage', image: DIR + '12.png'},
         {id: 13, shape: 'circularImage', image: DIR + '13.png'},
         {id: 14, shape: 'circularImage', image: DIR + '14.png'},
-        {id: 15, shape: 'circularImage', image: DIR + 'missing.png', brokenImage: DIR + 'missingBrokenImage.png', label:"when images\nfail\nto load"}
+        {id: 15, shape: 'circularImage', image: DIR + 'missing.png', brokenImage: DIR + 'missingBrokenImage.png', label:"when images\nfail\nto load"},
+        {id: 16, shape: 'circularImage', image: DIR + 'anotherMissing.png', brokenImage: DIR + '9.png', label:"fallback image in action"}
       ];
 
       // create connections between people
@@ -64,7 +65,8 @@
         {from: 10, to: 11},
         {from: 11, to: 12},
         {from: 12, to: 13},
-        {from: 13, to: 14}
+        {from: 13, to: 14},
+		{from: 9, to: 16}
       ];
 
       // create a network
--- a/examples/network/nodeStyles/circularImages.html
+++ b/examples/network/nodeStyles/circularImages.html
@@ -66,7 +66,7 @@
         {from: 11, to: 12},
         {from: 12, to: 13},
         {from: 13, to: 14},
-		{from: 9, to: 16}
+        {from: 9, to: 16}
       ];
 
       // create a network
--- a/lib/network/Images.js
+++ b/lib/network/Images.js
@@ -10,58 +10,87 @@ function Images(callback) {
 }
 
 /**
- *
+ * @param {string} url                      The Url to cache the image as 
+  * @return {Image} imageToLoadBrokenUrlOn  The image object
+ */
+Images.prototype._addImageToCache = function (url, imageToCache) {
+    // IE11 fix -- thanks dponch!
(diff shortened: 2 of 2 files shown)


AlexDM0:
Hi Brendon,

Thanks for looking into this! I was not aware of any issue with the current implementation but then again, we don't use this specific part much outside of our examples. 

Before I merge it in, I have three (small) questions. 

1) you mention you have tested it a lot, is this with the 4.6 version?
2) could you add some more comments in the code? This part of the library is usually taken for granted and not looked at for months. Any stupidly simple explanation would be great.
3) we tend to make function declarations part of the prototype in order to keep everything short and readable and to better show the flow of the program. This allows you to keep using "this" instead of me as well. Do you think you could adhere to this in your refactor as well?

Regards,

Alex

brendon1982:
Hi Alex,

1) For our production system I was using 4.4, but I've just updated our production system to use the latest develop version (including the changes in my pull request) and ran it though all out use case tests and it is working perfectly. Before submitting the pull request I also used all the examples that are part of the repo and use images to do testing.

2) I've added comments :D

3) I've moved the previously internal functions to the prototype and prefixed them with an underscore to indicate that they are 'private'. I still needed to use a self/me variable so that I could access the correct instance in the event handler functions, but I think that this is unavoidable.

Regards,
Brendon

AlexDM0:
Hi Brendon,

Thanks for the refactor! Regarding point 3, I can rewrite it in ES6 format after merging (we're trying to update vis completely, network has been mostly rewritten for ES6) which solves the need for self by using arrow functions. 

If you think it'd be particularly fun to do this then by all means, if not I'll be happy to do that refactor later on.

I'll merge your code in on Monday, as it's a little busy here at the moment. Let me know if you feel up to making an ES6 refactor. (callbacks to arrow functions, usage of class instead of prototype and an export line that will be used to import the class in the required modules).

Regards

brendon1982:
So it turned out to be particularly fun :D

AlexDM0:
Hi Brendon,

ES 6 is easy to work with isnt it? :)

do we need to update the import lines in the modules that use it? 

Thanks for doing all this!

brendon1982:
Yea ES6 is awesome cakes :D

I don't think that we need to update the import lines as I compiled it after I updated to ES 6 syntax, then checked all the examples that use images, then updated my production code and ran through my test scenarios and it all worked fine.

It is a pleasure to contribute to this library, you are always so friendly and helpful, and it is by far the best library for drawing interactive networks. It has saved us so so so much time, the little we give back pales in comparison to how much we've gotten :D

AlexDM0:
I'm glad you like it! I've merged in the code. We will release it in the next cycle. I'm not sure yet when this will be but they tend to happen often. 

Regards

AlexDM0:
Hi Brendon,

Is #1152 due to your code? It's the only change in that region.

Regards

AlexDM0:
Hi Brendon, 

Sorry, this was my fault. Your code works! :)

Cheers

brendon1982:
Hi Alex, 

I apologise for not responding sooner, I've been completely snowed under this week. Glad to see I didn't break anything!

:D

PR #1131, opened 2015-07-28
fix "TypeError: Cannot read property '_bound' of undefined" in angular2 environment

Braincompiler:
... since zone.js (provided by angular2) implements a custom removeEventListener function and access the listener in any case also in case the listener is undefined:
```javascript
obj.removeEventListener = function (eventName, fn) {
    // arguments[1] is the listener function (fn) and undefined in that case
    if(arguments[1]._bound && arguments[1]._bound[eventName]) {
      var _bound = arguments[1]._bound;
      arguments[1] = _bound[eventName];
      delete _bound[eventName];
    }
    var result = removeDelegate.apply(this, arguments);
    global.zone.dequeueTask(fn);
    return result;
  };
```
Since `Core#_startAutoResize` calls `Core#_stopAutoResize` BEFORE `Core#_onResize` is set the error occurs.

### DIFF
--- a/lib/timeline/Core.js
+++ b/lib/timeline/Core.js
@@ -265,7 +265,7 @@ Core.prototype.setOptions = function (options) {
           onRender: options.drawPoints
       };
     }
-    
+
     if ('hiddenDates' in this.options) {
       DateUtil.convertHiddenOptions(this.options.moment, this.body, this.options.hiddenDates);
     }
@@ -900,8 +900,10 @@ Core.prototype._stopAutoResize = function () {
   }
 
   // remove event listener on window.resize
-  util.removeEventListener(window, 'resize', this._onResize);
-  this._onResize = null;
+  if (this._onResize) {
+    util.removeEventListener(window, 'resize', this._onResize);
+    this._onResize = null;
+  }
 };
 
 /**


AlexDM0:
Hi,

Thanks for wanting to contribute! I think you should take this up with our angular branch though.
https://github.com/visjs/angular-visjs

I can't merge angular specific code into the main repo.

Regards

Braincompiler:
Hi,

this is neither angular1-specific nor an problem of your angular-implementation :) Its first an angular2-specific problem cause of the zone.js and IMHO a more general problem of your library since it could always happen in case the removeEventListener function was replaced/overloaded by any 3rd-part lib. Independent of the environment :)

I just want to share my solution since i run into the issue.

josdejong:
Thanks for bringing this up Christian. 

I would say though that this is a bug in this `zone.js` code: it's replacing a native method `removeEventListener` (which is a questionable practice) with a non-standards compliant version (!!!). This should be fixed in `zone.js` itself. [From MDN](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/removeEventListener):

> Calling removeEventListener() with arguments that do not identify any currently registered EventListener on the EventTarget has no effect.

Anyway, on the pragmatic side, I'm ok with putting an `if(this._onresize)` statement around the removeEventListener, it would be neat as we indeed call `_stopAutoResize` before `_startAutoResize` at a time that `this._onResize` is uninitialized. We can only accept your fix on the `master` branch though, can you redo your PR to the `develop` branch?

AlexDM0:
Hi,

Sorry, I

AlexDM0:
Hi,

Sorry, I assumed that the code you showed in your comment was part of your pull request. Hence my reply that I cant merge angular specific code into the main repo.

We'd be happy to accept your pull request in the develop branch.

Regards

PR #1142, opened 2015-07-31
Fix misspelling in milliseconds

dburdick:


josdejong:
Thanks for this fix. We can only accept fixes on the `develop` branch though, can you please redo the PR to the `develop` branch?

PR #1153, opened 2015-08-03
fix : selectEdge was fired even if user clicked on a node

JonathanGeoffroy:
The selectEdge handler is now fired only if user clicked on an edge, in order to select it.
 It's not fired if the node are selected because there are connected to the selected node.

### DIFF
--- a/lib/network/modules/InteractionHandler.js
+++ b/lib/network/modules/InteractionHandler.js
@@ -205,20 +205,18 @@ class InteractionHandler {
       selected = true;
     }
 
-    if (selectedEdgesCount - previouslySelectedEdgeCount > 0) { // edge was selected
-      this.selectionHandler._generateClickEvent('selectEdge', event, pointer);
-      selected = true;
-    }
-    else if (selectedEdgesCount - previouslySelectedEdgeCount < 0) { // edge was deselected
-      this.selectionHandler._generateClickEvent('deselectEdge', event, pointer, previousSelection);
-      selected = true;
-    }
-    else if (selectedEdgesCount === previouslySelectedEdgeCount && edgesChanges === true) {
-      this.selectionHandler._generateClickEvent('deselectEdge', event, pointer, previousSelection);
-      this.selectionHandler._generateClickEvent('selectEdge', event, pointer);
-      selected = true;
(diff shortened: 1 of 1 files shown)


lib/network/modules/InteractionHandler.js
```diff
@@ -205,21 +205,19 @@ class InteractionHandler {
...
+        this.selectionHandler._generateClickEvent('deselectEdge', event, pointer, previousSelection);
+        selected = true;
     }
-    else if (selectedEdgesCount === previouslySelectedEdgeCount && edgesChanges === true) {
-      this.selectionHandler._generateClickEvent('deselectEdge', event, pointer, previousSelection);
+
+    // Check if the user clicked on an edge
+    let selectedNode = this.selectionHandler._getSelectedNode();
+    let selectedEdge = this.selectionHandler._getSelectedEdge();
+    if (selectedNode === undefined && selectedEdge !== undefined) {
```
AlexDM0: I don't think you have covered the case where you have 4 edges selected (multiselect), then deselect one. What happens now is that you fire a deselect AND a select edge event. You should add the node check in the if statement above.

lib/network/modules/InteractionHandler.js
```diff
@@ -205,21 +205,19 @@ class InteractionHandler {
       selected = true;
     }
 
-    if (selectedEdgesCount - previouslySelectedEdgeCount > 0) { // edge was selected
```
AlexDM0: this one I refer to below

AlexDM0:
Hi Jonathan,

Seems fine :) Thanks a lot!

Cheers

JonathanGeoffroy:
Hi !

I'm glad my contribution is merged. Thanks for answering me so fast !

Concerning the bug that you mentioned, you're right: there definitively a bug where user select multiple edges, and then deselect one of them. I'll try to send another pull request as soon as possible.

AlexDM0:
Hi, 

I thought I misread the code though.. If you could verify this and do a new pull that would be great!

Cheers

AlexDM0:
Hi,

I reverted the pull and fixed the bug. I was afraid we'd lose track of it.

Regards

PR #1174, opened 2015-08-08
feat: call the "addEdgeStart" handler when the user begins to create …

JonathanGeoffroy:
…an edge.

If the "addEdgeStart" function is defined by user, call it when the user starts to create an edge between two nodes. The handler gives the clicked node's id as unique argument.

### DIFF
--- a/docs/network/manipulation.html
+++ b/docs/network/manipulation.html
@@ -100,6 +100,7 @@ var options = {
     initiallyActive: false,
     addNode: true,
     addEdge: true,
+    addEdgeStart: undefined,
     editNode: undefined,
     editEdge: true,
     deleteNode: true,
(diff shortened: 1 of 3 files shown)


JonathanGeoffroy:
Hi,

It's a good idea in my mind; just to be sure: if the user returns false: the  temporary edge isn't added into network. So the creation of the edge is stopped.
Otherwise (handler returns true, undefined, or the handler is *not* defined) the behavior is the same than today (user can drag the new edge in order to connect two nodes).

What about the mode? Is the edit mode disabled ? What did you mean by "preparations for the mode and any actions" ?

I'll try to change my PR, I hope I'll have enought time to change it.
      
Regards

AlexDM0:
Hi Jonathan,

Here the touch (and other) handlers are temporarily replaced:
https://github.com/almende/vis/blob/develop/lib/network/modules/ManipulationSystem.js#L326

These are automatically reverted when you call 
```
this.showManipulatorToolbar();
```

It checks whether or not to actually draw it itself. You will also need to trigger the event that was replaced after you have restored everything (restoring is calling the showManipulationToolbar). In this case, the touch event was overridden which you can call using:
```
this.body.eventListeners.onTouch(event);
```

I would suggest to insert your method here:
https://github.com/almende/vis/blob/develop/lib/network/modules/ManipulationSystem.js#L941

Come to think of it, it is probably likely that the mode will remain on when you click the wrong node, in which case all you need to call when the return type of your function is explicitly false (=== false) is:
```
this.body.eventListeners.onTouch(event);
return;
```

Regards

AlexDM0:
Hi Jonathan, 

Is there any news on this?

Cheers

PR #1175, opened 2015-08-08
docs: the function's name was wrong

JonathanGeoffroy:
I changed the documentation, because the real name of this function is "setData", but the documentation talked about "setDataSet".
However, as there is a "getDataSet" function, you may wan't to change the "setData" function to "setDataSet".

### DIFF
--- a/docs/data/dataview.html
+++ b/docs/data/dataview.html
@@ -307,7 +307,7 @@ var view = new vis.DataView(data, {
 
     <tr>
       <td>
-        setDataSet(data)
+        setData(data)
       </td>
       <td>none</td>
       <td>


josdejong:
Thanks. At least the docs should be consistent, so I will merge your PR now.

Renaming the function to `setDataSet` would look more consistent, though that's would not be accurate either: you can set both a `DataSet` as well as a `DataView` as data source for a `DataView` using the method `setData`. The `getDataSet` function though will traverse over a (possible) chain of DataViews until it stumbles upon the DataSet at the root, and return that.

Maybe a better name for `setData` would be `setSource`?

PR #1183, opened 2015-08-12
Fixes documentation for group option of node.

vukk:
I was reading the documentation and noticed unfinished sentence for group option of network nodes.
This is a fix for that, feel free to modify if the wording is not to your liking.


AlexDM0:
Looks good! Thanks for the pull!

Regards,

Alex

PR #1209, opened 2015-08-21
Introduced the Graph2d group option 'excludeFromLegend'.

Bernd0:
It should be possible to exclude a group from the legend. The group option 'excludeFromLegend: false' excludes the group.

I updated the docs as well. I added the new flag in the "Graph2d options" section. It uses the green field style. Not sure if that is right, as there is no equal option for graph2d.


josdejong:
Thanks @Bernd0, looks good!

Issue #1207, opened 2015-08-21
Timeline group ordering performance suggestion 


ChenMachluf:
Hi,
I use vis timeline with 40k~ object that i update dynamically.
The performance bottleneck in vis is that is reorder all the group on every add/remove/update to the dataset.

I create a version of vis that instead of call group.order after every add/remove/update , 
it make sure that always the items is sorted.
i use binaryIndex to search the correct index to push the item in or to remove it.

i create 2 examples (based on visjs.org example) of timeline with 100000 items , with update and add function(1,1000,10000,10000 jumps)

original version : http://embed.plnkr.co/dioxgHCCj2TStrypDi9f/preview
my version : http://embed.plnkr.co/ujsM82uH2K3oannUCIYx/preview

As you can see to update one item in original version take more then a second.
In my version its take around milisecond.
But if we go to update 10000 or 100000 the original version is more faster , 
actually to the original version its almost the same time no depends of the items count (cuz it sort one time  all the items). 

Same is applied also to the add ,but just in the 100000 zone original version is faster.

this version solved my perf problems , cuz i update 400< items every few seconds.
my version make it slower if the update/add/remove is with big numbers.

I will love to pull request it if u think its better that way

josdejong:
Hi Chen, that sounds like a nice improvement. A pull request would be great!

josdejong:
btw I ran both your examples, and I don't get the expected performance improvements:

Chrome:
original version outputs "creation took 1553 mils"
improved version outputs "creation took 1295 mils"

Firefox:
original version outputs "creation took 1104 mils"
improved version outputs "creation took 1530 mils"

any idea?

ChenMachluf:
wow , ff do Array.sort much faster then chrome.
and chrome is faster using splice.
I check it only on chrome.... /:
In ff the creation is more faster and if you update/insert more then 500 items it better use the existing version.

I think (no ideal) maybe use some of hybrid depends on the count of the change element (and maybe platform? or overkill?)

what you think?

josdejong:
Thanks for your PR in #1211, that clarifies your solution. It's still not clear whether and when this actually yields an improved performance, I can't reproduce it.

Some sort of hybrid solution sounds indeed like overkill, unless it solves real performance issues for multiple real use cases. But if not, I prefer to keep the code as simple as possible.

[...]

PR #1211, opened 2015-08-21
make group keep orderedItems order every times instead of call group.order after every change

ChenMachluf:
as mentiod in #1207 .
Im using binaryIndex to keep the orderedItems always ordered.
It help with performance when update/remove/add items.

Need to decide if using this method or keep with the exisiting one , 
or use some kind of hybrid method that decide based of how many items changed the best option.

I pull request it so you can see changes and decide if we could improve it to some hybrid mode.

### DIFF
--- a/lib/timeline/component/Group.js
+++ b/lib/timeline/component/Group.js
@@ -350,6 +350,7 @@ Group.prototype.add = function(item) {
     this.subgroups[item.data.subgroup].items.push(item);
   }
   this.orderSubgroups();
+  this._addToOrderedItems(item);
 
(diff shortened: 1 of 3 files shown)


mojoaxel:
@ChenMachluf Sorry for the long delay :hourglass_flowing_sand: As you can see in #1781 this project is seriously understaffed. 
Do you you have any other idea how to improve performance or should we close this?

mojoaxel:
Ok, @ChenMachluf I'm closing this one for now. Feel free to reopen it at any time, if you still think this should be merged.

PR #1240, opened 2015-08-27
Allow groups to gather nodes around respective gravity centers.

glampr:
I use the option 'groupGravityStrength' (which is an Int) to set if a group should gather its nodes around it.
```javascript
var network = new vis.Network(container, data, {
  ....
  groups: {
    people: {
      color: 'blue',
      groupGravityStrength: 2,
    },
    companies: {
      color: 'red',
      groupGravityStrength: 2.5,
    }
  },
  ...
});
```

### DIFF
--- a/lib/network/modules/PhysicsEngine.js
+++ b/lib/network/modules/PhysicsEngine.js
@@ -11,8 +11,9 @@ var util = require('../../util');
 
 
 class PhysicsEngine {
-  constructor(body) {
+  constructor(body, groups) {
     this.body = body;
+    this.groups = groups;
     this.physicsBody = {physicsNodeIndices:[], physicsEdgeIndices:[], forces: {}, velocities: {}};
 
     this.physicsEnabled = true;
@@ -136,25 +137,25 @@ class PhysicsEngine {
       options = this.options.forceAtlas2Based;
       this.nodesSolver = new ForceAtlas2BasedRepulsionSolver(this.body, this.physicsBody, options);
       this.edgesSolver = new SpringSolver(this.body, this.physicsBody, options);
-      this.gravitySolver = new ForceAtlas2BasedCentralGravitySolver(this.body, this.physicsBody, options);
+      this.gravitySolver = new ForceAtlas2BasedCentralGravitySolver(this.body, this.physicsBody, options, this.groups);
     }
     else if (this.options.solver === 'repulsion') {
       options = this.options.repulsion;
       this.nodesSolver = new Repulsion(this.body, this.physicsBody, options);
       this.edgesSolver = new SpringSolver(this.body, this.physicsBody, options);
-      this.gravitySolver = new CentralGravitySolver(this.body, this.physicsBody, options);
+      this.gravitySolver = new CentralGravitySolver(this.body, this.physicsBody, options, this.groups);
     }
     else if (this.options.solver === 'hierarchicalRepulsion') {
       options = this.options.hierarchicalRepulsion;
       this.nodesSolver = new HierarchicalRepulsion(this.body, this.physicsBody, options);
       this.edgesSolver = new HierarchicalSpringSolver(this.body, this.physicsBody, options);
-      this.gravitySolver = new CentralGravitySolver(this.body, this.physicsBody, options);
+      this.gravitySolver = new CentralGravitySolver(this.body, this.physicsBody, options, this.groups);
     }
     else { // barnesHut
       options = this.options.barnesHut;
       this.nodesSolver = new BarnesHutSolver(this.body, this.physicsBody, options);
       this.edgesSolver = new SpringSolver(this.body, this.physicsBody, options);
-      this.gravitySolver = new CentralGravitySolver(this.body, this.physicsBody, options);
+      this.gravitySolver = new CentralGravitySolver(this.body, this.physicsBody, options, this.groups);
     }
 
     this.modelOptions = options;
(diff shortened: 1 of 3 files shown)


josdejong:
Thanks for your contribution Giorgos, we will review it next week. 

One first thing: we can't accept a pull request on the `master` branch, can you redo it to `develop`?

lib/network/modules/components/physics/CentralGravitySolver.js
```diff
@@ -1,12 +1,30 @@
...
+    if (this.groups != null) {
+      self = this;
+      util.forEach(this.groups.groups, function(groupDefinition, groupName) {
+        let gravityStrength = parseInt(groupDefinition.groupGravityStrength);
+        if (gravityStrength > 0) {
+          self.gravityGroups[groupName] = {name: groupName, strength: gravityStrength};
+        }
+      });
+    }
+    this._positionGroupCentersInitially(this.gravityGroups);
```
AlexDM0: positioning should not be done in the physics solver. All positioning is already done before the physics starts.

lib/network/modules/components/physics/CentralGravitySolver.js
```diff
@@ -1,12 +1,30 @@
...
     this.setOptions(options);
   }
 
   setOptions(options) {
     this.options = options;
+
+    // Iterates through the groups and looks for the ones defining 'groupCentralGravity'
+    if (this.groups != null) {
+      self = this;
+      util.forEach(this.groups.groups, function(groupDefinition, groupName) {
```
AlexDM0: this way of iteration is much slower than the for in way and you won't need self = this

lib/network/modules/components/physics/CentralGravitySolver.js
```diff
@@ -1,12 +1,30 @@
...
+        let gravityStrength = parseInt(groupDefinition.groupGravityStrength);
```
AlexDM0: this should not be an int but a float

lib/network/modules/components/physics/CentralGravitySolver.js
```diff
@@ -1,12 +1,30 @@
...
+    this.groups = groups;
+    this.gravityGroups = {};
     this.setOptions(options);
   }
 
   setOptions(options) {
     this.options = options;
+
+    // Iterates through the groups and looks for the ones defining 'groupCentralGravity'
+    if (this.groups != null) {
```
AlexDM0: should be !==

lib/network/modules/components/physics/CentralGravitySolver.js
```diff
@@ -1,12 +1,30 @@
...
+      self = this;
```
AlexDM0: missed a let, but this won't be needed if you read the comment below

lib/network/modules/components/physics/CentralGravitySolver.js
```diff
@@ -1,12 +1,30 @@
...
+      util.forEach(this.groups.groups, function(groupDefinition, groupName) {
+        let gravityStrength = parseInt(groupDefinition.groupGravityStrength);
+        if (gravityStrength > 0) {
```
AlexDM0: how about groupDefinition.groupGravityStrength === undefined and an isNaN(parseFloat(undefined)) check

lib/network/modules/components/physics/CentralGravitySolver.js
```diff
@@ -35,7 +54,44 @@ class CentralGravitySolver {
...
     forces[node.id].y = dy * gravityForce;
   }
+
+  /**
+   * Calculate the forces the group center puts on the nodes based on the distance.
+   * @private
+   */
+  _calculateGroupForces(group, forces, node) {
+    if (group != null) {
+      let dx = group.centerX - node.x;
```
AlexDM0: centerX and centerY are not options that can be set, they should be calculated by the solver

AlexDM0:
Hi,

I don't quite agree with your implementation. I'd prefer an extra solver instead of this being worked into the existing solvers.  

The centerX and centerY should be calculated based on the positions of the nodes at each iteration, probably for performance you'd index this once and rebuild the index on change of options. You do not catch the case where the options change after initialization. This will allow groups to grow closer without having predefined centers, resulting in a much clearer picture.

Vis is not made for manually specifying attractors, nor do I think it's good idea. It is not in line with  autoconfigurability of a network in vis.

You'd also need to add documentation, add it to the allowed options, give it default values and add an example to showcase the improvement. 

The physics engine would need an option to enable or disable this, as well as a multiplier to allow quick variation of the physics. I even think you don't need to specify a number in the groups, but we can keep it in there as a special feature. The easiest way to do this is by also allowing boolean input, which would just set value 1 in the backend.

Regards

lib/network/modules/PhysicsEngine.js
```diff
@@ -136,25 +137,25 @@ class PhysicsEngine {
       options = this.options.forceAtlas2Based;
       this.nodesSolver = new ForceAtlas2BasedRepulsionSolver(this.body, this.physicsBody, options);
       this.edgesSolver = new SpringSolver(this.body, this.physicsBody, options);
-      this.gravitySolver = new ForceAtlas2BasedCentralGravitySolver(this.body, this.physicsBody, options);
+      this.gravitySolver = new ForceAtlas2BasedCentralGravitySolver(this.body, this.physicsBody, options, this.groups);
```
AlexDM0: you add it here but there is no implementation in this solver.

lib/network/modules/components/physics/CentralGravitySolver.js
```diff
@@ -1,12 +1,30 @@
+'use strict'
+
+var util = require('../../../../util');
+
 class CentralGravitySolver {
```
AlexDM0: should have it's own solver.

PR #1243, opened 2015-08-28
Add German locale to the network module

Tooa:
This pull request adds German locale to the network module. Best Uli


josdejong:
Thanks Uli!

PR #1263, opened 2015-09-02
Fixed groups.visible Description

littleguga:
The column named "required", not "default value", so i changed cell value from "true" to "no" and added default value to description


josdejong:
Thanks for the fix! For future PR's can you please do them to the `develop` branch instead of `master`? Thanks!

PR #1325, opened 2015-09-28
Bad call to getHiddenDurationBetween

lauzierj:
Bug in Range.js call to DateUtil.getHiddenDurationBetween. It was passing in this.options.moment which is not expected by getHiddenDurationBetween.


josdejong:
Thanks for the fix Jacob, good find !

PR #1348, opened 2015-10-06
WIP: added physics web worker implementation.

ericvandever:
enable physics option useWorker to push physics calculations to a separate thread using webworkers.

This has only been tested with default physics, without improved layout and without stabilization.  Hopefully someone who uses those features can assist.


ericvandever:
TODO: propagate changes to mass to physicsWorker

AlexDM0:
Hi Eric,

Thanks for picking this up! How is the performance comparison? Quick question, is this only for the stabilization phase? I tried this once for each iteration but that was by slow.

The improved layout only gives X and y coordinates to the nodes, after which a stabilization is done. It should not affect your code. 

Does it work for all physics solvers?

I'll review the code when I'm back from holiday next month. 

Regards,

Alex

ericvandever:
Hey Alex,

Welcome, it's been a really interesting effort and hopefully can benefit someone else as well.  

So your first question is course a very difficult one.  From a user perspective testing on my developer machine, there is a 2-3x benefit for 'solving' large networks.  I've been primarily using worldCupPerformance and performance random numbers for testing.  The 2nd major performance enhancement is that the physics is no longer blocking the rendering, so framerates are significantly improved for large networks.  Of course all that comes with some trade-offs, most noticeably a short delay (300-400ms for 10k nodes) to copy and transfer the necessary physics data into the worker thread.  I spent a lot of today working on an _physicsUpdate that could take changes versus having to rebuild it every time a _dataChanged event occurs.  I'm sure there is lots more room for optimization in this area, but you will always have a penalty for initial data transfer until the web browsers trust us to do our own synchronization. 

On stabilization - actually quite the opposite as just looking at it to address your question, I'm pretty sure the current implementation effectively breaks it by counting each request for physicsTick as a physicsTick.  To date all my testing has been with stabilization off.

On the support for physics solvers, I have tested the defaults (Barnes/Central/Spring) pretty well, and did a cursory look through the others to see if they used any additional properties.  The key failure is a physics solver uses some piece of data that was not copied over into the Worker thread.  I did run into some issues today with the springsolver using both edge.fromId and edge.from.id that I'd like some feedback on.  It would simply the data transfer quite a bit if we can assume they are equivalent when connected is true.  I'd guess there are at least a couple other properties that need to worked out and several outstanding TODO for edge related updates.

I look forward to your thoughts and insights on the approach.
-Eric

AlexDM0:
Hi Eric,

The data transfer seems acceptable, does it work with clustering and the dynamic datasets? In other words, does it listen to a change in the data?

I think it'd be fine if only edge.fromId is used. Edge.from.id should be the same anyway. 

I would love it if the stabilization could be done in a worker. It may take some changes but it saves all the back and forth copying which should be much much faster. Also the extra algorithms in there should be a good improvement. 

The performance sounds really good!

Regards

AlexDM0:
Hi Eric,

I see there are some new commits (I dont get notifications of this). What's the current state of the code? I made a branch for this, could you do the pull req to that branch?

Regards

ericvandever:
Glad to see you're back, I've been away from the office the last couple weeks.  

I was able to get stabilization working through the webworker and integrated into my product as well.  We're using it daily for our small (20ish) beta group and haven't heard any issues related to it.

The remaining outstanding issue is improved/tested support for all the other physics solvers.  
  - BarnesHut - pretty well tested/verified
  - FA2BasedReplusionSolver - I copy over the number of edges on every dataChange event, but don't have support for sending updates if the number of nodes change without a dataChange event (haven't verified if this ever happens)
  - Hierarchical - haven't tested at all
There is also some issue with the calculations when using dynamic edges that have a control node.  I haven't looked into it, just noticed that it behaves significantly differently then without the webworker.

Just created a PR for the new branch.

PR #1349, opened 2015-10-07
Move babelify from devDependencies to dependencies

oncletom:
As devDependencies are loaded only for top level npm modules, and transforms are applied along the dependency chain, babelify has to be provided.

This is to scope babelify at vis level – and to avoid parent apps to install babelify as well.

Cf. https://github.com/substack/node-browserify#browserifytransform

> Make sure to add transforms to your package.json dependencies field.

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,6 +29,7 @@
     "watch-dev": "gulp watch --bundle"
   },
   "dependencies": {
+    "babelify": "^6.0.2",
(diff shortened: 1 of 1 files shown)


Tooa:
Hi,
you need to raise your pull request against the developer branch (AFIK). Best Uli

oncletom:
Done, thanks :-)

PR #1353, opened 2015-10-08
Move babelify to dependencies

oncletom:
As devDependencies are loaded only for top level npm modules, and transforms are applied along the dependency chain, babelify has to be provided.

This is to scope babelify at vis level – and to avoid parent apps to install babelify as well.

Cf. https://github.com/substack/node-browserify#browserifytransform

> Make sure to add transforms to your package.json dependencies field.

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,6 +29,7 @@
     "watch-dev": "gulp watch --bundle"
   },
   "dependencies": {
+    "babelify": "^6.0.2",
     "emitter-component": "^1.1.1",
     "hammerjs": "^2.0.4",
     "keycharm": "^0.2.0",
@@ -39,7 +40,6 @@
   "devDependencies": {
     "babel": "^5.1.11",
     "babel-loader": "^5.0.0",
-    "babelify": "^6.0.2",
     "clean-css": "^3.2.1",
     "gulp": "^3.8.11",
     "gulp-concat": "^2.5.2",


josdejong:
Thanks for your idea Thomas. I would argue though that bablify should *not* be included in `dependencies`. If you just want to use vis.js, you don't need bablify, as vis.js is distributed with bundled files (i.e. `./dist/vis.min.js`). Only if you want to do custom bundling or directly use a specific source code file, you need bablify as the code uses ES6. But in that case you're project will (probably) already be using bablify or webpack.

oncletom:
I understand that by using the distributed build you don't need browserify.

What I coin here is the `main` field of package.json does not reference the distributed build and that I did not have `babelify` explicitly declared in my project so far.

*Assuming* people use Webpack or babelify is not a good guess: npm modules are not supposed to be "best consumed with Webpack/brunch/drop a frontend tool name here" like we used to have websites best used in IE9: this is the point of having transforms declared as `dependencies`, to enable any kind of top level tooling, now and later.

The idea of this is module's authors have the power to switch to another compiler, anytime, without having all their dependents to figure out they have to manually change their own deps.

If you disagree with this, I encourage you to propose a change in browserify and tell [babelify dependends](https://www.npmjs.com/browse/depended/babelify) they are wrong.

oncletom:
tl;dr `you need bablify as the code uses ES6` = stating explicitly `babelify` as a dependency.

Otherwise I am thinking of two options:

- point `main` to the distributed build filepath;
- stop declaring babelify as a transform (because as browserify doc states it, you have to declare your transforms as dependencies (and not devDependencies) so as it can work properly)

josdejong:
I think we're actually on the same page here...

> What I coin here is the main field of package.json does not reference the distributed build and that I did not have babelify explicitly declared in my project so far.

Which version of vis.js are you using? This was fixed 2 months ago, [the main points to `./dist/vis.min.js`](https://github.com/almende/vis/blob/master/package.json#L24). And if that where not the case, we should fix the main field, not saddling up people with having to build vis.js themselves. 

We do not to assume people are using webpack, browserify, ES6, or node-babel, thats why we distribute an ES5 bundle of vis.js and let people use this bundle by default. That's also why I think it's better not to make the distributed vis.js dependent on bablify, which would actually *force* people to run the code via some transformer (what you are arguing against). People should be able to just load vis.js right out of the box, straight in their browsers, or via any dependency manager without having to do special steps. And that's how it works right now :).

The docs describing [how to do a custom build](https://github.com/almende/vis#custom-builds) describe that you can do bundling with any tool (webpack, browerify, ...), and just works out examples using browserify. It doesn't force you in any way to use one specific tool.

josdejong:
Wait a second, I finally get why you wanted to move `babelify` from devDependencies to dependencies in the first place, it's because of the `browserify.transform` is not working right out of the box and requires you to install `babelify` first. 

I think we should indeed remove the `browserify.transform` field: it's not needed anymore as `main` points to a bundled ES5 file now.

oncletom:
Okay, that make sense :-) I suppose the CommonJS `require` bit worked but my browserify build hanged because of the declared transform but undeclared `babelify` in the `dependencies` field.

What I propose then, if I understood well the move operated a few months ago, is to close this PR and to open another one, which purpose would be to remove the `transform` field of the actual `package.json` file.

Also, if I do so, custom builds will have to explicitly declare the `babelify` transform (otherwise it would be picked up automatically from the `package.json` `transform` field).

What do you think?

josdejong:
Yes that's true, for custom builds you'll have to install `babelify` too. 

I was just thinking, now that the `main` field points to the bundled file, `none` of the modules in `dependencies` is needed there, we could move them all to `devDependencies` :D.

I don't have a very strong opinion here. It's a choice between automatically downloading modules which most users don't need, versus making things more smooth for "advanced" users doing custom bundling. Maybe you're original solution (moving `babelify` to `dependencies`) is best anyway, ease of use is very important. Shall we just go for that?

oncletom:
I think if we stick to `main` aiming at the `dist/` folder, we should not declare the the transform in `package.json` to avoid contaminating the dependency chain but rather keep it as `devDependencies`.

If you want to get rid of publishing distributed builds in `dist`, then `babelify` should be declared as a transform, moved to a `dependencies` and `main` aimed at `index.js`.

Up to you :-)

josdejong:
I think it's important to keep pointing `main` to the `dist/` folder: this allows using for example just the `DataSet` straight in node.js, and does not force people to install/use some ES6 transpiler before they can bundle vis.js in their apps. Only browserify has this smart feature where packages can define the transforms they need themselves, but others like webpack haven't. So this `browserify.transform` is not a solution for everyone, and people using webpack would still have an issue.

Ok then, I have removed the `browserify.transform` field from package.json, and moved all `dependencies` to `devDependencies`. Thanks for the discussion @oncletom !

oncletom:
Cool :-) 👍Thanks!

PR #1354, opened 2015-10-09
added option multiselectFiltersGroup

hansmaulwurf23:
Hi,

I added an option `multiselectFiltersGroup`. When holding Shift to select multiple items the currently selected item and the first selected item (if present) must reside in the same group. Upon matching this requirement the selected items must be within the range (as before) and filtered to be in this group. When no item was selected before, no filtering is done. When the first selected item is in another group as the currently selected item, the selection of the current item is canceled.

Martin


josdejong:
Thanks Martin, nice improvement! I was just thinking about the name for this option. `multiselectFiltersGroup` is quite ok. But maybe something like `multiselectPerGroup` is even more clear, what do you think?

hansmaulwurf23:
Hi Jos,

I think so, too. `multiselectPerGroup` is way better :-) I did the renaming and added the docs.

josdejong:
Nice, thanks!

PR #1357, opened 2015-10-10
Bottom panel no top

hansmaulwurf23:
Hi,

_long story short_: I added an option `noPanelTop` which can hold `none` (default), `both`, `top` and `bottom`. It shall affect the setting of the css property `top` for the vis-panels `top` and `bottom`. When set to anything but `none` the panels css property `top` is not set, so that one can set `bottom` via css.

_More detailed_: I am using the timeline in a fullscreen-ish kind, meaning that the time axis is in the top panel, which has fixed positioning. This was easily done by setting the css properties for it:
```css
div.vis-panel.vis-top {
  position: fixed;
  z-index: 10;
}
```
The `z-index` was necessary otherwise when scrolling the items move above the top axis. Using this made the grid lines disappear in the time axis, but i can live with that. I don't think there is another solution for the grid lines, since they reside within the center panel. 
Making the bottom panel behave the same "sticky" way was not so easy though. Setting `bottom` and the `position` to `fixed` did not work, because `top` is already set to be below the last row of items. If it's not set one can use `bottom` to stick it to the lower edge (`bottom: 0px;`). 

Don't get me wrong, I am open for other ideas to achieve a fixed positioned top and bottom panel, too. I am open to rename `noPanelTop` to something else as well, had no better wording for this. I will add this to the docs, of course, if you are ok with my PR :-)

Martin


josdejong:
Thanks for your PR. 

It's not entirely clear to me what the problem is you try to solve. If you give the timeline a fixed height (like `100%` or `600px`). The TimeAxis will just stay fixed at top or bottom, and the contents become vertically draggable. There is only a (long standing) request to show a vertical scrollbar when the contents do not fit to make this visually more clear.

hansmaulwurf23:
Hi Jos,

I wasn't aware of this timeline feature. Looks nice, but that missing vertical scrollbar is really an issue, because of the lack of visual feedback (that there is more to see) and the mouse wheel does not move the items panel. I got this whole 100%-height-timeline to work with both axis with css, but only if the css property `top` is not explicitly set. This solves the issue with the scrollbar, too ;-)

josdejong:
Ok then let's close this PR and go for implementation of a vertical scrollbar :D

PR #773, opened 2015-04-10
Added mouseover and mouseout events to the time bar.  This currently wor...

andyflatt:
...ks with RangeItems only.

PR #1367, opened 2015-10-14
Mouse over and out events for timeline items

andyflatt:
I have updated the example to use these events too.


andyflatt:
@josdejong  This is the recreation of https://github.com/almende/vis/pull/773 without the dist changes as requested.

josdejong:
Thanks @andyflatt , that's quite some time ago :)

A few remarks:
- It's expensive (performance wise) to add these event handlers to every individual item. Instead, you could attach these two event handlers to the ItemSet instead (event delegation). It can work exactly like [`ItemSet._onSelectItem`](https://github.com/almende/vis/blob/master/lib/timeline/component/ItemSet.js#L1650-L1676), which uses the method `.itemFromTarget(event)` to determine which item was tapped exactly. This also would not require any change in any of the item types.
- Can you please also update the docs with these two new events?

Thanks!

mojoaxel:
@andyflatt I would like to merge this. It would be very nice if you would make the changes, pointed out by [josdejong](/almende/vis/pull/1367#issuecomment-148639633). You can just make the changes in your local branch `andyflatt:develop` and push them. All changes get updated here in the pull request. Thanks!

Iskander508:
I have reimplemented the hover events via changes in ItemSet and created a different pull-request (Timeline hover (itemover, itemout) events #2099)

mojoaxel:
This feature got already implemented by #2099

PR #1381, opened 2015-10-19
DataSet Performance Enhancement

phimimms:
* Replaces for-in loops in DataSet.js and DataView.js with Object-key iteration of cached length for greater computational performance, as evident in these test cases https://jsperf.com/object-keys-iteration/113.


josdejong:
Thanks Philippe, nice performance improvement!

PR #1382, opened 2015-10-19
Ability to select edges / nodes without unselecting all

zefrog:
Hello,

I'm using your wonderfull library (the network / dataset parts) in a project, and was needing the ability to memorize a selection (both nodes and edges) to be able to reuse it later.

I propose in this pull request to add an optional parameter in `selectNodes` / `selectEdges` methods so that the selection is not lost if that parameter is set to false (`unselectAll`).

Hope this can help someone too. If something is wrong I'll be glad to make it better.

### DIFF
--- a/docs/network/index.html
+++ b/docs/network/index.html
@@ -922,10 +922,10 @@ function releaseFunction (clusterPosition, containedNodesPositions) {
             </tr>
             <tr class="hidden" parent="selectNodes">
                 <td class="midMethods">Returns: none</td>
-                <td>Selects the nodes corresponding to the id's in the input array. If highlightEdges is true or
-                    undefined,
-                    the neighbouring edges will also be selected. If unselectAll is true or undefined, this method unselects all other objects 
-                    before selecting
+                <td>Selects the nodes corresponding to the id's in the input array. If highlightEdges is true,
+                    the neighbouring edges will also be selected. If highlightEdges is undefined, the value of <code>selectConnectedEdges</code>
+                    is used (defaults to true).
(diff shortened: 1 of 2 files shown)


zefrog:
Also, I think it makes sense to use the `selectConnectedEdges` value when the highlightEdges parameter is ommited, because otherwise you are forced to specify it when selecting nodes programmatically, or the behavior will not be consistent with the user interraction.

What do you think ? Will this break anything else ?

josdejong:
Thanks, it's an interesting idea. 

I'm thinking into a different direction for this issue though: what we miss is a method `setSelection` (similar to `getSelection`) which allows setting both nodes and edges at once. If we have such a method, it becomes easy to "memoize and rerestore" a selection yourself: first call `getSelection`, then append the nodes/edges that you want to make selected too, then call `setSelection`.

zefrog:
I'm a bit confused about this, because it would cause some duplicated code. The method would be basically the combination of `selectNodes` and `selectEdges` but with the call of `unselectAll` in both methods it's not possible to call these.

josdejong:
We would move and merge the logic that's currently in `selectEdges` and `selectNodes` into a single method `setSelection`. The methods `selectEdges` and `selectNodes`can be reduced to simple wrapper methods calling `setSelection` internally with either the provided nodes or edges.

AlexDM0:
Additionally, if we go with more than one optional option, I'd prefer to have a JSON optoin object as argument. I agree with Jos that we should merge the selectEdges and selectNodes into a setSelection.

Cheers

zefrog:
I'll close this PR in favor of a new one with a setSelection method.

Regards.

PR #1395, opened 2015-10-24
Added allDraggable option for timeline to support dragging unselected item

liuqingc:
I added an option for timeline so that we could drag the item even it is not selected in the timeline. Please review. Thanks!

### DIFF
--- a/docs/timeline/index.html
+++ b/docs/timeline/index.html
@@ -666,6 +666,14 @@ function (option, path) {
         and set repeat to weekly.
       </td>
     </tr>
+
+    <tr>
+      <td>itemsAlwaysDraggable</td>
+      <td>boolean</td>
+      <td><code>false</code></td>
+      <td>If true, all items in the Timeline are draggable without being selected. If false, only the selected item(s) are draggable.</td>
+    </tr>
+
     <tr>
       <td>locale</td>
       <td>String</td>
--- a/lib/timeline/component/ItemSet.js
+++ b/lib/timeline/component/ItemSet.js
@@ -43,6 +43,7 @@ function ItemSet(body, options) {
 
     selectable: true,
     multiselect: false,
+    allDraggable: false,
 
     editable: {
       updateTime: false,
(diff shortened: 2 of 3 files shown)


josdejong:
Thanks for your PR @liuqingc.

We've had this behavior in the the past: moving items without having them selected. However, in practice this worked out very badly: people can easily move items accidentally (and unnoticed) whilst moving the timeline. It's worse when the Timeline is crowded with items. My opinion is that this behavior is undesirable and also should not be supported as an option.

liuqingc:
Hi @josdejong , thank you for your reply.

I understand the disadvantage of having this feature: people could move items accidentally without noticing they had moved the items. There are also disadvantages of not having this feature: 1) People might think that the items are not draggable if they fail to drag the unselected items. It could take some time for them to figure out how to drag the item. 2) Even when people know that they have to select the item first, they may still attempt to drag them but ended up dragging the timeline. 

I have been using the library for a while and people are always complaining that they are not able to move the item. We don't have a lot of items on the timeline and the items are of the same size. If an item was moved, there would be a content refreshing within the item. In such cases, we could avoid the disadvantage of having this feature.

I think the feature has two sides and the option could leave people choices to decide depending on their use cases.

josdejong:
> I think the feature has two sides and the option could leave people choices to decide depending on their use cases.

yeah, you're right with that. Well ok then let's add this new feature. 

I'm just thinking a bit about what would be the best name, `allDraggable` sounds a bit vague. How about something like `alwaysDraggable`, `itemsAlwaysDraggable`, ... ?

liuqingc:
Glad to hear that. We could go with 'itemsAlwaysDraggable' which is self-explaining.

Thanks,
Liuqing

josdejong:
Sounds good, lets go for `itemsAlwaysDraggable`. Can you update your PR accordingly? Then I will merge it. Thanks!

liuqingc:
Done. I have update the codes. Thanks.

josdejong:
Thanks!

PR #1413, opened 2015-11-04
Fix the timeline example link in docs

labianchin:

PR #1415, opened 2015-11-05
Fix the example link in timeline docs

labianchin:
Ping @AlexDM0

This has the same fix as #1413. Now with the timeline examples instead of time examples.  :)


AlexDM0:
tnx :)

PR #1421, opened 2015-11-06
setSelection feature

zefrog:
Following the PR #1382 and your remarks about a new `setSelection` method, this is a proposal for such a method, keeping backward compatibility.

Let me know if this is alright.

Regards.

### DIFF
--- a/lib/network/modules/SelectionHandler.js
+++ b/lib/network/modules/SelectionHandler.js
@@ -609,32 +609,35 @@ class SelectionHandler {
   setSelection(selection, options = {}) {
     let i, id;
 
-    if (!selection || !selection.nodes || !selection.edges)
-      throw 'Selection must be an object with nodes and edges properties';
-    // first unselect any selected node
+    if (!selection || (!selection.nodes && !selection.edges))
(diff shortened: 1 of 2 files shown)


lib/network/modules/SelectionHandler.js
```diff
@@ -601,6 +601,44 @@ class SelectionHandler {
...
 
+  /**
+   * Updates the current selection
+   * @param {{nodes: Array.<String>, edges: Array.<String>}} Selection
+   * @param {Object} options                                 Options
+   */
+  setSelection(selection, options = {}) {
+    let i, id;
+
+    if (!selection || !selection.nodes || !selection.edges)
```
josdejong: For convenience, we could allow only providing one of the two: `nodes` or `edges`, i.e. allow `Network.setSelection({nodes: [1,2,3]})`.

josdejong:
Thanks @zefrog , nice work. I added one inline comment but besides that it looks very neat.

zefrog:
Ok I considered your comment and rewrite a bit the function. I also removed the vars in selectEdges and selectNodes methods which were no longer used.

josdejong:
Thanks @zefrog !

PR #1423, opened 2015-11-07
cloning old data before updating to fix #1378

### LINKED ISSUE #1378: DataSet original data not available
Hi there

I'm not sure if this is a bug in the docs or if I just misunderstand the functionality of the DataSet and its events. The documentation at http://visjs.org/docs/data/dataset.html says in the events section:

> The update event has an extra field data containing the original data of the updated items, i.e. the gives the changed fields of the changed items.

If I look at the example on http://visjs.org/examples/timeline/interaction/eventListeners.html, I'd assume that if I for example move "item 3" from Thu 18 to Tue 23 to the right, the properties.data would show the old position. But instead the logEvent displays the current value in the rendered output:

```javascript
event="update", properties={"items":[3],"data":[{"id":3,"content":"item 3","start":"2013-04-23T22:00:00.000Z"}]}
```

Is this a different event? Or am I looking at the wrong place? It would be really nice to have the old data of an item (object before the update) for my specific use case. Thanks for this great library and keep the good work going!

Marco

hansmaulwurf23:
Hi Jos,

I think this will fix #1378. I was not sure whether to `extend` or `deepExtend` though, so please change if necessary.
Fixed the typo `Object.key` as well.

Cheers
Martin

### DIFF
--- a/lib/DataSet.js
+++ b/lib/DataSet.js
@@ -243,10 +243,11 @@ DataSet.prototype.update = function (data, senderId) {
   var addOrUpdate = function (item) {
     var id = item[fieldId];
     if (me._data[id]) {
+      var oldData = util.extend({}, me._data[id]);
       // update item
       id = me._updateItem(item);
       updatedIds.push(id);
-      updatedData.push(item);
+      updatedData.push(oldData);
     }
     else {
       // add new item
@@ -513,7 +514,7 @@ DataSet.prototype.forEach = function (callback, options) {
   var filter = options && options.filter,
       type = options && options.type || this._options.type,
       data = this._data,
-      itemIds = Object.key(data),
+      itemIds = Object.keys(data),
       i,
       len,
       item,
--- a/lib/DataSet.js
+++ b/lib/DataSet.js
@@ -274,7 +274,15 @@ DataSet.prototype.update = function (data, senderId) {
     this._trigger('add', {items: addedIds}, senderId);
   }
   if (updatedIds.length) {
-    this._trigger('update', {items: updatedIds, data: updatedData}, senderId);
+	var props = { items: updatedIds, oldData: updatedData };
+	// TODO: remove deprecated property 'data' some day
+	Object.defineProperty(props, 'data', {
+	  'get': (function() {
+	    console.warn('Property data is deprecated. Use DataSet.get(ids) to retrieve the new data, use the oldData property on this object to get the old data');
+	    return this.get(updatedIds);
+	  }).bind(this)
+	});
+    this._trigger('update', props, senderId);
   }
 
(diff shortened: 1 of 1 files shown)


josdejong:
Thanks Martin, that should indeed to the trick.

I was just thinking: maybe we should rename the property as well, rename 'data' to 'oldData' because you can easily misunderstand this. This would be a breaking change though.

hansmaulwurf23:
Hi Jos,

You are probably talking about the key in the arguments to the updated 
```javascript
this._trigger('update', {items: updatedIds, data: updatedData}, senderId);
```
trigger. I think it would make this more intuitive, but as you stated breaking the api. What about changing this to
 ```javascript
this._trigger('update', {items: updatedIds, data: updatedData, oldData: updatedData}, senderId);
```
Javascript is passing it as a reference, so the memory overhead is irrelevant (right?). This could be changed to use only `oldData` in latter versions and in the meanwhile marked as deprecated. 

Cheers
Martin

**Comment by josdejong** [comment]
That's true, we can keep the current `data` property working. However, if people rely on it right now, it would break their code anyway: after the change, `data` will contain the oldData whilst currently it contains the newData. 

If we consider this a bug, we can simply break the code by removing `data` and adding this new `oldData`, so people will figure out very quicky that their code relied on buggy behavior.

If we consider the current behavior as ok but wrongly documented, we could keep things backward compatible, but then `data` should contain the updated data, like it did before, and we could output a warning in the console.

How about something like this:

```js
var props = {
  items: updatedIds, 
  oldData: updatedData
};
// TODO: remove deprecated property `data` some day
Object.defineProperty(props, 'data', {
  'get': function () {
    console.warn('Property data is deprecated. ' + 
      'Use DataSet.get(ids) to retrieve the new data, '  + 
      'use the oldData property on this object to get the old data');
    return this.get(updatedIds);
  }.bind(this)
});
this._trigger('update', props, senderId);

```

hansmaulwurf23:
> consider the current behavior as ok but wrongly documented

Sounds good to me, really a nice solution, now! :-)

josdejong:
Thanks a lot, and sorry for the delay.

josdejong:
@hansmaulwurf23 I made a mistake: I was assuming that the `data` property contained all data of the update item, whilst it only contains the updated fields. I restored the behavior of the old `data` property, turns out it was actually used by Network, causing a problem: #1437. 

We want to do a release now, so let's keep both `oldData` and `data` for now. After that it may make sense to deprecated `data` again, create a util function to extract the changed fields, and use that in Network.

hansmaulwurf23:
Ok, thanks for the update. Will keep it in mind when updating.

PR #1440, opened 2015-11-17
Webworkers network

ericvandever:
Import of Physics webworker work-in-progress

See https://github.com/almende/vis/pull/1348 for discussion

### DIFF
--- a/docs/network/physics.html
+++ b/docs/network/physics.html
@@ -139,7 +139,8 @@ var options = {
       fit: true
     },
     timestep: 0.5,
-    adaptiveTimestep: true
+    adaptiveTimestep: true,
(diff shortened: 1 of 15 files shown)


AlexDM0:
Hi Eric,

I'll merge it once I have time to take a look at it. The FA2Based should be exactly the same as the BarnesHut, only a few methods are overwritten. 

If this all works, why would we not make it the only option? Just move all physics to the workers.

For additional gains, the initial positioning with the improved layout should also be in a worker. It blocks a bit as well.

Cheers

Triet:
Whats the current status on this work? This seems to be very promising in terms of performance. Can we get this merged to the main branch?

PR #1463, opened 2015-11-26
Graph2d updates

ludost:
Adding 0-axis oriented shading, documentation and example.

Also contains some minor fixes: 
-DataSet potential undefined input fix.
-Removed some incorrect documentation for graph2d.

shading


josdejong:
Thanks, looks good. Thanks for updating the docs and providing a clear example :).

I'm not sure about your fix in commit [Prevent potential null pointer error](https://github.com/almende/vis/commit/9bf7452041ece6e7a0b95c227d7e2fe5045872df). I don't think it's a bad thing if the DataSet throws an exception when the provided data contains null or undefined values. It could be a sign that something is wrong with your data. With your fix, these cases are silently ignored. Do you have a use case where the data intentionally contains null or undefined values?

josdejong:
Thanks for the update!

PR #1464, opened 2015-11-26
Add link/button to new example 20, shading.

ludost:
Companion to pull-request #1463.


josdejong:
Thanks, we will merge this when we do the first next release.

PR #1478, opened 2015-12-01
Fix vis.DataView as input for Graph2d

ludost:
This PR will allow the use of DataView as input for Graph2d.


josdejong:
Thanks Ludo!

PR #1495, opened 2015-12-07
Cleanup, performance and new shading & stacking

ludost:
This pull request contains a large set of changes to the Graph2d code:
-Introduction of group-based shading, in which a specific group is used as shading base. (shading between groups)
-Introduction of stacking of line graphs
-Re-centralization of render calls, cleaner code
-Performance optimizations: less data copies and slow array initialization

Still todo: 
-Move icon rendering to the type specific files (where current data rendering is also done)
-In case of stacking line graphs with shading, the shading should default to the next stacked line
-Documentation of the line-stacking feature

### DIFF
--- a/examples/graph2d/17_dynamicStyling.html
+++ b/examples/graph2d/17_dynamicStyling.html
@@ -79,7 +79,8 @@
                         <select id="fill" onchange="updateStyle()">
                             <option value="">none</option>
                             <option value="top">top</option>
-                            <option value="bottom" selected="selected">bottom</option>
+                            <option value="bottom">bottom</option>
+                            <option value="zero" selected="selected">zero</option>
                         </select>
                     </td>
                 </tr>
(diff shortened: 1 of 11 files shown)


examples/graph2d/17_dynamicStyling.html
```diff
@@ -210,7 +211,7 @@
                 style: 'square' // square, circle
             },
             shaded: {
-                orientation: 'bottom' // top, bottom
+                orientation: 'zero' // top, bottom
```
AlexDM0: didnt you just remove this option in the documentation above?

lib/timeline/component/GraphGroup.js
```diff
@@ -29,20 +26,28 @@ function GraphGroup (group, groupId, options, groupsUsingDefaultStyles) {
   this.visible = group.visible === undefined ? true : group.visible;
 }
 
+function insertionSort (a,compare) {
```
AlexDM0: a bit of documetnation what this does?

lib/timeline/component/GraphGroup.js
```diff
@@ -50,35 +55,37 @@ GraphGroup.prototype.setItems = function(items) {
   }
 };
 
+GraphGroup.prototype.getItems = function () {
```
AlexDM0: does this need a getter? can't we just access the itemsData directly? Why the abstraction?

PR #1501, opened 2015-12-10
Don't use SVGElement.offsetHeight/Width

ttjoseph:
Fixes DOM API deprecation warning in Chrome.
Please see: https://www.chromestatus.com/features/5724912467574784


AlexDM0:
Hi,

We cannot accept pull requests to the master branch, can you retarget the develop branch?

Additionally, I think a lot of the things you have changed are DOM elements as opposed to SVGElements. Should those be altered as well?

Regards

PR #1505, opened 2015-12-10
Use SVGElement.getBoundingClientRect() instead of .offstHeight

ttjoseph:
Fixes DOM API deprecation warning in Chrome.
Please see: https://www.chromestatus.com/features/5724912467574784

This is an improved version of PR #1501 that only actually involves an SVGElement.


AlexDM0:
Thanks!

PR #1544, opened 2016-01-04
fixes positioning of colorpicker BUT ALSO HAS A BUG

felixhayashi:
@AlexDM0 not sure but for some reason, once the colorpicker is open, the buttons won't work correctly and the huecircle jumps back.

However the PR fixes the positioning issue (see screenshots below).

@AlexDM0 could you take over at that point please?

### DIFF
--- a/lib/network/css/network-colorpicker.css
+++ b/lib/network/css/network-colorpicker.css
@@ -1,14 +1,17 @@
 
 div.vis-color-picker {
   position:absolute;
+  top: 0px;
+  left: 30px;
   margin-top:-140px;
   margin-left:30px;
-  width:293px;
-  height:425px;
+  width:310px;
+  height:444px;
+  z-index: 1;
   padding: 10px;
   border-radius:15px;
   background-color:#ffffff;
-  display:none;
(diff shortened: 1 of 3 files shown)


AlexDM0:
Hi Felix, 

I'll see what i can do!

Thanks!

AlexDM0:
Hi Felix,

Could you add an example of when it does not work? The example of the vis website works fine!

Cheers

AlexDM0:
nevermind... rebuilding helps :)

AlexDM0:
Hi Felix,

I've fixed the bug on the develop branch. The problem was that you insert the color picker into a div that has a click listener for said color picker :).

I've added an on close callback to restore the onclick method and removed it on showing the color picker.

Thanks for the pull!

Cheers

felixhayashi:
> I've fixed the bug on the develop branch. The problem was that you insert the color picker into a div that has a click listener for said color picker :).  I've added an on close callback to restore the onclick method and removed it on showing the color picker.

Ahh that's what was going on. Thanks for taking care of it!

-Felix

PR #1583, opened 2016-01-18
Point main field in package.json to non-minified script

sapegin:
I think a bad practice to use minified files as a “main” script and it makes life of Webpack users more difficult because we have to use hacks like this:


josdejong:
Fine with me, thanks.

sapegin:
:+1:

PR #1637, opened 2016-02-06
Use Axis Labels for Tooltip

jacklightbody:
Simple change to show the axis labels on the tooltip instead of x,y,z. Even though the tooltip html is modifiable, seems like this should be the default.


josdejong:
Thanks, good idea. I cannot accept your PR on the master branch. Since it's just 3 lines of code I've already improved this in the `develop branch` (see https://github.com/almende/vis/commit/1e1f59a7a4343f74e2ef4191ef57cda614c3742c).

PR #1655, opened 2016-02-14
Use parseFloat instead of Number.parseFloat

ttjoseph:
Internet Explorer does not support the ES6 Number.parseFloat (see http://caniuse.com/#feat=es6-number). Instead, substitute the regular parseFloat method.


josdejong:
Thanks for the fix @ttjoseph !

PR #1656, opened 2016-02-15
Allow partial edge editing without reconnecting the nodes

alessiodm:
Add the boolean option `editEdgeNodes` to the `Edge` and the
`ManipulationSystem`, in such a way we can just edit link properties
in editMode without necessarily touching the nodes it connects.

Given that the property is both global at the manipulation level
and local per-edge, we should have enough flexibility for
different use-cases.

The behavior fallbacks to the current one in case nothing is set.

### DIFF
--- a/docs/network/edges.html
+++ b/docs/network/edges.html
@@ -128,6 +128,7 @@ var options = {
       opacity:1.0
     },
     dashes: false,
+    editEdgeNodes: undefined,
     font: {
       color: '#343434',
       size: 14, // px
@@ -325,6 +326,12 @@ network.setOptions(options);
                 repeated until the distance is filled.
                 <i>When using dashed lines in IE versions older than 11, the line will be drawn straight, not smooth</i>.
             </td>
+        </tr>
+        <tr>
+            <td>editEdgeNodes</td>
+            <td>Boolean</td>
+            <td><code>undefined</code></td>
+            <td>When true, the edge will offer the option to reconnect to different nodes in edit-mode.</td>
         </tr>
         <tr class='toggle collapsible' onclick="toggleTable('optionTable','font', this);">
             <td><span parent="font" class="right-caret"></span> font</td>
--- a/docs/network/manipulation.html
+++ b/docs/network/manipulation.html
@@ -102,6 +102,7 @@ var options = {
     addEdge: true,
     editNode: undefined,
     editEdge: true,
+    editEdgeNodes: true,
     deleteNode: true,
     deleteEdge: true,
     controlNodeStyle:{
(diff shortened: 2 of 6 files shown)


AlexDM0:
Hi,

I'm not sure I agree with putting manipulation options in the edges. All you're doing is calling the callback on the edge without changing anything.. I assume you'd make some kind of popup for the edge with more things to tweak? Why is this better than using a custom popup and an on click event?

Setting this per edge makes for a very unintuitive user interface, as edges are similar and they don't behave the same anymore. Sure you can layout them differently but I think you're merging your use case in with the vis functionality. 

Regards

alessiodm:
Hi Alex,

   thank you for your prompt reply. I might have missed an easier solution for what I'm trying to achieve, in case I apologize in advance. But I'll try to explain.

This is what I was struggling with: I wanted to edit just a property of the edge (e.g., the label), but entering in edit mode and returning from the callback forced me to reconnect the nodes anyway. That step is not necessary and forces the user to always perform an action (i.e., reconnecting a node with the control nodes) that is (possibly) semantically useless for the specific application.

Given that, I wanted an option to skip the phase of "node-reconnection" every time I enter in edit mode for the edge, allowing just my "popup" to save the properties I needed and that's it. But I didn't find an easy way to do that.

So I initially introduced the `editEdgeNodes` option only at the `ManipulationSystem` level. But thinking broader about it, that global option might be limiting in case we want to reconnect specific edges, but not others. With a global option we force either to always have node-reconnection or never. Adding the same property on the edge (with an 'overriding' purpose) allows to have fine-grained control to the editing of single edges if the use-case needs it.

Let me know if I might be of more help. Thanks.

AlexDM0:
Hi,

Why not ignore the manipulation system and use the click event to get an edge id, then get the edge data from the dataset, edit it through a popup and update the dataset. 

That seems to do exactly what you want. 

Regards

alessiodm:
Hi Alex,

thank you for your feedbacks. I'm not sure I've understood what you are suggesting. Are you suggesting to call the `setData` method on the `Network` every time I change the property of an edge?
In that case, I'm basically forcing a re-construction of the network graph completely. That might work, but it causes the network to stabilize again (I have the `physics` enabled), and I really would like to apply a very punctual change only on the selected edge.
Is there any other method to apply a change only on the specific edge (e.g., changing its color) without causing the network to rebuild or using `setData(...)` on it again avoiding a subsequent stabilization?
Thank you again for your time and your kindness.

Best,
Alessio

AlexDM0:
Hi,

No that's not what I mean. I suggest you take a look at the dynamic data examples on our website and enter the wonderful world of the vis datasets. 

Cheers

alessiodm:
OMG I totally missed the wonderful world of the `DataSet`s feature! Thank you so much Alex!
I guess this PR is not useful then; it would be just extending the `ManipulationSystem` but for a purpose that using `DataSet`s we can already fulfill.

AlexDM0:
Have fun :)

PR #1714, opened 2016-03-07
Allow bold font in network node labels

vViktorPL:
For some reason I needed to display bold labels in network nodes not only on hover and I didn't find such feature.


AlexDM0:
Hi,

You need to add this to the option validator list as well.

Regards

PR #1729, opened 2016-03-15
Add rtl option in timeline 

yotamberk:
This completes "A horizontalOrientation options to support a right-to-left timeline." except you can add `options.rtl = true/false/undefined`


josdejong:
Thanks @yotamberk, nice job! I've merged your PR.

And sorry for the late reply.

PR #1796, opened 2016-04-14
Fixed updating directions of the arrows with the string type

abxu:
Fixed a small bug in the update of the directions of arrows with the String type.

If I have A -> B, and I do :
```javascript
edges.update({id: "1", from: "A", to: "B", arrows: "from"});
```
for A <- B but the result was A <-> B.


AlexDM0:
Thanks a lot! You're absolutely right!

PR #1800, opened 2016-04-15
BugFix - Network.options.nodes

SonIcco:
As the documentation on vis.js nodes can be styled with align: left|center

allOptions does not include this option and throw an error on option validation.


AlexDM0:
Thanks!

PR #1802, opened 2016-04-17
Create Export.js file

hadjloo:
Add ability to save current view of canvas

### DIFF
--- a/lib/export.js
+++ b/lib/export.js
@@ -1,6 +1,0 @@
-function ExportCanvasToPng(canvasId, downloadLinkId, exportFileName) {
-    var canvas = document.getElementById(canvasId);
-    var dataURL = canvas[0].toDataURL('image/png');
-    document.getElementById(downloadLinkId).href = dataURL;
-    document.getElementById(downloadLinkId).download = exportFileName;
-},
--- a/lib/export.js
+++ b/lib/export.js
@@ -1,0 +1,6 @@
+function ExportCanvasToPng(canvasId, downloadLinkId, exportFileName) {
+    var canvas = document.getElementById(canvasId);
+    var dataURL = canvas[0].toDataURL('image/png');
+    document.getElementById(downloadLinkId).href = dataURL;
+    document.getElementById(downloadLinkId).download = exportFileName;
+},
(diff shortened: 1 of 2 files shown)


AlexDM0:
Hi,

I'm not sure why we would need this actually. Why would we want to include this? 

If you feel there is a solid usecase then I'd imagine this to be in the network code, exporting the network and we'd need documentation. 

Regards

hadjloo:
As a network admin, there is always a need to have a quick look of network map, to use it in presentations or to describe it to managers, so having a feature to export the network map will come handy

for documentation this is the sample case

1 - Add a link on page for import, it should have a selector, either css class or id
2 - call **ExportCanvasToPng()** with these parameter
 2.1 - **canvasId**: selector of network map canvas
 2.2 - **downloadLinkId**: selector of download link
 2.3 - **exportFileName**: file name to download

AlexDM0:
Why not network.exportToPNG()?

hadjloo:
That is not a bad idea, but actually I think it could be used in other libraries too. I'm not sure in other libraries it could be counted as a valuable feature,but it is better to have a universal feature which is consistent in all libraries,  instead of a feature that is available in one library and missed in others

AlexDM0:
Well we are ONE library ;). There are many things different from others. Specifically, vis is not a util lib. 

Anyone can google and find this method as stackoverflow has it. Anyone who needs it can write it themselves or copy it over. Supporting this is actually more work than having people who need it Google, hence my question of do we need it. If we add it, next feature requests are: can we export it to jpg, pdf, vector, can we put in custom sizes etc. is that really worth it? The aim for vis is to be an interactive network. Not an offline image making tool. There are plenty of those. 

Regards

hadjloo:
If the problem is supporting other formats, I agree to not include it in library, otherwise have a tiny export feature could come handy, all in all, I agree it can be only added in Network lib

AlexDM0:
Unless you want to commit to expanding the feature based on feedback Id accept it if integrated in network. If not, then I'd rather not commit to an export feature that won't grow past the initial commit.

hadjloo:
I moved the function to network library, I'll close this, and create another pull request

hadjloo:
I merges the pull request

AlexDM0:
Hi, I'm not sure we need to supply the canvas ID when its in the network? You'll also need to add it to the documentation.

PR #1822, opened 2016-04-25
reset switched{Day,Month,Year} in next function, not in isMajor function

herregroen:
resetting these variables inside the isMajor function caused subsequent calls to falsely report not being major labels.

This lead to problems in the TimeAxis _repaintLabels function where nextIsMajor is defined by calling isMajor, which resets switched{Day,Month,Year} and causes the isMajor of the next loop to always return false.


AlexDM0:
Thanks! Sorry for the late reply!

PR #1849, opened 2016-05-14
timeline: make end optional on range

raphink:
When setting up ranges, `end` is mandatory. It would be useful if it could default to the current date.


raphink:
@brichbe the reason I'd like to have this is because this would allow to just ommit it when using an external data file in JSON. Not having this features means you have to use `new Date()` in Javascript so you cannot rely on a JSON data file.

PR #1862, opened 2016-05-19
Set content width to avoid overflow (Fix #1853)

### LINKED ISSUE #1853: timeline: set width of bar
In order to place content in bars, `timeline` pushes `vis-item-content` to the right with things like:

```html
<div class="vis-item-content" style="left: 564.664px;"
```

When using large content, this leads to the content overflowing the div:

screenshot from 2016-05-14 15-27-10

By adjusting the width, this can be fixed:

```html
<div class="vis-item-content" style="left: 564.664px;width: calc(100% - 564.664px);"
```

screenshot from 2016-05-14 15-37-10

raphink:
This is the same as https://github.com/almende/vis/pull/1853, except on the develop branch.

@AlexDM0 is that good for you?

### DIFF
--- a/dist/vis.js
+++ b/dist/vis.js
@@ -22462,6 +22462,7 @@ return /******/ (function(modules) { // webpackBootstrap
           this.dom.content.style.right = contentStartPosition + 'px';
         } else {
           this.dom.content.style.left = contentStartPosition + 'px';
+          this.dom.content.style.width = 'calc(100% - ' + contentStartPosition + 'px)';
         }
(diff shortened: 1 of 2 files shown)


AlexDM0:
Hi,

You'll need to add it to the file in the lib folder, the dist is rebuilt all the time so it would be overwritten.

Regards

raphink:
There, I made the change in both.

AlexDM0:
thanks!

PR #1872, opened 2016-05-25
Add specific class with value to major gird lines

leninzprahy:
It will be possible to highlight specific lines, for example in rpercentage chart 100%, by adding `.vis-val-100 {}` to css...

### DIFF
--- a/lib/timeline/component/DataAxis.js
+++ b/lib/timeline/component/DataAxis.js
@@ -392,7 +392,7 @@ DataAxis.prototype._redrawLabels = function () {
       }
       if (this.master === true) {
         if (isMajor) {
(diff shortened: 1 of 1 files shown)


AlexDM0:
This is what the timebars are for. Check the custom time examples.

Regards

leninzprahy:
I think, that timebars are for vertical lines, but this PR only add class to horizontal lines in grid.
For example in Graph2d

AlexDM0:
I see, @ludost what do you think? We cannot accept pull requests on the master branch, only develop.

Cheers

leninzprahy:
I understand. Should I to send the PR to the develop branch?

AlexDM0:
I have the feeling there should be a better solution for this usecase.. @ludost is currently maintaining the graph2d so I'll let him decide.

Regards

PR #1884, opened 2016-06-03
Add the source data to Point3d objects in Graph3D

TRManderson:
This allows more complex tooltips, using information sent from the server + the source object

### DIFF
--- a/examples/graph3d/11_tooltips.html
+++ b/examples/graph3d/11_tooltips.html
@@ -24,6 +24,11 @@
 
       // Create and populate a data table.
       data = new vis.DataSet();
+	  var extra_content = [
+		'Arbitrary information',
+		'You can access data from the point source object',
+		'Tooltip example content',
+	  ];
 
       // create some nice looking data with sin/cos
       var steps = 5;  // number of datapoints will be steps*steps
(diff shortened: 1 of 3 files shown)


lib/graph3d/Graph3d.js
```diff
@@ -589,6 +590,7 @@ Graph3d.prototype._getDataPoints = function (data) {
       point.x = data[i][this.colX] || 0;
       point.y = data[i][this.colY] || 0;
       point.z = data[i][this.colZ] || 0;
+      point.source = data[i];
```
mojoaxel: About the naming: I personally would prefer `point.data` over `point.source`

mojoaxel:
@TRManderson Could you maybe extend the [graph3d tooltips example](/almende/vis/blob/develop/examples/graph3d/11_tooltips.html) to use the new data? In any case this new feature should be documented and explained [here](/almende/vis/blob/develop/docs/graph3d/index.html). Thanks!

TRManderson:
Updated to include docs and example change. I couldn't come up with a compelling extension to the tooltips example, but I think I demonstrated the functionality non-horribly.

examples/graph3d/11_tooltips.html
```diff
@@ -34,10 +39,10 @@
           var z = custom(x,y);
           if (withValue) {
             var value = (y - x);
-            data.add({x:x, y:y, z: z, style:value});
+            data.add({x:x, y:y, z: z, style:value, extra: extra_content[(x*y) % extra_content.length]});
           }
           else {
-            data.add({x:x, y:y, z: z});
+			data.add({x:x, y:y, z: z, extra: extra_content[(x*y) % extra_content.length]});
```
mojoaxel: Can you please change the ident every where from "tab" to "spaces". (I also don't like it, but ithe the projects standart.)

TRManderson:
Sorry about that - dodgy vim config meant I was typing tabs when I thought I had spaces. I've rebased develop with the changes requested.

mojoaxel:
@TRManderson Thanks!

Issue #507, opened 2014-12-11
Feature request - animating edges in a network


anzeljg:
Would it be possible (is it already) to create a computer network (using appropriate images for nodes: e.g. computer, router, switch, firewall, internet, etc.) and then show the animation how a data packet travels through the network. E.g. data packet (could be the circle) going in animated way along the edges from node1 (e.g. computer) to node2 (e.g. switch) to node3 (e.g. firewall) to node4 (e.g. Internet).

I know that it is possible to use images/icon and I've successfully created a network using OSA icon library (http://www.opensecurityarchitecture.org/cms/library/icon-library) based on Tango icons. All I really want to know is, if there is a way to simulate network traffic in an animated way?

Kind regards, Gregor

AlexDM0:
Hi,

Over the last year a lot of feature requests have been made. We have just introduced our new website which has a list of the requested features. We have placed this request on that list. 

The list can be found here:
http://visjs.org/featureRequests.html

An explaination of the new system can be found here:
http://visjs.org/blog.html#New\ website\ for\ vis\.js\!

I would like to stress that this does not mean we abandon this request. Discussion here will continue if needed on this feature but we will close it to keep our Github issue page more of a bug-todo list.

Future feature requests will still be made here and then added to the website by us.

Regards,

Alex

anzeljg:
Thank you for informing me about that, but on that web page my issue #507is linked to issue #530.

Kind regards, Gregor

AlexDM0:
My bad, i'll fix it later today. Thanks for letting me know!

Regards

AlexDM0:
fixed it :)

edvineshagh:
Hi,

I added a proptype method that animates the network traffic; but, I don't know where to upload the draft version of m module.  It isn't really ready to integrate with vis.

The module has pre/post frame and animation handler.  For example, I used pre/post animation to update the node value based on traffic size.
The size of the circle is determined by the traffic size from each node.

![trafficannimation](https://cloud.githubusercontent.com/assets/4054499/8289887/041a53d8-18d4-11e5-9363-731673ea1c63.gif)

I have two issues with this implementation:
<ul>
   <li> it does not render well in IE</li>
   <li> it does not use vis internal animation engine</li>
</ul>
Despite these limitations; someone may still find it useful; or want to add/update to it.
Where can I post this draft version of the code:

Here is the usage:
<pre>

[...]

PR #1907, opened 2016-06-14
Fixed getPoint for same node edges

aanm:
If an edge was point out to the same node, the `getPoint` returns the wrong coordinates.

Signed-off-by: André Martins <aanm90@gmail.com>

### DIFF
--- a/lib/network/modules/components/edges/BezierEdgeDynamic.js
+++ b/lib/network/modules/components/edges/BezierEdgeDynamic.js
@@ -133,8 +133,16 @@ class BezierEdgeDynamic extends BezierEdgeBase {
    */
   getPoint(percentage, viaNode = this.via) {
     let t = percentage;
-    let x = Math.pow(1 - t, 2) * this.fromPoint.x + (2 * t * (1 - t)) * viaNode.x + Math.pow(t, 2) * this.toPoint.x;
-    let y = Math.pow(1 - t, 2) * this.fromPoint.y + (2 * t * (1 - t)) * viaNode.y + Math.pow(t, 2) * this.toPoint.y;
+    let x, y;
+    if (this.from === this.to){
+      let [cx,cy,cr]  = this._getCircleData(this.from)
+      let a = 2 * Math.PI * (1 - t);
+      x = cx + cr * Math.sin(a);
+      y = cy + cr - cr * (1 - Math.cos(a));
+    } else {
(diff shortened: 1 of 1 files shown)


mojoaxel:
@aanm It looks good but I can't figure out an simple example to test the changed `getPoint` function. Can you provide some code for testing this issue?

aanm:
@mojoaxel Sure, can you please tell me where do you want me to put the test? Do you want it here or in a file under `vis/test`? If a file under `vis/test` which one?
Thanks

mojoaxel:
@aanm Most peole create a [jsbin](http://jsbin.com/?html,output) or [sfiddle](https://jsfiddle.net/). If it's an short example just just post it here. 
The code looks good but I want to test it before I merge it. What settings are needed that this function is executed? That's why I need an example.

aanm:
@mojoaxel oh okay. so the problem is simple, I was following this thread https://github.com/almende/vis/issues/507 and when I used the code provided by @edvineshagh I saw that it didn't work for edges that were pointed to the same node. That's why I figure out the getPoints function had a bug for the edges that were pointed out to the same node. Can you test based on this?

mojoaxel:
Today I found the time to look into this again. I still don't get it. :disappointed:

I tried to create an example with a node having an edge from and to itself. Also enabled default dynamic smoothing and arrows. As far as I can see the `getPoint` is never called on edges that have the same from and to. I can' test this.

@aanm Please provide a simple example that shows a use case for this pull request. Thx!

aanm:
@mojoaxel There you go https://jsfiddle.net/2xc5r20j/ :-)
Like I said, the thread #507 is the one that calls the `getPoint` function. The animation code is a little buggy but the gist is there.
sorry I was planing on provide an example but I hadn't time to create it.

yotamberk:
@mojoaxel What's the status on this PR?

mojoaxel:
> What's the status on this PR?

The code looks good but I hab problems creating an example to debug the changes. I think we can merge it, but I would feel better if I actually understood when this is needed.

mojoaxel:
Shi**. This should not go into the master branch...I'll revert this.

mojoaxel:
@aanm Sorry, but I had to revert the changes because they were made to the master branch. If you still want them to be added please create a new PR to the develop branch. Thx.

aanm:
@mojoaxel this way? #2161

PR #1922, opened 2016-06-22
Select the closest edge, not just any nearby

root9b:
This helps quite a bit for complex graphs with dense connections, where many edges are packed into a small space. With the default algorithm, any edge within 10 of the mouse will get selected, so there are often edges you are unable to select at all even though you can see them, because others are nearby. With this patch, you can be much more precise with where you are clicking and which edge you are selecting.

### DIFF
--- a/lib/network/modules/SelectionHandler.js
+++ b/lib/network/modules/SelectionHandler.js
@@ -219,15 +219,32 @@ class SelectionHandler {
    * @returns {undefined}
    */
   getEdgeAt(pointer, returnEdge = true) {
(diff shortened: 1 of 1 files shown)


mojoaxel:
The issue descriped is realy easy to reproduce with the [worldCup example](http://visjs.org/examples/network/edgeStyles/smoothWorldCup.html).

mojoaxel:
I tested it. Looks Good!

mojoaxel:
@root9b Thanks very much! Nice improvement! Sorry that you had to wait so long for this to get merged.

PR #1931, opened 2016-06-29
Restore toolbar if delete callback not called

jbrobst:
When the function given for the manipulation.deleteNode or manipulation.deleteEdge options did not call the callback provided (due to not wanting to delete after all), the manipulation toolbar was not being redrawn after _clean because the call to showManipulatorToolbar only occurred within the callback. To fix it I just moved the call outside of the callback. I also moved the startSimulation emit because the former logic did it if even if nothing was deleted, but I'm not completely sure this is necessary.


jbrobst:
I realized you're probably supposed call the callback with null. The manipulation documentation is kind of confusing about this, because the statement "To cancel, call the callback function with null as argument or without arguments" is only present under the entry for editEdge and does not indicate this is the case for deleteNode and deleteEdge as well. The delete entries both say "These are the items that will be deleted if the callback is performed," implying that the callback does not need to be performed.

PR #1944, opened 2016-07-05
fix #1723 (Strange behaviour in the time axis bar when hiddenDates setted)

### LINKED ISSUE #1723: Strange behaviour in the time axis bar when hiddenDates setted
Hi, 

here I am again =) 

When I set the scale to "day" and hiddenDates to 

```
[
      {start: '2016-01-01 00:00:00', end: '2016-01-01 08:00:00', repeat: 'daily'},
      {start: '2016-01-02 17:00:00', end: '2016-01-03 08:00:00', repeat: 'daily'},
      {start: '2016-01-02 00:00:00', end: '2016-01-04 00:00:00', repeat: 'weekly'}
]
```

The first day in the time asix scale is not displayed even there's the necessary space.

Here's my demonstration:
http://jsbin.com/hipugus/edit?html,output

in this case even if there is space the 15 day label is not displayed.

Without the  hiddenDates param there's no problem.

Thank you! 
Nikola

chiarishow:
Hi,

this is a fix for the 1723 issue (link https://github.com/almende/vis/issues/1723)

this strange behaviour happened because the hidden time, before the timeline range, was not excluded from the calculation.  

Also the same thing happened with time after the timeline range. 

this can potentially fix other visualization issues using timeline with hidden time, like having an item dragged before the starting range. 

Thanks,
Nikola

### DIFF
--- a/lib/timeline/DateUtil.js
+++ b/lib/timeline/DateUtil.js
@@ -378,7 +378,7 @@ exports.getHiddenDurationBeforeStart = function (hiddenDates, start, end) {
     var startDate = hiddenDates[i].start;
     var endDate = hiddenDates[i].end;
 
-  if (startDate >= start && endDate <= end) {
+    if (startDate >= start && endDate <= end) {
       duration += endDate - startDate;
     }
   }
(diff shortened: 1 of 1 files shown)


mojoaxel:
Looks good. 
@chiarishow Thanks!

PR #1946, opened 2016-07-06
fix known security issue with handlebars version

doubleoevan:
This one line pull request addresses a security issue with handlerbars version exposed by http://retirejs.github.io/retire.js/


josdejong:
Thanks for updating the handlebars example.

Issue #1602, opened 2016-01-25
Bundle without external dependencies


vlasenko:
Distributed vis.min.js inside bower package is non-optimal, it includes hammer.js and moment.js inside it. Why not simply declare them as dependencies inside bower package, this will greatly reduce vis.min.js size. This is especially important for projects that, for example, utilize moment.js already, which is the case pretty often.

josdejong:
Thanks for your suggestion.

To keep things simple we provide a "full" bundle with everything included which works right out of the box in all cases. It's indeed often the case that you already use moment, or that you only need one of the visualizations. To optimize for that, you can simply create your own bundle, or integrate bundling right away with the build script of your application.

https://github.com/almende/vis#custom-builds

vlasenko:
While this approach is fine when no package manager used. For bower package it seems very strange and not convenient for the developers that use you package. If dependencies were declared in the bower.json of vis.js on hammer.js and moment.js, it hadn't do any harm to vis.js but only good to its users. And this is the de-facto approach for bower packages.

Custom builds might still be needed, but they are needed when you want to customize build of vis.js itself. For the issue I refer to, the problem is not about customizing vis.js itself, the problem is that you don't want to have quite a big size increase, because of vis.js dependencies. This is unexpected and can be easily fixed.

josdejong:
It will be convenient for bower users to have a build of vis.js excluding moment.js and hammer.js. I don't recommend using bower at all, but well, there are still a lot of people using it. 

We have plans to generate bundles containing individual visualizations, when we tackle that we can see if we also generate bundles excluding hammer.js/moment.js. 

In general, the JS community is moving away from pre-bundled libraries. We need a bundle step in all our applications anyway, and bundling everything right from the source allows bundlers to efficiently bundle stuff without duplicates and exclude non-used parts of libraries (see for example [rollup](http://rollupjs.org/)).

mojoaxel:
I also like the idea of the "all-inclusive-bundle" for simple webpages. But as mentioned before it is a pain if you are using a bundler like webpack together with bower.

> .. or integrate bundling right away with the build script of your application.

To make this possible please (re-)add the `lib` folder to the bower delivery. Than it would be possible to require a suitable `index*.js` file in your project and let the bundler figure out the dependencies.

> In general, the JS community is moving away from pre-bundled libraries.

[...]

PR #1961, opened 2016-07-14
improvements of the bower delivery

mojoaxel:
There are some improvements to the bower delivery:
* removed `package.json` and `gulpfile.js` because if you really want to build your own dist files, bower is maybe not the right package-manager to do this!?
* reverted https://github.com/almende/vis/pull/906/commits/856348d108d8eacd0726637239373d18417f9e0e by re-adding the `lib` folder to the bower delivery (see https://github.com/almende/vis/pull/906#commitcomment-18069283). This is necessary if you want to use the still existing `index*.js` files e.g. with webpack. (see https://github.com/almende/vis/issues/1602#issuecomment-232595261).
* added `keywords` because they are [recommended](https://github.com/bower/spec/blob/master/json.md#keywords).

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -8,11 +8,25 @@
     "type": "git",
     "url": "git://github.com/almende/vis.git"
   },
+  "keywords": [
+    "vis",
+    "visualization",
+    "web based",
+    "browser based",
+    "javascript",
+    "chart",
+    "linechart",
+    "timeline",
+    "graph",
+    "network",
+    "browser"
+  ],
   "ignore": [
+    "gulpfile.js",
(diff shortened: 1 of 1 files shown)


mojoaxel:
I just realized that just including the index*.js files will not work, because there are missing external dependencies. Some are not (yet) availible with bower:

* emitter-component (not availible an bower)
* keycharm (not availible an bower)
* propagating-hammerjs
* hammerjs 
* moment

Therefor including the `lib` folder and the `index*.js` files makes no sense at this point. I will create a new PG to fix this.

Issue #1866, opened 2016-05-20
Why are there no dependencies


Spocer:
Hi,

While going over modules for my project, I noticed that Vis has no listed dependencies in the package file. This seems odd as I thought and modules or other packages which are packaged with Vis should be noted in the dependencies section, not the devDependencies section. This seems to be the only module I use which is like this.

I know that Vis depends on Hammer, so why for instance is that only listed in devDeps and not deps?

Thanks.

AlexDM0:
Hi,

Thats because the compiled version of vis does not have dependencies. 

Regards

Spocer:
This seems like an odd practice. So far, this is the only library I have come across which distributes itself in this manner. Is there a reason for pre-packaging Vis in this fashion? I see in some older versions (4.8.x) this was not the case. What necessitated this change?

AlexDM0:
As far as I know it has always been this way. Reason being that the dist just works.

PR #1967, opened 2016-07-15
make bundleing with webpack possible

mojoaxel:
I moved the following dependencies from `devDependencies` to `dependencies`:
* emitter-component
* moment
* propagating-hammerjs
* hammerjs
* keycharm

I use the timeline component in a project. If I include `dist/vis-timeline-graph2d.min.js` i have the problem that `moment` and `hammerjs` are included twice.

So the best way ist to include `/vis-timeline-graph2d.js` with webpacks babel-loader and let the module-bundler figure out the rest. So I don't want to build my own `dist` version, but want to resolve the dependencies "on the fly". This is also recessary for things like [tree shaking](https://medium.com/@Rich_Harris/tree-shaking-versus-dead-code-elimination-d3765df85c80#.eo0uxcn8t) and other fancy stuff.

The only think that has to be changed for that ist to install the needed dependencies together with vis itself. Voila!

I tried it and it works with webpack. The vis code now is only half the size!


mojoaxel:
this "fixes" #1866

yotamberk:
Hi @mojoaxel, 
Can you fix the conflicts and submit the fix?
I want to add this in =)

mojoaxel:
> Can you fix the conflicts and submit the fix?

done.
@yotamberk Can you please make sure that this fix does not break anything. Thanks!

yotamberk:
It seems to be working fine!
I'm merging it =)

mojoaxel:
Yeah!! Thanks!

Issue #1156, opened 2015-08-03
dragging between groups with scrolling


datom:
Hi,

i have many groups and want to move a item from one group to another. This works fine until the window is big enough to display all the groups at once. But when the window is smaller and i want do move a item from one group to another, its no more possible because its always scrolling and no more dragging.

For example:
http://visjs.org/examples/timeline/other/groupsPerformance.html

> run: timeline.setOptions({'height': '100px'})

now try to move i item between the groups. it always scrolling and not moving the item.

Dragging horizontal works right but vertical is not as it should be.

Regards

josdejong:
Thanks, good point that should be fixed.

Besides that, it would be nice if the timeline automatically moves horizontally/vertically when you drag an item close to the border of the window, in case you want to drag it to some position currently not visible.

datom:
yes that would be nice. i think on touch devices it would be very helpful

datom:
any news when this will be fixed?

josdejong:
I'm not sure. We're kind of busy ourselves, I can't make an estimation. 

Anyone can jump in to help out here though, it's an open source project after all :)

thomas-barone:
Sorry for the butchered commit history!

I had the same issue, I've created a pull request #1976 for the solution that worked for me

yotamberk:
This issue is resolved with  #1976 or with using `option.verticalScroll`

mojoaxel:
> This issue is resolved with #1976 or with using option.verticalScroll

Sorry, but this still doesn't work.

PR #1976, opened 2016-07-20
Fix issue #1156: Vertical scrolling when dragging

thomasbarone:
Added a flag to check whether an item in the Itemset is currently being dragged, this is used to stop the timeline from scrolling when an item is being dragged.

A better solution might be to change some of the drag event handlers, and stop certain mouse events from propogating, but this fix worked for me :)


mojoaxel:
Works fine! 
Thanks @thomasbarone !

PR #1986, opened 2016-07-22
implement a smarter zoom in/out that gets back to the original window

daattali:


mojoaxel:
It's not much but It really makes zooming more accurate.
@daattali Thanks!

mojoaxel:
manually merged to `develop` with adedd6a853c33184ab7c819eea1815a791267e8c

PR #1987, opened 2016-07-24
Resolve issue #1857 Fixed node positioning with improvedLayout:true

### LINKED ISSUE #1857: Nodes as legends example is not aligning legend
nodes not aligned

Whenever I reload I get different positions. Do you know why is this caused?

Example is [here](http://visjs.org/examples/network/exampleApplications/nodeLegend.html).

Stexxen:
Resolve #1857 
When improvedLayout was true, perturbing the nodes was being performed on all nodes regardless of whether they were set as fixed.

Using http://visjs.org/examples/network/exampleApplications/nodeLegend.html as an example
_before patch_
image

_after patch_
image


mojoaxel:
@Stexxen Thanks!

PR #2012, opened 2016-08-11
Remove single node cluster restriction with option

Stexxen:
Default functionality remains the same, but restriction can be lifted with a config option.

Also created Example to demonstrate new option and clusters of clusters.


mojoaxel:
replaced with #2013

PR #2013, opened 2016-08-11
Removed restriction to allow clusters of a single node.

Stexxen:
Default functionality remains the same, but restriction can be lifted with a config option.
Also created Example to demonstrate new option and clusters of clusters.

### DIFF
--- a/docs/network/index.html
+++ b/docs/network/index.html
@@ -1181,7 +1181,18 @@ var options = {
                     node
                     any way you want. This is also the style object that is provided in the processProperties function
                     for
-                    fine tuning. If undefined, default node options will be used.
+                    fine tuning. If undefined, default node options will be used.<br/><br/>
+                    Default functionality only allows clustering if the cluster contains 2 or more nodes. To allow clustering of single nodes you can use the <code>allowSingleNodeCluster : true</code> property.
+<pre class="prettyprint lang-js">
+var clusterOptions = {
+    joinCondition: function (potentialNodeOptions) {
+        return potentialNodeOptions.cid === 1;
+    },
+    clusterNodeProperties: {
+        allowSingleNodeCluster: true
+    }
+};
+</pre>
                 </td>
             </tr>
             <tr><td id="event_clusterEdgeProperties">clusterEdgeProperties</td>
--- a/docs/network/index.html
+++ b/docs/network/index.html
@@ -1182,16 +1182,11 @@ var options = {
                     any way you want. This is also the style object that is provided in the processProperties function
                     for
                     fine tuning. If undefined, default node options will be used.<br/><br/>
-                    Default functionality only allows clustering if the cluster contains 2 or more nodes. To allow clustering of single nodes you can use the <code>allowSingleNodeCluster : true</code> property.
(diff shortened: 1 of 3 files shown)


Stexxen:
@mojoaxel thanks for merging my other pull.
This one is not a bug fix but a new feature, or rather the removal of a restriction. I've created an example, but don't think this pull entirely complies with misc/how_to_publish.md. Do you have any advice on what else I need to do?
Thanks,
S.

examples/network/other/clustersOfclusters.html
```diff
@@ -0,0 +1,77 @@
...
+        );
+        network.on('click', function (params) {
+            if (params.nodes.length == 1) {
+                if (network.isCluster(params.nodes[0]) == true) {
+                    network.openCluster(params.nodes[0]);
+                }
+            }
+        });
+        cluster();
+        cluster();
```
mojoaxel: One of this calls should be enough to show the new feature ;-)

mojoaxel:
@Stexxen Very nice! Why not!
Can you please add docs for the new option `clusterNodeProperties` [here](/almende/vis/blob/develop/docs/network/index.html#L1186). 
You just can commit changes in your `single_cluster` branch to update this pull-request.

After that I see no reason not to merge this.

mojoaxel:
@Stexxen I still think the new option `allowSingleNodeCluster` should be documented!?

mojoaxel:
@Stexxen Can you please add some documentation. If you can not do it please let us know so we can do it. Thanks!

Stexxen:
apologies, I will work on this today.

Stexxen:
@mojoaxel documentation now added, and removed a couple of the calls to cluster(). I've still left 3 in as the example is called clusterOfClusters after all ;-)

docs/network/index.html
```diff
@@ -1181,7 +1181,18 @@ <h4 id="optionsObject">Cluster methods options object</h4>
                     node
                     any way you want. This is also the style object that is provided in the processProperties function
                     for
-                    fine tuning. If undefined, default node options will be used.
+                    fine tuning. If undefined, default node options will be used.<br/><br/>
+                    Default functionality only allows clustering if the cluster contains 2 or more nodes. To allow clustering of single nodes you can use the <code>allowSingleNodeCluster : true</code> property.
+<pre class="prettyprint lang-js">
+var clusterOptions = {
```
mojoaxel: you should call it "options" to be coherent with the other examples

docs/network/index.html
```diff
@@ -1181,7 +1181,18 @@ <h4 id="optionsObject">Cluster methods options object</h4>
...
+    joinCondition: function (potentialNodeOptions) {
```
mojoaxel: The "joinCondition" is from a other example ans should be removed.

docs/network/index.html
Stexxen: ok

docs/network/index.html
Stexxen: "joinCondition" is not always optional. Thats why I chose to leave it in, so it is a complete example, but please confirm you want it removed?

PR #2034, opened 2016-08-21
timeline: use css style instead of delete image

mojoaxel:


mojoaxel:
@All Yes? No? Maybe so?

PR #2035, opened 2016-08-21
fixes BoxItem's initial width if content width is dynamic; fixes #2032

### LINKED ISSUE #2032: Timeline + Groups - Two cases of severe rendering issues
Using Timeline with Groups and stack=true (default).

## 1)
This issue took me a long time to figure out the cause...

**Severe rendering issues happen when we define `white-space: normal` for the class `.vis-item-content` and no fixed-width is defined.**

Basically, vis.js fails to calculate positions properly.

I made a modified version of the Groups demo:
http://jsbin.com/remukoqede/edit?html,output

Tip: because the demo creates random elements, some times (page reloads) it will be more clearly visible than others.

## 2)
When trying to center group labels in their containers, my first approach was this:

```css
.vis-labelset > .vis-label {
	display: table;
}
.vis-labelset > .vis-label > .vis-inner {
	display: table-cell;
	vertical-align: middle;
	text-align: center;
}
```

But this also was causing positioning issues for vis.js.
The workaround I found is using flexbox, despite its compatibility:

```css
.vis-labelset > .vis-label {
	display: flex;
	justify-content: center;
	align-items: center;
}
```

## 1+2)
In the configuration I'm building for a project I was getting an infinite amount of `WARNING: infinite loop in redraw?` and vis.js basically couldn't resize each group. You can see that it resizes pixel by pixel when timeline is being dragged in any direction.

http://jsbin.com/payugiyeyo/edit?html,output

mojoaxel:


mojoaxel:
Can please somebody review this. Maybe there is a better way to fix this?

yotamberk:
Seems to solve the problem. 
I think it is the legit place to put it too.

mojoaxel:
@yotamberk Thanks! You seem to be the "rtl" specialist here. So I think it is safe to merge ;-)

PR #1958, opened 2016-07-13
fixed minor issues on the timeline example overview page

mojoaxel:
fixes #1797 
fixes #1937

PR #2050, opened 2016-08-28
fixed timeline examples preview images

mojoaxel:
This reserves a change done in 2657ba15ac5a548a988dec577030bbc22a86e4bc because the preview images on the timeline example disappeared.


mojoaxel:
see #1958

Issue #972, opened 2015-06-12
Using Nodes.update() for Cluster


chase1991:
Is there any way to change the style of cluster nodes, for instance, using nodes.update()?

Thanks

AlexDM0:
No, this is not possible at the moment. It may be possible in future releases but for now I cannot promise anything.

Regards

chase1991:
thanks, Alex.

dragonzone:
The way to work around this is to open up the cluster, and re-cluster it with the style you want, but make sure when you re-cluster it that your cluster ID is not the same b/c you'll get a phantom link from https://github.com/almende/vis/issues/1055, which i being fixed. I've done this on my project.

mojoaxel:
Since 996298793e06e45c21d9d362fe603a1d8925367c there is a function `updateClusteredNode` for that.

fixed with #2055

Issue #1920, opened 2016-06-21
Update Cluster node properties


sarjandesai:
Is There any way of updating ClusterNode Properties? 

My requirement is to change the background and someother properties of the ClusterNode on click event of Network.

When i try  **update** on cluster node its converting it to normal Node. I tried to assign all cluster node related properties as well.

Sample Code :

```
var Nodes = network.body.nodes;
var xNodes = Object.keys(Nodes).map(function(k) { return Nodes[k] });// JSON to array
			
			for(var x = 0;x < xNodes.length;x++)
			{
				if(network.isCluster(xNodes[x].id))
				{
					nodes.update({ id: xNodes[x].id, label: xNodes[x].options.label,cid:xNodes[x].options.cid, shape: 'image', image: xNodes[x].options.image,color:{border: '#406897', background: '#efeded',highlight:{background:'gray',border:'blue'}}, shapeProperties: {useBorderWithImage:true}, clusterEdgeProperties : xNodes[x].clusterEdgeProperties, containedEdges : xNodes[x].containedEdges , containedNodes : xNodes[x].containedNodes, isCluster : true});
				}
				else
				{
					nodes.update({ id: xNodes[x].id, label: xNodes[x].options.label,cid:xNodes[x].options.cid, shape: 'image', image: xNodes[x].options.image,color:{border: '#406897', background: '#efeded',highlight:{background:'gray',border:'blue'}}, shapeProperties: {useBorderWithImage:true}});
				}
			}
```

mojoaxel:
duplicate of #972

mojoaxel:
fixed with #2055

PR #2055, opened 2016-08-30
Added getBaseEdge, getClusteredEdges updateEdge and updateClusteredNo…

Stexxen:
…de to allow option changes of the clustered edges as they are no longer the original base Edge

changing edge color is a good example.
i.e network.clustering.updateEdge(originalEdge.id, {color : '#aa0000'});
will now update all edges including the ones that were created when clustering.

Also when clicking on a clustered edge the event returns the clustered Edge id, which doesn't mean a lot.
This can be now converted back into the original edge with
network.clustering.getBaseEdge(clusteredEdge.id)

Similar with updateClusteredNode
This new method allows the options to be changed on a clustered Node.

i.e clicking on a clustered node gives its clustered node id.
Want to change the clustered image to a star then do this
 network.clustering.updateClusteredNode(clusteredNodeId, {shape : 'star'});

### DIFF
--- a/lib/network/modules/Clustering.js
+++ b/lib/network/modules/Clustering.js
@@ -15,12 +15,6 @@ class ClusterEngine {
     this.body.emitter.on('_resetData', () => {this.clusteredNodes = {}; this.clusteredEdges = {};})
   }
 
-  setOptions(options) {
-    if (options !== undefined) {
-
-    }
-  }
-
   /**
   *
   * @param hubsize
@@ -316,6 +310,9 @@ class ClusterEngine {
       // create the edge and give a reference to the one it replaced.
       let newEdge = this.body.functions.createEdge(clonedOptions);
       newEdge.clusteringEdgeReplacingId = edge.id;
+
+      // also reference the new edge in the old edge
+      this.body.edges[edge.id].edgeReplacedById = newEdge.id;
 
       // connect the edge.
       this.body.edges[newEdge.id] = newEdge;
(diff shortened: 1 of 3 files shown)


Stexxen:
@mojoaxel Just wanted to see if this is something that could be added? If so, I can create some examples and get docs updated as well. (Have to get the docs done for the other pull as well.)

Stexxen:
This would also probably resolve #972 and #1920

mojoaxel:
@Stexxen :tada: Wow! Wonderfull! Thank you for this nice Pull Request! 

Could you please add documentation an ideally a new example!? Than it would be my pleasure to merge this! :smile: 

Also feel free to comment on the issues you mentioned.

mojoaxel:
I have some comments on the code. I hope you don't mind!

lib/network/modules/Clustering.js
```diff
@@ -696,6 +693,71 @@ class ClusterEngine {
...
+  }
+
+  /**
+  * Get a stack of clusterEdgeId's (+base edgeid) that a base edge is the same as. cluster edge C -> cluster edge B -> cluster edge A -> base edge(edgeId)
+  * @param edgeId
+  * @returns {Array}
+  */
+  getClusteredEdges(edgeId) {
+    let stack = [];
+    let max = 100;
```
mojoaxel: Why not `this.body.edges.length` instead of the fixed max of 100?

lib/network/modules/Clustering.js
```diff
@@ -696,6 +693,71 @@ class ClusterEngine {
...
+    let counter = 0;
+
+    while (edgeId !== undefined && this.body.edges[edgeId] !== undefined && counter < max) {
```
mojoaxel: `while` loops are always a little dangerous because a small typo can easily cause infinite loops. Why not a `for` 0..max instead?

lib/network/modules/Clustering.js
```diff
@@ -696,6 +693,71 @@ class ClusterEngine {
     return stack;
   }
 
+  /**
+  * Using a clustered nodeId, update with the new options
+  * @param nodeId
```
mojoaxel: the `newOptions` parameter is missing :wink:

lib/network/modules/Clustering.js
```diff
@@ -696,6 +693,71 @@ class ClusterEngine {
...
+    var node = this.body.nodes[nodeId];
+    if (node !== undefined) {
+      node.setOptions(newOptions);
+      this.body.emitter.emit('_dataChanged');
+    }
+  }
+
+  /**
+  * Using a base edgeId, update all related clustered edges with the new options
+  * @param nodeId
```
mojoaxel: the `newOptions` parameter is missing :wink:

lib/network/modules/Clustering.js
```diff
@@ -696,6 +693,71 @@ class ClusterEngine {
...
+      edgeId = this.body.edges[edgeId].edgeReplacedById;
+      counter++;
+    }
+    stack.reverse();
+    return stack;
+  }
+
+  /**
+  * Get the base edge id of clusterEdgeId. cluster edge (clusteredEdgeId) -> cluster edge B -> cluster edge C -> base edge
+  * @param edgeId
```
mojoaxel: param should be named be `clusteredEdgeId`

lib/network/modules/Clustering.js
```diff
@@ -15,12 +15,6 @@ class ClusterEngine {
     this.body.emitter.on('_resetData', () => {this.clusteredNodes = {}; this.clusteredEdges = {};})
   }
 
-  setOptions(options) {
```
mojoaxel: Thanks for removing this!

lib/network/modules/Clustering.js
Stexxen: can you elaborate further?
this loop uses the new `edgeReplacedById` to move through the edges, a bit like a linked list.

lib/network/modules/Clustering.js
Stexxen: I was using the pre-existing findNode method as a template on how this kind of function should be written, (which contains the 'max' method). I agree an arbitrary value isn't ideal.

You may find this slightly counterintuitive (as i did) but `this.body.edges` is not an array but an object. At the time I also wondered why DataSet was not used to hold the internal representation of the data.

Stexxen:
Glad you like it :-) I'll fix the code docs and get on with the proper docs and an example.

lib/network/modules/Clustering.js
mojoaxel: I see! My mistake. Makes sense now that I actually look at it ;-)

mojoaxel:
@Stexxen Thanks very much. I'll try to test and merge it as soon as possible.

Stexxen:
@mojoaxel That the last update. So let me know if your testing pops up anything and I'll get it fixed.
Thanks,
S.

mojoaxel:
Thanks again! It's a real improvement!

PR #2062, opened 2016-09-02
Hide vertically hidden ranged items in groups that are not visible

yotamberk:


mojoaxel:
@yotamberk Thanks!

I would like to test this. Is there an issue this PR fixes? Which example can I use to see the changes?

lib/timeline/component/item/RangeItem.js
```diff
@@ -43,8 +43,14 @@ RangeItem.prototype.baseClassName = 'vis-item vis-range';
  */
 RangeItem.prototype.isVisible = function(range) {
   // determine visibility
-  return (this.data.start < range.end) && (this.data.end > range.start);
-};
+var isVisible =
```
mojoaxel: The code formatting is broken here. Please check the indent of this function.

lib/timeline/component/Group.js
```diff
@@ -475,6 +493,15 @@ Group.prototype.order = function() {
 Group.prototype._updateVisibleItems = function(orderedItems, oldVisibleItems, range) {
   var visibleItems = [];
   var visibleItemsLookup = {}; // we keep this to quickly look up if an item already exists in the list without using indexOf on visibleItems
+   
+  if (!this.isVisible) {
```
mojoaxel: This is the problem because examples with no groups are no longer working. Please rethink this lines again.

yotamberk:
Hey 😊
I'll fix the problem that was caused by my changes and fix what you commented on. 
I have an example but it is not simple just by looking at the timeline. You'll need to follow an item as it disappears from view of the timeline window. I'll add an example.

Have a good day!
Yotam

נשלח מה-iPhone שלי

‫ב-4 בספט׳ 2016, בשעה 11:34, ‏‏Alexander Wunschik ‏<notifications@github.com> כתב/ה:‬

lib/timeline/component/item/RangeItem.js
yotamberk: I've fixed it

lib/timeline/component/Group.js
yotamberk: I solved it by add equals in the 
`Group.prototype._isGroupVisible = function (range) {
   var isVisible = 
   (this.top <= range.body.domProps.centerContainer.height - range.body.domProps.scrollTop) 
   && (this.top + this.height >= - range.body.domProps.scrollTop);
   return isVisible;
 }`
So that when there are no groups, it will see it as one group with top 0 and we get 0==0 which leads to visible items.
This seems to solve the problem I caused. Examples now work well. Something doesn't sit well with me in this solution...

 think there should be a calculation of the initial group height and it should not be 0. It should calculate the height as if all the groups' items in the horizontal range are visible initially and then remove them if they aren't in the vertical range.
I'm working on a more elegant solution then the one supplied in this merge request.

examples/timeline/groups/verticalItemsHide.html
```diff
@@ -0,0 +1,164 @@
...
+  };
+
+
+  // create a Timeline
+  var container = document.getElementById('mytimeline');
+  timeline = new vis.Timeline(container, null, options);
+  timeline.setGroups(groups);
+  timeline.setItems(items);
+
+// $(document).ready(function(){
```
mojoaxel: What about all this comments. Can you remove them?

yotamberk: Sure, will do

נשלח מה-iPhone שלי

‫ב-5 בספט׳ 2016, בשעה 14:16, ‏‏Alexander Wunschik ‏<notifications@github.com> כתב/ה:‬

yotamberk:
Hey, I've fixed the code so all examples work again. 
This MR allows hiding vertically any range items in groups that are not in the visible window of the timeline.
Hiding of all the other types of items will be submitted in another MR.

mojoaxel:
Thanks. I'll check and merge it in the next days.

yotamberk:
Great! Thanks!

mojoaxel:
The files in the `dist` should not be changed. :-|
Can you please reset them:

```
git reset HEAD~1 dist
git commit --amend --no-edit
```

yotamberk:
Done. Sorry for that...

mojoaxel:
I tested a few examples. Looks good!

mojoaxel:
@yotamberk Thanks again also for your patience!

yotamberk:
Ofcourse. I'm working on some other features and bug fixes. Going through
TODOs and trying to understand what is plausible. Hope to have a scroll bar
option for timeline by next week

בתאריך 14 בספט׳ 2016 1:56 לפנה״צ,‏ "Alexander Wunschik" <
notifications@github.com> כתב:

mojoaxel:
>  I'm working on some other features and bug fixes.

@yotamberk :+1:

Issue #1993, opened 2016-07-27
[enchancement] more options for the edges' endings


kagestrife:
Other than just an arrow, could we get a circle, a rectangular, etc? This may be trivial, but it's a really important feature, due to the use of different symbols in various graphical standards (i.e. in biology).
I am using the visNetwork library (R Shiny) and the mods there directed me here, to the vis repository.

mojoaxel:
@kagestrife Good Idea! What kind of symbols would you like/need?

kagestrife:
@mojoaxel Hi and sorry for the late response, I was on vacation and could not comment ASAP.

About the edges' endings, what I would like to be added is an " O ", an " X ", a " | ", a rectangular and maybe a tail " < ".

Let me know if it is doable, and thank you for the help. If I come up with other ideas, I will let you know.

mojoaxel:
@kagestrife Do you maybe have any demo chart for this kind of biological data? We would need to create an example for this and "real" data would be nice.

This enhancement could be done in about 1-3 hours. If I see a nice use case for this maybe I'm willing to spend the time ;-)

kagestrife:
@mojoaxel It's a particular algorithm that needs this set of icons. An example of a graph like this is: https://encrypted-tbn3.gstatic.com/images?q=tbn:ANd9GcQRB1FsCKIoUPKIhSTGNiFMR2BrITt9bxhBZB7ul3dfzx3gSUtabw

Where instead of circles, X's or |'s might be needed. Also, let us know when the updated version is available. A simple short example on how to use it, would be more than helpful as well.

mojoaxel:
@kagestrife Is this an official diagram type? Has this type maybe even a name? Where is it used? I'm asking because it would be much more interesting to provide a often used diagram type instead of implementing something just for you :-)

kagestrife:
@mojoaxel Yes, its an official diagram type and its called "Partial Ancestral Graphs". 
https://en.wikipedia.org/wiki/Ancestral_graph
There are some algorithms (such as FCI on the package pcalg) that output such graphs, which may use circle endpoints to indicate ambiguous orientations.
But, beyond that, we (or another user) may need some different end points for the edges that are not currently under a known network, such as the ones we were referring before.

mojoaxel:
I still have problems finding a definition or something about these "Ancestral Graphs". There are no examples of the symbols you described. 

I could take the time to implement this new feature but I would need a specific definition what kind of symbols you want to use and what they should represent.

[...]

PR #2066, opened 2016-09-04
Network: edge-endpoint 'circle'

mojoaxel:
Make it possible to also define an `type` of edge-endpoint (`arrow` or `circle`):

image

implements #1993

### DIFF
--- a/lib/network/modules/components/edges/util/EdgeBase.js
+++ b/lib/network/modules/components/edges/util/EdgeBase.js
@@ -412,6 +412,7 @@ class EdgeBase {
     let node2;
     let guideOffset;
     let scaleFactor;
+    let type;
     let lineWidth = this.getLineWidth(selected, hover);
 
     if (position === 'from') {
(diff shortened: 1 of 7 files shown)


mojoaxel:
Can anybody please review this!

lib/network/shapes.js
```diff
@@ -206,9 +206,9 @@ if (typeof CanvasRenderingContext2D !== 'undefined') {
 
 
   /**
-   * Draw an arrow point (no line)
+   * Draw an arrow an the end of an line with the given angle.
```
Tooa: There are two typos: "**an** the end of **an**" -> "at the end of a"

lib/network/modules/components/edges/util/EdgeBase.js
```diff
@@ -491,8 +495,13 @@ class EdgeBase {
     ctx.fillStyle = ctx.strokeStyle;
     ctx.lineWidth = this.getLineWidth(selected, hover);
 
-    // draw arrow at the end of the line
-    ctx.arrow(arrowData.point.x, arrowData.point.y, arrowData.angle, arrowData.length);
+    if (arrowData.type && arrowData.type.toLowerCase() === 'circle') {
```
Tooa: Suggestion: Should we better have something like an enum type holding the different edge types (circle, arrow, ...)? Then the types would be stored in one place rather than distributed through the whole code as strings. However, for now this isn't a problem.

mojoaxel: Good Idea! Let'd do that if more edge types are implemented.

mojoaxel:
@yotamberk Could you please have a look at this!? I consider this a basis to implement further edge-types and not a completed new feature.

Issue #2023, opened 2016-08-17
Timeline: Chrome (Mac/Debian): Stuttering touch/drag of timeline produced by mousewheel usage in another viewsc


kuehn-sba:
When playing around in Chrome of my Mac with the embedded timeline within our web system, I've noticed that the timeline begins heavily to stutter when previously the mouse wheel was used in any scrollable content. 

The problem remains even exist when the URL is reloaded again. The only way to get a "smooth scrolling" of timeline back, is to close Chrome browser or to open a new tab.

You can reproduce it, when:

1. Open: http://visjs.org/examples/timeline/other/performance.html
2. draw 10000
2. Zoom out timeline to the max and **touch and drag** the timeline a lot to the left (it will scroll smooth all the time)
3. Now resize the browser window to a small size, so that a vertical scrollbar appears
4. Drag the scrollbar to the bottom without using the mouse wheel !!! (only by dragging the scrollbar)
5. Resize the browser window to big size and **touch and drag** timeline a lot to the left -> (new items will appear and the scrolling of timeline is smooth all the time - no problem at all)
6. Now Resize browser window to a very small window, so that a vertical scrollbar appears again
7. And now use only the mouse wheel to scroll the scrollbar to the bottom
8. Resize the browser window to a big size and **touch and drag** the timeline a lot to left (now the timeline will start heavily to stutter!!)
9. Reload URL -> zoom out to max-> scroll timeline to left -> it will start to stutter again after **touch and drag** the timeline

The problem was found when using Chrome on Unix (as ubuntu) and Mac.
That problem has nothing to do with resizing browser view, because stuttering appears in general, when using a mouse wheel for scrolling a scrollable element (e.g. combobox...) before.

I hope you can reproduce and identify the problem.

Maybe it has something to do with Hammer.JS, what you use internal for touching. 
I can see inside your source that you use Hammer.JS - v2.0.6 - 2015-12-23.
For testing purpose i had changed to hammer.js 2.0.8 code(replaced lines 11206-13776), but nothing helps (still stuttering).

kuehn-sba:
my tests so far:

line 16847 Range.prototype._onDrag emits rangechange event:

line 17819 
    this.on('rangechange', function () {
      if (this.initialDrawDone === true) {

// debug outputs are without stuttering

        this._redraw();
      }
    }.bind(this));

line 18348 Core.prototype._redraw = function () {

 // debug outputs are stuttering

}

kuehn-sba:
[...]

PR #2091, opened 2016-09-13
Use requestAnimationFrame to throttle redraws

gkubisa:
It allows the browser to determine the best rate at which redraws should happen and avoids choppy scrolling when Chrome starts throttling `setTimeout` (see https://docs.google.com/document/d/163ow-1wjd6L0rAN3V_U6t12eqVkq4mXDDjVaA4OuvCA/edit#)


mojoaxel:
@gkubisa Thanks very much! Thats a good idea! It may be even fix #2023.

Could you please remove the changes in the `dist` folder. We normally no not update this folder in the developing branch. This should work something like this:
```
git reset HEAD~1 dist
git commit --amend --no-edit
git push --force
```

gkubisa:
@mojoaxel, I've just removed those changes to the `dist` directory

yotamberk:
@mojoaxel This merge is causing a serious problem with the throttle redraw. I causes asynchronous behavior and causes  #2307. 
I don;t know how to solve this problem. I think we need to revert this merge.

gkubisa:
@yotamberk, reverting this PR would cause another regression (#2023). Would it be possible to just trigger synchronous redraw on initialization? If necessary, certain other actions could also trigger a synchronous redraw, eg updating the chart data programmatically.

yotamberk:
@gkubisa I understand. I've tried solving the bug it caused without much success... Can you take a look at this?

yotamberk:
Nvm, I've submitted a PR to solve the async problem (#2386)

PR #2093, opened 2016-09-14
Fix grammar/syntax in loading instructions

clottman:
Two small fixes to text of loading instructions in the timeline's docs.


mojoaxel:
@clottman Thanks!

Issue #2064, opened 2016-09-04
Timeline: Items jump to left:0 after selecting


mojoaxel:
In the [current develop branch](f1a403b6be570e81c86652e54aa61c726b7ff504) there there seems to be a bug with selected items:

1. Open the [basisUsage](/almende/vis/blob/master/examples/timeline/basicUsage.html) example.
2. Select the items by clicking on them.

The Items jump to left:0. 
This seems not to be the case for the items in the first row.

![vis_issue2062](https://cloud.githubusercontent.com/assets/600565/18229893/646d2d80-728a-11e6-8419-ad3a71ee1b9c.gif)

mojoaxel:
fixed with 6539ca74ed149d5d9695f64b61831d23f02908c9

PR #2106, opened 2016-09-19
Fixed position of box item (#2064)

Iskander508:


mojoaxel:
@Iskander508 Thanks you really much! Especially because it was my error that caused that problem in the first place!

Issue #2092, opened 2016-09-13
Gaps between timeline intervals


clottman:
On a data set where the items are sequential timeline ranges with no gaps in the data, there are gaps (white lines you can see around 13:20 and 13:55 on the example shown) between intervals at some zoom levels. Zooming in or sliding the timeline makes the gaps disappear, but often causes new ones to appear in different locations. 

I'm setting a custom background-color and color on .vis-item border: none as in the image shown. Otherwise, the default template is used. The problem persists even when default borders are used. Besides start & end, the only other option used is stack: false.

![image](https://cloud.githubusercontent.com/assets/4480480/18492397/ae771bbe-79d0-11e6-8af3-1ff87e7f88b7.png)

Same timeline, zoomed in. Gap has moved to ~14:55. 
![image](https://cloud.githubusercontent.com/assets/4480480/18492698/47cfbc0c-79d2-11e6-8c5c-4cfca2075652.png)

mojoaxel:
@clottman Thanks for reporting this. I would like to create an running jsbin demo to get to the button of this. Could you maybe provide code and styles to reproduce this issue!?

clottman:
Thanks for your help @mojoaxel. Here's a [JSBin](https://jsbin.com/huwatilose/edit?html,js,output), and the corresponding [Gist](https://gist.github.com/anonymous/568996767601b20e584f99b7794c36b9) reproducing the issue. You'll probably have to zoom in to see it; it can be pretty intermittent.

mojoaxel:
@clottman Thanks for reporting this! And thanks for the dummy-code, that helps A LOT!

It's a rounding issue! The left `position` and the `width` are calculated as float values. The browser then decides how to round them. In some circumstances a gap of 1px can occur. In the example below the width of elementA is rounded down while the left position of the next elementB is only just rounded up.

```
ElementA: left: 213.018px; width:14.4835px
ElementB: left: 227.501px;
```

We could try to improve the calculation of the left/width values!?

clottman:
It looks like the left & width values are calculated to match up unrounded, though I'm not familiar with the underlying code. Would it be possible to round the values of Element A explicitly before calculating the left value for Element B?

mojoaxel:
fixed with d9576bee502314ec4686a505b51ba24d0d742388
Thanks @Iskander508

clottman:
Thank you @lskander508 and @mojoaxel!

PR #2107, opened 2016-09-19
floating-point rounding compensation to range-item width (#2092)

Iskander508:


mojoaxel:
@Iskander508 Thanks very much! It looks like it works.
Just for my understanding: It looks to me like the box-width is now rounded up every time. Can this not result in an overlap of 1px sometimes?

Iskander508:
Yes, that's right, neighboring ranges can overlap by 1 pixel. It is not the cleanest solution for sure.
The overlapping could happen even in the old version in some cases.
e.g.:
range A: start 0.7, end 2.4
range B: start 2.4, end 3.4

A:
start: 0.7 => 1
width: 2.4 - 0.7 = 1.7 => 2
end: => 3 

B:
start: 2.4 => 2
width: 3.4 - 2.4 = 1.0 => 1
end: => 3

B is completely overlapped by A, no compensation used.

yotamberk:
@mojoaxel this merge causes #2318. I don;t understand this PR and what it solved. Can you explain? 
If this PR is important and we don't want to revert, we can solve #2318 by subtracting 0.5 a pixel in the start time.

mojoaxel:
@yotamberk This was a fix for #2092 . It basically numerically rounds up the item-width. It's very possible that this causes #2318 , but I see no good way to deal with that kind of numerical inaccuracy.

yotamberk:
@mojoaxel no problem. Now it's clear to me the importance of this PR. 
There are way to overcome the overlapping as mentioned in the issue #2318

PR #2109, opened 2016-09-20
timeline: add missing require for util module

drkenneth:
Added missing require for `util`, which is used in `Component.prototype.setOptions` but never declared.

Caught this while debugging some down-stream sanity checker failures.


mojoaxel:
@drkenneth Thanks!

PR #2124, opened 2016-09-28
Fix blur edge for dense networks

### LINKED ISSUE #1718: Network - blurEdge not always fired
Hi.

I noticed that the blurEdge event is not always fired. To reproduce, load the following example, open the console, and go nuts w/ the mouse. Eventually you will see that the blur and hover count diverge. It seems to get worse when 2 edges are very close to each other.

https://jsfiddle.net/njpazbuz/

Interestingly, the edge thickness is always correct.

### LINKED ISSUE #2122: Edge hover is not properly removed in dense network
Hi,

take a look at [this example](http://jsbin.com/nuhexutabo/1/edit?html,output) and dive with the cursor through the middle of the network. You will notice that the "hover" is not cleaned after the cursor leaves the edge. 

The issue is maybe related to [this pull-request](https://github.com/almende/vis/pull/1922). However, the problem also occurs with the latest develop branch! I think [this](https://github.com/almende/vis/blob/0666cb142c1fedaf63580a3cd94cb99f210d4325/lib/network/modules/SelectionHandler.js#L498) is the related code for hovering and [here](https://github.com/almende/vis/blob/0666cb142c1fedaf63580a3cd94cb99f210d4325/lib/network/modules/InteractionHandler.js#L572) is the hovering issued.

EDIT:

I think I got the problem. See my inline comments.

```
hoverObject(object) {
    [...]
    // removing all edge hover highlights
    for (let edgeId in this.hoverObj.edges) {
        [...]
        // If onMouseMove in the InteractionHandler passes not undefined to the hoverObject 
       // method, but rather another edge, the hover of the old edge is not removed. 
       // This is quite likely to happen for dense networks
        else if (object === undefined) {
          this.blurObject(this.hoverObj.edges[edgeId]);
          delete this.hoverObj.edges[edgeId];
          hoverChanged = true;
        }
      }
    } 
```

Best Uli

Tooa:
This pull request fixes #2122. It probably also fixes #1718.

Best Uli

### DIFF
--- a/lib/network/modules/SelectionHandler.js
+++ b/lib/network/modules/SelectionHandler.js
@@ -534,8 +534,9 @@ class SelectionHandler {
           this.hoverObj.edges[edgeId].hover = false;
           delete this.hoverObj.edges[edgeId];
         }
-        // if the blur remains the same and the object is undefined (mouse off), we blur the edge
-        else if (object === undefined) {
+        // if the blur remains the same and the object is undefined (mouse off) or another
+        // edge has been hovered, we blur the edge
+        else if (object === undefined || object instanceof Edge) {
           this.blurObject(this.hoverObj.edges[edgeId]);
           delete this.hoverObj.edges[edgeId];
           hoverChanged = true;


Tooa:
Can someone review this? You can test it with the network proposed in #2122. Thanks.

yotamberk:
Looks good to me

PR #2143, opened 2016-10-12
Fix Vertical visibility for all item types

yotamberk:
the calculation of every single item's visibility was fixed according to its type.

Vertical visibility is determined only by if the group itself is visible. Only groups that are visible in the timeline will expose there itemSets, then tested for horizontal visibility (for each of the item in these items is in range according to the item's `isVisible` function - differs by item type) and only then rendered to the DOM.

Every item has an `isVisible` function that checks if the item is in range. This calculation is different from item type to item type (ok... it is similar to ranged items and background items), and should not be defaulted in the `Item` class as mentioned in a `TODO` - hence I removed the comment.

The calculation for the `msPerPixel` is a simple calculation to insure the true width of box and point items is correct. Because these item types don't have an end time, I use the width of the item + it's start time to simulate it's end point with the `widthInMs` and subtract/add it according to `align` option.


mojoaxel:
@yotamberk Can you recommend an example to test this!?

yotamberk:
@mojoaxel It's a little bit hard to see because up until now the items would be visible horizontally even when not in range. I used debugging tools and the devtools to fix this. I can try to ad some kind of example. Give me a few hours =)

mojoaxel:
@yotamberk No I thinks that's ok. I ecpect you know what you are doing :-)

mojoaxel:
@yotamberk Thanks!

yotamberk:
Thanks again =)

PR #2158, opened 2016-10-17
Textual changes

wimrijnders:
Small text fixes to still some niggles:

- added line to `CONTRIBUTING.md` to prevent others making the same initial mistake I made.
- added `.directory` to `.gitignore` to prevent these files from showing up (This must bug others also, I can't be the only one using linux).

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -14,6 +14,7 @@ There are a few preferences regarding **code contributions**:
 - vis.js follows the node.js code style as described
   [here](http://nodeguide.com/style.html).
 - When implementing new features, please update the documentation accordingly.
+- Make changes in the `develop` branch, not the `master` branch.
 - Send pull requests to the `develop` branch, not the `master` branch.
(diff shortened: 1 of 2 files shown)


mojoaxel:
> added .directory to .gitignore to prevent these files from showing up (This must bug others also, I can't be the only one using linux).

I'm on linux, too. The ".directory" problem is only a Dolphin (KDE) thing.

CONTRIBUTING.md
```diff
@@ -14,6 +14,7 @@ There are a few preferences regarding **code contributions**:
 - vis.js follows the node.js code style as described
   [here](http://nodeguide.com/style.html).
 - When implementing new features, please update the documentation accordingly.
+- Make changes in the `develop` branch, not the `master` branch.
```
mojoaxel: ;-) the next line makes it clear, but why not!

wimrijnders:
Yes, but there must be other dolphin users!

....No, it's not clear enough. It's way too easy to pull straight from master. Which is what I did :-(.

PR #2160, opened 2016-10-17
All throw's use Error instances

wimrijnders:
Throw statements shouldn't be using bare strings; [Motivation](https://www.nczonline.net/blog/2009/03/10/the-art-of-throwing-javascript-errors-part-2/).

### DIFF
--- a/lib/graph3d/Filter.js
+++ b/lib/graph3d/Filter.js
@@ -111,7 +111,7 @@ Filter.prototype.getValues = function() {
  */
 Filter.prototype.getValue = function(index) {
   if (index >= this.values.length)
-    throw 'Error: index out of range';
+    throw new Error('Index out of range');
 
   return this.values[index];
 };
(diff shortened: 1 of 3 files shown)


wimrijnders:
I don't consider this a style change. This improves the way that the code interacts with the browser, which is essential for e.g. debugging. Please read the motivation link.

Also, I have no intention of changing anything outside of Graph3d. If you can agree with the motivation, perhaps an effort can be made for globally changing all throw's. But I'm not sure that I want to do it.

yotamberk:
I understand and agree with this change. I just want to make sure with @mojoaxel that we are on the same mindset. I'd like his opinion before we merge it.

wimrijnders:
Hokay. I'm preparing PR #5 in the meantime.

mojoaxel:
I also did not consider this a style change. This is good and useful.
The only problem I see is that it maight be considered a breaking change on some browsers, but I don't think that is _realy_ a problem. So let's go for it!

PR #2162, opened 2016-10-17
Translation of all data points to single method.

wimrijnders:
Consolidated code for performing translation on all data Points into a single method `_calcTranslations()`.


wimrijnders:
That's it for today. More tomorrow.

PR #2165, opened 2016-10-18
Moved recurring code for StepNumber to start() method

wimrijnders:
The following code fragment:

```js
step.start();
  if (step.getCurrent() < this.valueMin) {
    step.next();
}
```

Occurs every single time a `StepNumber` instance is used. This leads me to conclude that it should be part of the `start()` method. I've made its running optional, default is the original working.

### DIFF
--- a/lib/graph3d/StepNumber.js
+++ b/lib/graph3d/StepNumber.js
@@ -115,12 +115,28 @@ StepNumber.prototype.getStep = function () {
 };
 
 /**
- * Set the current value to the largest value smaller than start, which
- * is a multiple of the step size
+ * Set the current to its starting value.
+ *
+ * By default, this will be the largest value smaller than start, which
+ * is a multiple of the step size.
+ *
+ * Parameters checkFirst is optional, default false.
+ * If set to true, move the current value one step if smaller than start.
  */
-StepNumber.prototype.start = function() {
+StepNumber.prototype.start = function(checkFirst) {
+  if (checkFirst === undefined) {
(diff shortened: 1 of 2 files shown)


lib/graph3d/StepNumber.js
```diff
@@ -115,13 +115,29 @@ StepNumber.prototype.getStep = function () {
...
-StepNumber.prototype.start = function() {
+StepNumber.prototype.start = function(checkFirst) {
+  if (checkFirst === undefined) {
+    checkFirst = false;
+  }
+
   this._current = this._start - this._start % this._step;
+
+  if (checkFirst) {
+    if (this.getCurrent() < this._start) {
```
yotamberk: Does `this,_start' refer each time to the correct `this.<coordinate>Min`?

wimrijnders: Yes. If you follow the logic in the `StepNumber` c'tor:

- `this.setRange()` is called
- `in setRange()`:

```js
  this._start = start ? start : 0;
```
The only ways that the new value will not be set are:

- if it's zero, but that is the default anyway
- if it's `undefined` or `null`, but then the range is invalid
- if it's perhaps a boolean or another non-numeric type, in which case it had no business being passed.

wimrijnders: ...so unless the passed start value for some reason changes between the initialization of the `StepNumber` instance and the call to `start()` (which is not the case in the code), we are good.

PR #2176, opened 2016-10-18
Added structure for default values

wimrijnders:
This adds the initial defaults structure. It currently only does the simple fields; the rest require a bit more work.

Sorry, my mind is not clear enough to do the rest right now. Will add it bit by bit.

### DIFF
--- a/lib/graph3d/Graph3d.js
+++ b/lib/graph3d/Graph3d.js
@@ -64,7 +64,7 @@ var DEFAULTS = {
   height           : '400px',
   filterLabel      : 'time',
   legendLabel      : 'value',
-  xLabel           : 'poep',
+  xLabel           : 'x',
   yLabel           : 'y',
   zLabel           : 'z',
   xValueLabel      : function(v) { return v; },
--- a/lib/graph3d/Graph3d.js
+++ b/lib/graph3d/Graph3d.js
@@ -39,7 +39,9 @@ var OPTIONKEYS = [
   'animationPreload',
   'animationAutoStart',
   'axisColor',
-  'gridColor'
+  'gridColor',
+  'xCenter',
+  'yCenter'
 ];
 
 
@@ -78,7 +80,9 @@ var DEFAULTS = {
   animationInterval: 1000,      // milliseconds
   animationPreload : false,
   axisColor        : '#4D4D4D',
-  gridColor        : '#D3D3D3'
+  gridColor        : '#D3D3D3',
+  xCenter          : '55%',
+  yCenter          : '50%'
 
 
   // Following not in defaults (yet) but present in user settings
@@ -173,9 +177,6 @@ function Graph3d(container, data, options) {
   // The rest of the fields.
   // These require special attention in some way
   // TODO: handle these
-
-  this.defaultXCenter = '55%';
-  this.defaultYCenter = '50%';
 
   this.showLegend  = undefined; // auto by default (based on graph style)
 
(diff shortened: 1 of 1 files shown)


lib/graph3d/Graph3d.js
```diff
@@ -147,15 +210,16 @@ function Graph3d(container, data, options) {
   // TODO: customize axis range
 
   // colors
-  this.axisColor = '#4D4D4D';
-  this.gridColor = '#D3D3D3';
   this.dataColor = {
```
mojoaxel: Why not move `dataColor` also to the DEFAULTS object?

lib/graph3d/Graph3d.js
```diff
@@ -85,45 +152,41 @@ function Graph3d(container, data, options) {
...
-  var passValueFn = function(v) { return v; };
-  this.xValueLabel = passValueFn;
-  this.yValueLabel = passValueFn;
-  this.zValueLabel = passValueFn;
-  
-  this.filterLabel = 'time';
-  this.legendLabel = 'value';
   this.showLegend  = undefined; // auto by default (based on graph style)
 
   this.style = Graph3d.STYLE.DOT;
```
mojoaxel: What's with `showLegend` an `style`?

lib/graph3d/Graph3d.js
wimrijnders: If you wait a moment, I can get `xCenter` and `yCenter` into the defaults.

wimrijnders:
Here you go; these were easy enough to do.

wimrijnders:
Look, I see the 'Changes requested' message; I hope you understand that the rest is a bit harder to do than simply copying over. I stated in the initial message that this change does the 'simple' fields only, and that the rest requires 'a bit more work'.

Let me put it this way: do you agree until now? If so, there are two options:

1. Accept this PR and let me add the rest of the defaults incrementally in separate PR's.
2. Leave this PR open and wait till the rest has been done.

Both are fine with me. I have a personal preference for incremental PR's, but keeping this one open works for me as well.

lib/graph3d/Graph3d.js
wimrijnders: showLegend is non-trivial:

```js
  this.showLegend        =  (this.defaultShowLegend !== undefined) ? this.defaultShowLegend : isLegendGraphStyle;
```

wimrijnders:
I would appreciate a response to previous in due time.

Also, I have to say I am a bit miffed that you slapped a 'Changes requested' on this PR without actually requesting changes. I really don't want to have to guess your thoughts.

Could you please explicitly state what changes you request? Thanks.

wimrijnders:
So. You approved without adding a single word of response.

Alexander, I will state this explicitly because it is bothering me: I consider this non-communication of yours very, very bad form. Please change this.

mojoaxel:
> I consider this non-communication of yours very, very bad form. Please change this.

I'm very sorry! I'll try to communicate more.
Let me just say, it was late yesterday and actually I just agreed with you: *"Accept this PR and let me add the rest of the defaults incrementally in separate PR's"*. That is why I just approved it. I should have just added "comments" and not "request changes". My mistake.
No bad atmosphere intended! :peace_symbol:

wimrijnders:
Thank you very much! I honestly appreciate this response. Time for emoticons:

:+1: :sunny: :smile_cat: :musical_note: :beers:

mojoaxel:
By the way: You are welcomed in our new chat-room (that sounds sooo 90s): https://gitter.im/vis-js

wimrijnders:
Thank you, Unfortunately, the link is not working.

mojoaxel:
> Thank you, Unfortunately, the link is not working.

I fixed it

PR #2182, opened 2016-10-19
Graph3d: Added fields with special handling to default settings.

wimrijnders:
As long as this PR is not merged, I will continue to add other fields.
But it's no problem if you merge this right away; I'll just do a new PR for the other fields.

### DIFF
--- a/lib/graph3d/Graph3d.js
+++ b/lib/graph3d/Graph3d.js
@@ -82,13 +82,23 @@ var DEFAULTS = {
   axisColor        : '#4D4D4D',
   gridColor        : '#D3D3D3',
   xCenter          : '55%',
-  yCenter          : '50%'
+  yCenter          : '50%',
 
 
   // Following not in defaults (yet) but present in user settings
   // These will be initialized as 'undefined'
   //'showAnimationControls',
   //'animationAutoStart'
+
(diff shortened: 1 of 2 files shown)


wimrijnders:
Previous commit adds field `style` to the defaults.

However, I ran into some trouble during testing; I was testing with inconsistent data and graph3d went into infinite loops.

For this reason, extra checks for inconsistent data have been added: scan for `Bugfix` in code.
Also, StepNumber has been hardened a bit to avoid going into infinite loops again.

I realize that this extends the scope of this change; the bugfixes and hardening should have been a separate PR. However, I could not separate the scopes in separate branches. I hope you can find this OK.

wimrijnders:
This commit undoes an unnecessary edit I made. The STYLE enum does not need to change scope.

wimrijnders:
Thanks.

I have a feeling this PR is getting too big. Perhaps it's a better idea to merge and start a new one?

wimrijnders:
OK still open. Adding two more fields.

wimrijnders:
Whoops :blush:

wimrijnders:
Guys, this is as far as I can take it for now. You can merge if you're good with the changes.

I need to think a bit harder now how to deal with the `xMax` family of fields.
I have an idea, but it doesn't fit into this PR.

PR #2189, opened 2016-10-19
Updates DataSet and DataView remove event payload

ZacBrownBand:
The `remove` event that is triggered in DataView and DataSet only includes the ids of the items that were removed, and since the items are no longer available, it is hard to tell what items were removed.

This pr would add an `oldItems` property to the object that is sent with the `remove` event. This is supplementary to the `items` containing the ids to maintain reverse compatibility.

This effects
- Items getting removed from DataSets
- Items getting removed from DataViews
- Clearing a DataSet using `ds.clear()`
- Setting the data for a DataView using `dv.setData()` or when the DataView is created

Also updates the documentation for `dataset` Callback section.

### DIFF
--- a/lib/DataSet.js
+++ b/lib/DataSet.js
@@ -669,14 +669,14 @@ DataSet.prototype._sort = function (items, order) {
 DataSet.prototype.remove = function (id, senderId) {
   var removedIds = [],
       removedItems = [],
+      ids = [],
       i, len, itemId, item;
 
-  if (!Array.isArray(id)) { // force everything to be an array for simplicity
-    id = [id];
-  }
+  // force everything to be an array for simplicity
+  ids = Array.isArray(id) ? id : [id];
 
-  for (i = 0, len = id.length; i < len; i++) {
-    item = this._remove(id[i]);
+  for (i = 0, len = ids.length; i < len; i++) {
+    item = this._remove(ids[i]);
     if (item) {
       itemId = item[this._fieldId];
       if (itemId) {
@@ -685,7 +685,7 @@ DataSet.prototype.remove = function (id, senderId) {
       }
     }
   }
- 
+
   if (removedIds.length) {
     this._trigger('remove', {items: removedIds, oldData: removedItems}, senderId);
   }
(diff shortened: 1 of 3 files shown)


mojoaxel:
@ZacBrownBand Thanks! Looks complete. Could you maybe provide a simple example that demonstrates how to use this!? That would be great.

ZacBrownBand:
A simple example would be if you wanted to know how long something was in a dataset for. The item in the dataset could have a `created` property marking the time is was added to the dataset. When handling the `remove` event you would have access to that data, not just the id. While handling the remove, you could just do a simple time subtraction,

To accomplish this now, you would have to maintain a hashmap of the ids and times items were added, so when you get the remove event with the id you can look up the time.

This was inspired by an application I am working that uses vis for data management. We have an issue were we have internal and external ids since we manage data from multiple sources, this leads to use needing more than just the id in special cases. By having the full object that was removed it is much easier to do entity resolution and ensure other things are updated accordingly.

mojoaxel:
@ZacBrownBand I see. I guess you need a kind of special case to use this. But why not! Thanks!
Did you test your code? I don't want to write a test for that...

ZacBrownBand:
I wrote unit tests in my application, I'm happy to add one in here!

mojoaxel:
> I wrote unit tests in my application, I'm happy to add one in here!

We don't have any unit tests. But if you want to start feel free to provide a pull request ;-)

mojoaxel:
Sadly this introduced an error in DataView.refresh in [line 120](https://github.com/ZacBrownBand/vis/blob/d4ef91f3fe22d1e499d43f83aa79f24ba12131e5/lib/DataView.js#L120). I'll fix it.
@ZacBrownBand That happens! Next time please run `npm test` before committing.

ZacBrownBand:
@mojoaxel Sorry about that, thanks for catching it. I didn't realize there was `npm test`, I know for next time.

lib/DataSet.js
```diff
@@ -668,25 +668,26 @@ DataSet.prototype._sort = function (items, order) {
...
+      i, len, itemId, item;
+
+  // force everything to be an array for simplicity
+  ids = Array.isArray(id) ? id : [id];
+
+  for (i = 0, len = ids.length; i < len; i++) {
+    item = this._remove(ids[i]);
+    if (item) {
+      itemId = item[this._fieldId];
+      if (itemId) {
```
p1erce: This brings in some old bugs. Like I want to remove {id: 0, from: 4, to: 5} from DataSet edges.

PR #2195, opened 2016-10-20
Common handling for settings with prefix 'default'

wimrijnders:
This adds the final user settings to the defaults. Handling of the settings is now complete.

### DIFF
--- a/lib/graph3d/Graph3d.js
+++ b/lib/graph3d/Graph3d.js
@@ -146,21 +146,21 @@ var DEFAULTS = {
   },
 
   // Following stored internally with field prefix 'default'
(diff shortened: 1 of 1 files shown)


wimrijnders:
That's quick. Do you really check all the changes?

mojoaxel:
> That's quick. Do you really check all the changes?

No, I had a quick look and realized it's again just refactoring. I trust that you test your changes.

wimrijnders:
Testing: Always. For every PR and every change I make within the PR.

It's not completely refactoring; I added some extra functionality to make the handling of the last settings possible.
On the other hand, it's not a spectacular change so maybe you're right.

Issue #1060, opened 2015-07-03
Scroll bar in the timeline


almusawi:
Hi,
First I want to thanks you for this amazing work.

In my timeline, I made a fixed hieght so the timeline will not become big. but in this case some of the items will be up or down, hidden in the timeline window.
my question is, would it be possible to add scroll bar on the right side of the timeline to show that they can see the items by draging up and down?

Thanks in advance

AlexDM0:
Hi,

Glad you like it :).

At the moment it is not possible to display a scrollbar for vertical scrolling. There is a longstanding feature request for this (#273) but we haven't gotten around implementing a custom scrollbar yet. 

Regards,

Alex

almusawi:
Hi again,

Thanks for your answer :)

Best regards,
Mohammed

micmeemac:
Thank you, Alex, for considering this feature for future development. 

The drag gesture feels a bit tedious in the desktop context and a scroll would be comparatively heavenly. 

Do you have any pointers for if I want to implement this myself? Obviously I'd want to be able to configure the pan to happen on either mouse scroll or drag depending on context.

josdejong:
@micmeemac the vertical scroll position is handled via the `Core.props.scrollTop` property and some helper functions `Core._setScrollTop`, `Core._getScrollTop`, `Core._updateScrollTop`, at the bottom of Core.js: https://github.com/almende/vis/blob/master/lib/timeline/Core.js

micmeemac:
Cool :) Thank you

daniel-pfeiffer:
@micmeemac very glad you're picking this topic up!  When panning up and down I find it confusing/frustrating that groups scroll along.  Ideally I would see the scrolling within each group, while all groups are always visible.
This might not be ideal for everybody, so there might be a choice to scroll either the whole body or each group that doesn't fit.

daniel-pfeiffer:
Oops, outlined my solution in #273, which is still the one listed in Feature requests.

Issue #466, opened 2014-11-22
vertical scroll bar on timeline?


Blugh:
Hi guys

Is there any possibility of adding a vertical scroll bar to the timeline when there a lot of groups? The current solution of setting maxHeight is very user unfriendly as the user has to keep dragging through a long list and nor is it immediately obvious that there is hidden content on the timeline.

PR #2196, opened 2016-10-20
Add a vertical scroll option for timeline [solves #273, #1060, #466]

### LINKED ISSUE #273: Show vertical scrollbar when contents do not fit vertically
Hello,

I was wondering if it would be possible to add an option to show a horizonal scrollbar on the groups when the list gets long.  I know currently you allow horizontal drag, but a scrollbar option would be nice too.  It sometimes hard to tell that there are more groups otherwise.

Thanks!

yotamberk:
fixes #273, #1060, #466

### DIFF
--- a/examples/timeline/other/verticalScroll.html
+++ b/examples/timeline/other/verticalScroll.html
@@ -2,7 +2,6 @@
 <head>
   <title>Timeline | Vertical Scroll Option</title>
 
-  <script src="../../../docs/js/jquery.min.js"></script>
   <script src="../../../dist/vis.js"></script>
   <link href="../../../dist/vis.css" rel="stylesheet" type="text/css" />
 
--- a/examples/timeline/other/verticalScroll.html
+++ b/examples/timeline/other/verticalScroll.html
@@ -94,7 +94,7 @@
   // specify options
   var options = {
     stack: true,
-    verticalScroll: false,
+    verticalScroll: true,
     maxHeight: 400,
     start: new Date(),
     end: new Date(1000*60*60*24 + (new Date()).valueOf()),
--- a/examples/timeline/other/verticalScroll.html
+++ b/examples/timeline/other/verticalScroll.html
@@ -5,72 +5,36 @@
   <script src="../../../dist/vis.js"></script>
   <link href="../../../dist/vis.css" rel="stylesheet" type="text/css" />
 
-
-  <style type="text/css">
-    body {
-      color: #4D4D4D;
-      font: 10pt arial;
-    }
-  </style>
   <script src="../../googleAnalytics.js"></script>
 </head>
 
-<body onresize="/*timeline.checkResize();*/">
+<body>
 
 <h1>Timeline vertical scroll option</h1>
 
 <div id="mytimeline"></div>
-<br>
-
+ 
 <script>
-  function showVisibleItems() {
-    var a = timeline.getVisibleItems();
-    console.log(a);
-    document.getElementById("visibleItemsContainer").innerHTML = ""
-    document.getElementById("visibleItemsContainer").innerHTML += a;
-  };
-
-  // get selected item count from url parameter
-  var count = 1000;
 
   // create groups
-  var groups = new vis.DataSet([
-    {id: 1, content: 'Truck&nbsp;1'},
-    {id: 2, content: 'Truck&nbsp;2'},
-    {id: 3, content: 'Truck&nbsp;3'},
-    {id: 4, content: 'Truck&nbsp;4'},
-    {id: 5, content: 'Truck&nbsp;5'},
-    {id: 6, content: 'Truck&nbsp;6'},
-    {id: 7, content: 'Truck&nbsp;7'},
-    {id: 8, content: 'Truck&nbsp;8'},
-    {id: 9, content: 'Truck&nbsp;9'},
-    {id: 10, content: 'Truck&nbsp;10'},
-    {id: 11, content: 'Truck&nbsp;11'},
-    {id: 12, content: 'Truck&nbsp;12'},
-    {id: 13, content: 'Truck&nbsp;13'},
-    {id: 14, content: 'Truck&nbsp;14'},
-    {id: 15, content: 'Truck&nbsp;15'},
-    {id: 16, content: 'Truck&nbsp;16'},
-    {id: 17, content: 'Truck&nbsp;17'},
-    {id: 18, content: 'Truck&nbsp;18'},
-    {id: 19, content: 'Truck&nbsp;19'},
-    {id: 20, content: 'Truck&nbsp;20'},
-    {id: 21, content: 'Truck&nbsp;21'},
-    {id: 22, content: 'Truck&nbsp;22'},
-    {id: 23, content: 'Truck&nbsp;23'},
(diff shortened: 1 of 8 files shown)


yotamberk:
solves #273, #1060, #2196

examples/timeline/other/verticalScroll.html
```diff
@@ -0,0 +1,118 @@
...
+
+<body onresize="/*timeline.checkResize();*/">
+
+<h1>Timeline vertical scroll option</h1>
+
+<div id="mytimeline"></div>
+<br>
+
+<script>
+  function showVisibleItems() {
```
mojoaxel: looks like this function is never called. I think it should be removed.

examples/timeline/other/verticalScroll.html
```diff
@@ -0,0 +1,118 @@
...
+    var a = timeline.getVisibleItems();
+    console.log(a);
+    document.getElementById("visibleItemsContainer").innerHTML = ""
+    document.getElementById("visibleItemsContainer").innerHTML += a;
+  };
+
+  // get selected item count from url parameter
```
mojoaxel: As far as I can see it there are no url parameters.

examples/timeline/other/verticalScroll.html
```diff
@@ -0,0 +1,118 @@
...
+  var count = 1000;
+
+  // create groups
+  var groups = new vis.DataSet([
```
mojoaxel: This could be easily done in a loop, but ok - I don't mind.

examples/timeline/other/verticalScroll.html
```diff
@@ -0,0 +1,118 @@
...
+    {id: 23, content: 'Truck&nbsp;23'},
+    {id: 24, content: 'Truck&nbsp;24'},
+    {id: 25, content: 'Truck&nbsp;25'},
+
+  ]);
+
+  // create items
+  var items = new vis.DataSet();
+
+  var order = 1;
```
mojoaxel: Better use `order` and `truck` as the loop variables, than you don't need `j`and `i`.

examples/timeline/other/verticalScroll.html
```diff
@@ -0,0 +1,118 @@
...
+      item: 10, // minimal margin between items
+      axis: 5   // minimal margin between items and the axis
+    },
+    orientation: 'top'
+  };
+
+
+  // create a Timeline
+  var container = document.getElementById('mytimeline');
+  timeline = new vis.Timeline(container, null, options);
```
mojoaxel: ok, this one is mainly just taste but it would be shorter to write:
```js
timeline = new vis.Timeline(container, items, groups, options);
```
I expect that also to be faster, because every `setGroup` and `setItems` propably requires a rerendering.

lib/timeline/component/ItemSet.js
```diff
@@ -151,9 +151,9 @@ function ItemSet(body, options) {
   this.groupTouchParams = {};
   // create the HTML DOM
 
-  this._create();
-
   this.setOptions(options);
```
mojoaxel: Are you sure `setOptions` should be called AFTER `_create`? Looks dangerous to me.

lib/timeline/Core.js
```diff
@@ -705,9 +724,15 @@ Core.prototype._redraw = function() {
   // calculate the widths of the panels
   props.root.width = dom.root.offsetWidth;
   props.background.width = props.root.width - borderRootWidth;
-  props.left.width = dom.leftContainer.clientWidth   || -props.border.left;
+
+  var scrollbarWidth = 0;
+  if (this.options.verticalScroll) {
+    scrollbarWidth = 16;
```
mojoaxel: Sadly the width of the scrollbar can be different in each browser. You need a hack to get it. Here is a nice explanation: http://stackoverflow.com/a/13382873/722162

examples/timeline/other/verticalScroll.html
```diff
@@ -0,0 +1,118 @@
+<html>
+<head>
+  <title>Timeline | Vertical Scroll Option</title>
+
+  <script src="../../../dist/vis.js"></script>
+  <link href="../../../dist/vis.css" rel="stylesheet" type="text/css" />
+
+
+  <style type="text/css">
```
mojoaxel: I'm not sure haw other examples handle this, but I think we don't need this style here. The reason is to keep the examples as clean an copyable as possible.

examples/timeline/other/verticalScroll.html
```diff
@@ -0,0 +1,118 @@
...
+    body {
+      color: #4D4D4D;
+      font: 10pt arial;
+    }
+  </style>
+  <script src="../../googleAnalytics.js"></script>
+</head>
+
+<body onresize="/*timeline.checkResize();*/">
```
mojoaxel: You should remove the `onresize`.

examples/timeline/other/verticalScroll.html
```diff
@@ -0,0 +1,118 @@
...
+
+<h1>Timeline vertical scroll option</h1>
+
+<div id="mytimeline"></div>
+<br>
```
mojoaxel: The `<br>` is not needed ;-)

examples/timeline/other/verticalScroll.html
```diff
@@ -0,0 +1,118 @@
...
+<script>
+  function showVisibleItems() {
+    var a = timeline.getVisibleItems();
+    console.log(a);
+    document.getElementById("visibleItemsContainer").innerHTML = ""
+    document.getElementById("visibleItemsContainer").innerHTML += a;
+  };
+
+  // get selected item count from url parameter
+  var count = 1000;
```
mojoaxel: It looks like `count`is only used once. I think the variable is not needed.

mojoaxel:
@yotamberk I also have the strange behavior, that the scrollbar shrinks if i scroll down. 
I think it would be better to force a full rendering if the scrollbar is active. Now it feels strange!

yotamberk:
No problem! I'm fixing the other PR and then will work on these changes. No need to apologize! Reviews are necessary and I myself require them to improve! 
Thanks for the review =)

lib/timeline/component/ItemSet.js
yotamberk: Ummm I think you read it wrong. `_create` is after `_setOptions`. originally it was first create and then setOptions. I didn't understand why but I'm going to return it as it was before...

lib/timeline/Core.js
yotamberk: fixed =)

yotamberk:
@mojoaxel I don't see the shrinking problem. Can you elaborate?

mojoaxel:
> I don't see the shrinking problem. Can you elaborate?

Here you can see how the left scrollbar gets smaller and smaller as I scroll down:
visjs_2196

mojoaxel:
@yotamberk And also please resolve the new conflict. Thx!

mojoaxel:
@yotamberk That is really a big improvement! :tada: 
Thanks a lot!! :+1:

yotamberk:
My pleasure! Working on some other cool features! 
- more advanced React support, 
- updating range items (end time updating with current time)
- and multi timed ranged items
- bug fixes and improvments

PR #2197, opened 2016-10-20
vertical scroll when zoomKey isn't triggered

### LINKED ISSUE #2193: Horizontal Scroll on mouse wheel
Hi, 
Sorry for asking question here... 
Is there any way to do horizontal scroll on mouse wheel, 
I want to Zoom by ctrlKey and mouse wheel but if only mouse wheel i want to have a horizontal scroll. 

Can some one help on this? Its urgent...

yotamberk:


lib/timeline/Core.js
```diff
@@ -161,6 +161,35 @@ Core.prototype._create = function (container) {
   this.dom.root.addEventListener('mousewheel', onMouseWheel);
```
mojoaxel: I would suggest to use the chance an change the existing mousewheel handler to something more generic:
```js
if (this.dom.root.addEventListener) {
	// IE9, Chrome, Safari, Opera
	this.dom.root.addEventListener("mousewheel", MouseWheelHandler, false);
	// Firefox
	this.dom.root.addEventListener("DOMMouseScroll", MouseWheelHandler, false);
} else {
	// IE 6/7/8
	this.dom.root.attachEvent("onmousewheel", MouseWheelHandler);
}
```

lib/timeline/Core.js
```diff
@@ -161,6 +161,35 @@ Core.prototype._create = function (container) {
...
+      me._redraw();
+      me.emit('scroll', event);
+    }
+
+    // Prevent default actions caused by mouse wheel
+    // (else the page and timeline both scroll)
+    event.preventDefault();
+  }
+
+  this.dom.center.addEventListener('mousewheel', onMouseWheelCenter);
```
mojoaxel: There is already a mousewheel handler on the root element. For performance-reasons we should only attach it once and reuse the `onMouseWheel` function.

yotamberk:
Fixed comments! On to the next PR

PR #2200, opened 2016-10-20
Line graph draw segments by depth order

wimrijnders:
See title. This change is relevant for two reasons:

1. For my future goal of having multiple graphs in one view. When combining graphs, the line graphs must also respect the depth order. This is for properly overlapping the graph elements of various graph types when drawing.

2. This makes the code for the line graph more similar to the other graph types, which allows for DRY'ing the bar graph drawing code.

Wrt. 1 this is not a complete solution, but it's a first step.


wimrijnders:
> looks good. I like the use of the new function _line.

Yes. That's the joy of DRY code.

PR #2202, opened 2016-10-21
Graph3d: make setting dotSizeRatio public

wimrijnders:
Oversight. `dotSizeRatio` wasn't public in the code, but the documentation says it is.


wimrijnders:
While I'm at it, DRY'd the dotSize calculation.

PR #2205, opened 2016-10-22
Added generic graph drawing loop; isolated point drawing of graph style 'Bar'

wimrijnders:
All graph drawing routines follow the following structure:

```js
Graph3d.prototype._redrawGraph = function() {
  var ctx = this._getContext()

  if (this.dataPoints === undefined || this.dataPoints.length <= 0)
    return; // TODO: throw exception?

  this._calcTranslations(this.dataPoints);

  for (i = 0; i < this.dataPoints.length; i++) {
    point = this.dataPoints[i];

    // Draw data point here
  }
};

```

It is possible to consolidate this loop and move graph-style specific point drawing
to separate routines. This PR shows how that will look like; I've done only a single
graph style (`Bar`) to illustrate this.

Changes:

- made generic routine `_redrawDataGraph()`, usable for all graph styles.
- made specific routine `_redrawBarGraphPoint()`, which draws a single point for style `Bar`
- as an extra, isolated the bar drawing code in `_redrawBar()`, so it can be reused.
- in method `redraw()`, pass `_redrawBarGraphPoint` as a reference to `_redrawDataGraph()`

The idea is:

- to make the drawing loop generic for all graph styles
- isolate the drawing of points for specific graph style in separate methods.
- do this for *all* present graph styles.

# Notes

**1.** This is a **very important update** in regard to handling multiple graphs in a view.
The reason is that the elements of the multiple graphs will need to be drawn together at the same time, so that the overlap of the various elements are correct in the view.

**2.** I would much rather prefer to use *graph-style specific classes* here, instead of passing
a point-drawing method.
However, I fully appreciate that implementing that in one go will be a 'Big Bang' change
which is totally unoverseeable for reviewing. For this reason, I am offering the changes in smaller,
bite-size chunks. The classes will have to wait.


wimrijnders:
Added that pending deletion of `delete.png`

mojoaxel:
> Added that pending deletion of delete.png

Sorry to be "pedantic" again. but isn't the "delete.png" something from the timeline? Why are you deleting this here in a graph3d pull-request?

wimrijnders:
God, I'm making such a mess of this.

PR #2206, opened 2016-10-22
Fix React support for group and item templates and add example

### LINKED ISSUE #2207: Feature Request: Support React component for groupTemplate and template
I'm a React user and need to inject React components to the timeline templates. 
It would be nice if it was possible to support React components to the templates via `ReactDOM.render` in the timeline `options.template` and  `options.groupTemplate`.

yotamberk:
fixes  #2207


examples/timeline/other/usingReact.html
```diff
@@ -0,0 +1,131 @@
...
+    <!--
+    For ease of use, we are including the React, ReactDOM and Babel CDN
+    builds to make getting started as fast as possible.
+    
+    In production, you'll want to instead look at using something 
+    like Gulp, Grunt or WebPack (my personal recommendation) 
+    to compile JSX into JavaScript. Also, check out:
+    http://facebook.github.io/react/docs/tooling-integration.html
+    -->
+    <script src="https://fb.me/react-15.0.1.js"></script>
```
mojoaxel: I don't trust fb. I would prefer getting all fexternel libs (in all examples) from "cdnjs.cloudflare.com". But thats just me!

yotamberk: I tend to agree but because React belongs to fb, i thought it would be a nice gesture. I also took it from React website exampless so i dont think it's too unreliable to day the last. Do you prefer i chane it?

examples/timeline/other/usingReact.html
```diff
@@ -0,0 +1,131 @@
...
+              });
+            }
+        }
+
+        var GroupTemplate = React.createClass({
+          render: function() {
+            var { group } = this.props;
+            return <div>
+              <label>{group.content}</label>
+              <button>do stuff</button>
```
mojoaxel: The "do stuff" button realy confused me, because I expected it to do something. How about replacing it with some simple meaningless content?
How about : 
```jsx
<span>&#x1f69a;</span>
<label>{group.content}</label>
```

examples/timeline/other/usingReact.html
```diff
@@ -0,0 +1,131 @@
...
+            </div>
+          }
+        })
+
+        var ItemTemplate = React.createClass({
+          render: function() {
+            var { item } = this.props;
+            return <div>
+              <label>{item.content}</label>
+              <button>do stuff</button>
```
mojoaxel: The "do stuff" button realy confused me, because I expected it to do something. How about replacing it with some simple meaningless content?
How about : 
```jsx
<span>&#x1F355;</span>
<label>{item.content}</label>
```

examples/timeline/other/usingReact.html
```diff
@@ -0,0 +1,131 @@
...
+            </div>
+          }
+        })
+
+        // specify options
+        var options = {
+          rtl: true,
```
mojoaxel: I would recommend going with the default options as far it makes sense in the example. Changing to much in one example may confuse new users.

examples/timeline/other/usingReact.html
```diff
@@ -0,0 +1,131 @@
...
+            ReactDOM.unmountComponentAtNode(element);
+            return ReactDOM.render(<GroupTemplate group={group} />, element);
+          },
+          orientation: 'top'
+        }
+     
+
+        var VisTimeline = React.createClass({
+            componentDidMount: function() {
+                initTimeline()
```
mojoaxel: semicolon missing ;-)

examples/timeline/other/usingReact.html
```diff
@@ -0,0 +1,131 @@
...
+          },
+          template: function (item, element) {
+            ReactDOM.unmountComponentAtNode(element);
+            return ReactDOM.render(<ItemTemplate item={item} />, element);
+          },
+          groupTemplate: function (group, element) {
+            ReactDOM.unmountComponentAtNode(element);
+            return ReactDOM.render(<GroupTemplate group={group} />, element);
+          },
+          orientation: 'top'
```
mojoaxel: I needed a second to realize that the "orientation" option had nothing to do with the template function. I would put it on top (e.g. line 93) to separate it from the "groupTemplate" option.

mojoaxel:
@yotamberk Thats a nice start with react. Maybe we need a extra documentation page for that at some point!?

yotamberk:
@mojoaxel I think we should add documentation when more issues pop up about react. Currently there is not enough to add special documentation for React.

PR #2212, opened 2016-10-22
Add point draw methods for styles 'dot' and 'dot-line'

wimrijnders:
- Added point drawing routines for `dot` and `dot-line`
- Moved selection of point drawing method to separate method
- DRY'd color selection for graph styles `bar` and `dot`

Style `dot-line` was actually absurdly easy to implement; it's just `dot`
with an extra line.

### DIFF
--- a/lib/graph3d/Graph3d.js
+++ b/lib/graph3d/Graph3d.js
@@ -817,6 +817,35 @@ Graph3d.prototype.setOptions = function (options) {
   }
 };
 
+
+/**
+ * Determine which point drawing method to use
+ */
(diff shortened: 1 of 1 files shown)


wimrijnders:
I test every time I make a pull request. Usually, I try out all the options in the tooltips demo to see if they all still work. For the current work, I think this is proof enough that it works.

> As the features stack up, it's getting harder to test manually..

The thing to note is that with all these changes I'm making, the external working of `Graph3d` doesn't change at all! It's just one big refactoring :wink: .

PR #2216, opened 2016-10-24
Fix small timeline bug when defining timeline with no options and fix docs explanations

yotamberk:


yotamberk:
Oh, ummm I meant to wait with that... I found a solution for a small hack i did

PR #2228, opened 2016-10-24
Fix scroll bugs with vertical scrollbar

yotamberk:


lib/timeline/Core.js
```diff
@@ -202,29 +207,29 @@ Core.prototype._create = function (container) {
...
   } else {
     // IE 6/7/8
-    this.dom.root.attachEvent("onmousewheel", onMouseWheel.bind(this));
+    this.dom.center.attachEvent("onmousewheel", onMouseWheel.bind(this));
   }
 
   function onMouseScrollSide(event) {
-    var current = this.scrollTop;
-    var adjusted = -current;
+    if (!me.options.verticalScroll) return;
```
Tooa: You should add curly braces here in order to be consistent with the code style.

lib/timeline/Core.js
```diff
@@ -1070,6 +1072,7 @@ Core.prototype._onPinch = function (event) {
  * @private
  */
 Core.prototype._onDrag = function (event) {
+  if (!event) return
```
Tooa: Add curly braces and missing `;`. See my comment above.

lib/timeline/Timeline.js
```diff
@@ -109,6 +104,11 @@ function Timeline (container, items, groups, options) {
   // current time bar
   this.currentTime = new CurrentTime(this.body);
   this.components.push(this.currentTime);
+  
+  // apply options
+  if (options) {
```
Tooa: Here we have the braces. Is this some Javascript style convention to omit them if it's a return? I'm not sure.

lib/timeline/Range.js
```diff
@@ -390,6 +390,8 @@ Range.prototype._onDragStart = function(event) {
  * @private
  */
 Range.prototype._onDrag = function (event) {
+  if (!event) return
```
Tooa: See above.

lib/timeline/Core.js
mojoaxel: I thinks, you can leave it as it is.

yotamberk:
@mojoaxel I kind of prefer without the curly braces. But it is a question of style. I think you can accept it like this. If we decide later differently, i'll be glad to change it.

PR #2229, opened 2016-10-24
#2100 fix and cluster by connection fix

### LINKED ISSUE #2100: Network: Zoom button doesn't work like expected
Hi,
Zoom button doesn't work correctly, it zooms to the center of network, not to the center of screen.
It supposed to work like traditional (google) map zoom button.
Not sure if it is bug.

Thanks and best regards

CapitanMorgan:
Fixes #2100, and makes the zoom button on network behave in the more expected fashion. Did this by defining a viewCenter for the canvas, changed the Zoom In/Out buttons function to mirror that of the scroll zoom function. Also fixes an error found with clusterByConnection leaving phantomEdges on the screen. In addition updated a test that fails when you rebuild the library and rerun the tests.

A generic example with the current vis.js combined distributable files
https://jsfiddle.net/p72kewye/35/

A generic example with the proposed changes to the vis.js combined distributable files.
https://jsfiddle.net/xzvxsLjh/1/


mojoaxel:
@CapitanMorgan By the way: Thanks very much for this bugfix!!

CapitanMorgan:
Fixed spacing, and a few other style items.

PR #2230, opened 2016-10-25
Added class Range for Min/Max value pairs.

wimrijnders:
This adds a class for DRY'ing the handling of Min/Max value pairs. In particular, the combinations:

- xMin, xMax
- yMin, yMax
- zMin, zMax
- valueMin, valueMax

This is a preparatory step for implementing multiple graphs in a view. With multiple graphs, handling the Min/Max values is going to get much more complicated. This PR serves to reduce that particular pain.

### DIFF
--- a/lib/graph3d/Graph3d.js
+++ b/lib/graph3d/Graph3d.js
@@ -326,15 +326,23 @@ Graph3d.prototype.getDistinctValues = function(data, column) {
 }
 
 
+/**
+ * Get the min/max values for the passed data column.
+ *
+ * Result is returned as a Range instance with min/max members
+ * properly set.
+ */
 Graph3d.prototype.getColumnRange = function(data,column) {
-  var minMax  = new Range();
+  var range  = new Range();
 
+  // Adjust the range so that it covers all values in the
+  // passed data elements.
   for (var i = 0; i < data.length; i++) {
     var item = data[i][column];
-    minMax.add(item);
+    range.adjust(item);
   }
 
-  return minMax;
+  return range;
 };
 
 
@@ -370,6 +378,31 @@ Graph3d.prototype._checkValueField = function (data) {
       + ' for graph style \'' + this.style + '\''
     );
   }
+};
+
+
+/**
+ * Set default values for range
+ *
+ * The default values override the range values, if defined.
+ *
+ * Because it's possible that only defaultMin or defaultMax is set, it's better
+ * to pass in a range already set with the min/max set from the data. Otherwise,
+ * it's quite hard to process the min/max properly.
+ */
+Graph3d.prototype._setRangeDefaults = function (range, defaultMin, defaultMax) {
+  if (defaultMin !== undefined) {
+    range.min = defaultMin;
+  }
+
+  if (defaultMax !== undefined) {
+    range.max = defaultMax;
+  }
+
+  // This is the original way that the default min/max values were adjusted.
+  // TODO: Perhaps it's better if an error is thrown if the values do not agree.
+  //       But this will change the behaviour.
+  if (range.max <= range.min) range.max = range.min + 1;
 };
 
 
@@ -450,7 +483,7 @@ Graph3d.prototype._dataInitialize = function (rawData, style) {
   if (withBars) {
     xRange.expand(this.xBarWidth / 2);
   }
-  xRange.override(this.defaultXMin, this.defaultXMax);
+  this._setRangeDefaults(xRange, this.defaultXMin, this.defaultXMax);
   this.xRange = xRange;
   this.xStep = (this.defaultXStep !== undefined) ? this.defaultXStep : xRange.range()/NUMSTEPS;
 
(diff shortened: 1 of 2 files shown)


wimrijnders:
To note: this changes has been tested against all `Graph3d` demo's and works for all of them.

lib/graph3d/Graph3d.js
```diff
@@ -322,18 +327,13 @@ Graph3d.prototype.getDistinctValues = function(data, column) {
 
 
 Graph3d.prototype.getColumnRange = function(data,column) {
-  var minMax;
+  var minMax  = new Range();
```
yotamberk: I think `range` would be a better name for this variable.

lib/graph3d/Graph3d.js
```diff
@@ -322,18 +327,13 @@ Graph3d.prototype.getDistinctValues = function(data, column) {
...
   for (var i = 0; i < data.length; i++) {
     var item = data[i][column];
-
-    if (i === 0) {
-       minMax = { min: item, max: item};
-    } else {
-      if (minMax.min > item) { minMax.min = item; }
-      if (minMax.max < item) { minMax.max = item; }
-    }
+    minMax.add(item);
```
yotamberk: I think the `add` function name is not too clear. I see it sets the min and max when receiving an item, so maybe it shoulb be called `set`?

yotamberk: It's also a bit weird that a `range`/`minMax` gets an item and sets/adds with it. I would expect to get a range or a coordinate value. Am I missing something?

lib/graph3d/Range.js
```diff
@@ -0,0 +1,84 @@
...
+Range.prototype.override = function(min, max) {
+  if (min !== undefined) {
+    this.min = min;
+  }
+
+  if (max !== undefined) {
+    this.max = max;
+  }
+
+  if (this.max <= this.min) this.max = this.min + 1;
```
yotamberk: this overrides the `max` supplied so maybe this should throw an error at the beginning to inure that `min < max`

lib/graph3d/Range.js
```diff
@@ -0,0 +1,84 @@
...
+ */
+Range.prototype.expand = function(val) {
+  if (val === undefined) {
+    return;
+  }
+
+  this.min -= val;
+  this.max += val;
+
+  if (this.max <= this.min) this.max = this.min + 1;
```
yotamberk: same as above

wimrijnders: Very good review! I like it that you're so critical.

lib/graph3d/Graph3d.js
wimrijnders: I prefer the term `add()`; this is a question of good commenting/documentation to state the intent.

Wrt. setting/adding: this method `getColumnRange()` deals with single data values. It is not on the level of ranges or coordinates. It is actually the method which initializes a particular range.

Again, this could use some commenting to state the intent.

lib/graph3d/Range.js
wimrijnders: The reason of the existence of this line, in fact this whole method, is compatibility with previous working. I'm careful not to disrupt the previous implementation.

But to be honest, it's a bit kludgy and has nothing to do with actual min/max ranges. I'll consider doing this differently.

lib/graph3d/Graph3d.js
wimrijnders: Agreed.

lib/graph3d/Graph3d.js
yotamberk: So maybe call it `init`?

wimrijnders: The idea of the call is to adjust the range so that all values passed fit in that range. Note that it's in a loop over all data points.

`init()` implies it's a one-time thing; but it's actually an adjustment of the range. So, perhaps `adjust()`?

yotamberk: sure that works for me

wimrijnders:
OK processed your feedback. Also added more comments. I hope you agree this is better.

lib/graph3d/Graph3d.js
```diff
@@ -321,20 +326,23 @@ Graph3d.prototype.getDistinctValues = function(data, column) {
 }
 
 
+/**
+ * Get the min/max values for the passed data column.
+ *
+ * Result is returned as a Range instance with min/max members
+ * properly set.
+ */
```
mojoaxel: I just noticed, that the parameters and return value is missing in this jsdoc header. So if you want to to be perfect you can add them: http://usejsdoc.org/tags-param.html

lib/graph3d/Graph3d.js
```diff
@@ -321,20 +326,23 @@ Graph3d.prototype.getDistinctValues = function(data, column) {
...
 Graph3d.prototype.getColumnRange = function(data,column) {
-  var minMax;
+  var range  = new Range();
 
+  // Adjust the range so that it covers all values in the
+  // passed data elements.
```
mojoaxel: Why not one line?

lib/graph3d/Graph3d.js
```diff
@@ -455,39 +477,49 @@ Graph3d.prototype._dataInitialize = function (rawData, style) {
...
-    this.valueMin = (this.defaultValueMin !== undefined) ? this.defaultValueMin : valueRange.min;
-    this.valueMax = (this.defaultValueMax !== undefined) ? this.defaultValueMax : valueRange.max;
-    if (this.valueMax <= this.valueMin) this.valueMax = this.valueMin + 1;
+    this._setRangeDefaults(valueRange, this.defaultValueMin, this.defaultValueMax);
+    this.valueRange = valueRange;
+  }
+
+
+  // check if a filter column is provided
+	// Needs to be started after zRange is defined
```
mojoaxel: indent

lib/graph3d/Range.js
```diff
@@ -0,0 +1,85 @@
...
+// class Range
+//
+// Helper class to make working with related min and max values easier.
+//
+//
+// The range is inclusive; a given value is considered part of the range if:
+//
+//    this.min <= value <= this.max
+//
+////////////////////////////////////////////////////////////////////////////////
```
mojoaxel: You should use the official jsdoc header style. Here is an example: https://github.com/almende/vis/blob/develop/lib/graph3d/StepNumber.js#L1

lib/graph3d/Graph3d.js
wimrijnders: Because:

- Like this it fits within the opening brace
- I'm totally paranoid about the 80 column limit

In other words, this is purely subjective and personal. No other reason.

wimrijnders: Oh whatever. I'll adjust to one line.

wimrijnders:
OK, all feedback wrt. commening have been addressed. In addition:

For `Range.js`, `Graph3d.js`,

- Review current usage of `@param` and `@returns`, fixed if necessary
- Reformatted current commenting for better readability
- Reformatted commenting wrt. 80 column limit.

Scanning the jsdoc documentation, I see plenty more tags which are of use, eg `@private` and `@class`. I will leave these for now.

wimrijnders:
Thank you! Did my best with them.

But still plenty to do to get all jsdoc comments in order. Some other time.

PR #2234, opened 2016-10-26
Add german locale to timeline

Tooa:
This pull request adds german locale to the timeline component. I also added a missing export statement for the Italian locale. 

Best Uli


mojoaxel:
@Tooa Danke!

PR #2235, opened 2016-10-26
Support parent dir for RTL timeline and fix a small bug for editable point items

### LINKED ISSUE #2027: Auto-detect "Right To Left" support
Hi.

Currently, if the html doc is an RTL (`<html dir="rtl">`), the timeline isn't rendered 100% correctly - at first glance it looks OK, but as it's zoomed out, the dates are displayed on wrong place.

If it's possible, it'd be great to have the timeline rendered from right to left (to support Hebrew and Arabic) according to `dir` attribute - i.e., the earliest date on the right and later dates go to the left.

Thanks

yotamberk:
We can now define a rtl timeline buy reading the closest parent dir. closes #2027.

and fix a small bug I caused in a previous PR which made editable point items unuseable...


yotamberk:
@mojoaxel it already does it =)

lib/timeline/Timeline.js
```diff
@@ -57,12 +57,17 @@ function Timeline (container, items, groups, options) {
...
-  if (options) {
-    this.options.rtl = options.rtl
-  }
+
+
   // Create the DOM, props, and emitter
   this._create(container);
 
+  if (!options || (options && typeof options.rtl == "undefined")) {
+    this.options.rtl = window.getComputedStyle(this.dom.root, null).direction == "rtl";
```
mojoaxel: This line should be better something like this:
```js
var directionFromDom, domNode = this.dom.root;
while (!directionFromDom && domNode) {
    directionFromDom = window.getComputedStyle(domNode, null).direction;
    domNode = domNode.parentElement;
}
this.options.rtl = (directionFromDom && (directionFromDom.toLowerCase() == "rtl"));
```

yotamberk: I'll add this to another PR? Or do you want to revert?

mojoaxel: You could just change it and push the changes to your branch. Would be cool, if you test it before, I didn't ;-)

yotamberk: I tested it and submitted another PR =)

Issue #1881, opened 2016-06-02
Vis External Drag and Drop


faizanfarooq6:
Hello,

Is it possible to create an external item drag and drop on Vis timeline?

For example an external drag and drop on daypilot library: http://javascript.daypilot.org/demo/scheduler/external.html

PR #2238, opened 2016-10-28
Add drag and drop support 

### LINKED ISSUE #2180: Dragging external events
Hey,
First of all, great work! You have no idea how much this timeline has helped me in my projects.
Anyway, my question is, is there any way I could drag an external event(like "Eat Breakfast") and place it on the timeline (say 8:00am Oct14 2016)? I was going through the events(not the event I talked about in the previous line) of timeline and I couldn't find hovering or getting properties based on mouse's cursor on the timeline. for eg: we have the click event which returns the properties of the timeline (start, click location etc).
Is there a clickOut or hover or mouseX-Y event for the timeline so that when I am trying to drag an external event, I could place it inside the timeline base on my cursor's x and y coordinates inside the timeline. for eg: Dragging some timeEvent and then when I enter the timeline and drop it (say at 8:00am Oct14 2016), the event gets added to that particular time on the timeline. I know how to add external events and how drag and drop works, but I don't know how to get the properties when dropping the event. I'd appreciate any help.

Thanks!

yotamberk:
closes #2180, #1881

### DIFF
--- a/examples/timeline/other/drag&drop.html
+++ b/examples/timeline/other/drag&drop.html
@@ -21,7 +21,18 @@
     <li draggable="true" class="item">
       item 1.
     </li>
+    <li draggable="true" class="item">
+      item 2.
+    </li>
   </ul>
+
+<h1>other random elements</h1>
+<h1>other random elements</h1>
+<h1>other random elements</h1>
+<h1>other random elements</h1>
+<h1>other random elements</h1>
+<h1>other random elements</h1>
+<h1>other random elements</h1>
 </div>
 
 <script>
--- a/examples/timeline/other/drag&drop.html
+++ b/examples/timeline/other/drag&drop.html
@@ -5,7 +5,6 @@
 
   <script src="../../../dist/vis.js"></script>
   <link href="../../../dist/vis.css" rel="stylesheet" type="text/css" />
-  <link href="https://maxcdn.bootstrapcdn.com/font-awesome/4.7.0/css/font-awesome.min.css" rel="stylesheet" type="text/css" />
 
   <script src="../../googleAnalytics.js"></script>
 
@@ -20,8 +19,12 @@
       margin-bottom: 5px;
       padding: 5px 12px;
     }
-    i {
-      margin-right: 10px;
+    li.item:before {
+      content: "≣";
+      font-family: Arial, sans-serif;
+      display: inline-block;
+      font-size: inherit;
+      cursor: move;
     }
   </style>
 </head>
@@ -37,13 +40,13 @@
   <h3>Items:</h3>
   <ul class="items">
     <li draggable="true" class="item">
-      <i class="fa fa-bars"></i> item 1 - box
+      item 1 - box
     </li>
     <li draggable="true" class="item">
-       <i class="fa fa-bars"></i>item 2 - point
+      item 2 - point
     </li>
     <li draggable="true" class="item">
-       <i class="fa fa-bars"></i>item 3 - range
+      item 3 - range
     </li>
   </ul>
 </div>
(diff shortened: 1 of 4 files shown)


examples/timeline/other/drag&drop.html
```diff
@@ -0,0 +1,109 @@
...
+      start: new Date(),
+      end: new Date(1000*60*60*24 + (new Date()).valueOf()),
+    };
+
+    event.dataTransfer.setData("text/plain", JSON.stringify(item));
+  }
+
+  var items = document.querySelectorAll('.items .item');
+
+  items.forEach(function(item) {
```
mojoaxel: That doesn't work! `querySelectorAll`returns a `NodeList` and not an array. This could be fixed by:
```js
var items = Array.prototype.slice.call(document.querySelectorAll('.items .item'));
```
but that's not beautiful.
Better use an old style `for` loop instead the `forEach`!

examples/timeline/other/drag&drop.html
```diff
@@ -0,0 +1,109 @@
...
+
+<h1>Timeline Drag & Drop Example</h1>
+
+<p>For this to work, you will have to define your own <code>'dragstart'</code> eventListener on each item in your list of items (make sure that any new item added to the list is attached to this eventListener 'dragstart' handler). This 'dragstart' handler must set <code>dataTransfer</code> - notice you can set  the item's information as you want this way.</p>
+
+<div id="mytimeline" ></div>
+<div>
+  <h3>Items:</h3>
+  <ul class="items">
+    <li draggable="true" class="item">
```
mojoaxel: These ".items" should get a style with a max-width, a border, a drag-cursor and maybe a icon that shows that they can be dragged (something like "≣")

examples/timeline/other/drag&drop.html
```diff
@@ -0,0 +1,127 @@
...
+  <title>Timeline | Drag & Drop</title>
+
+  <script src="../../../dist/vis.js"></script>
+  <link href="../../../dist/vis.css" rel="stylesheet" type="text/css" />
+  <link href="https://maxcdn.bootstrapcdn.com/font-awesome/4.7.0/css/font-awesome.min.css" rel="stylesheet" type="text/css" />
+
+  <script src="../../googleAnalytics.js"></script>
+
+  <style type="text/css">
+    li {
```
mojoaxel: These generic styles can break somebodies application if copied without thinking. Better write: `li.item {`

examples/timeline/other/drag&drop.html
```diff
@@ -0,0 +1,127 @@
...
+      list-style: none;
+      width: 150px;
+      color: #1A1A1A;
+      background-color: #D5DDF6;
+      border: 1px solid #97B0F8;
+      border-radius: 2px;
+      margin-bottom: 5px;
+      padding: 5px 12px;
+    }
+    i {
```
mojoaxel: You could something like this, without needing FontAwesome at all :-)
```css
li.item:before {
    content: "≣";
    font-family: Arial, sans-serif;
    display: inline-block;
    font-size: inherit;
    cursor: move;
}
```

yotamberk: Agreed... Done!

yotamberk:
@mojoaxel There were some changes that had to be done there to support different kind of drop handlers (adding an item according to the item type in the drops' event data)

examples/timeline/other/drag&drop.html
```diff
@@ -0,0 +1,130 @@
+<!DOCTYPE HTML>
+<html>
+<head>
```
mojoaxel: Sorry I just noticed that the UTF-8 coding is missing: `<meta charset="utf-8"/>`
Without it the unicode-symbols are broken:
image

yotamberk: Done!

yotamberk:
> Another thing that I noticed: It would be nice if the newly added item would automatically be in the "edit" state (yellow), so you could move it just after dragging it onto the timeline. What do you think? Could that be achieved easily?

done!

mojoaxel:
@yotamberk Thank you so much! That's really a great new feature!

PR #2239, opened 2016-10-28
Add zoomIn and zoomOut to timeline core

### LINKED ISSUE #1955: Feature request: Add a "zoom" function
I discovered this library today and it is absolutely amazing. Really great job.  I do think that a zoom function is missing, it'd be great to programmatically be able to tell the timeline to zoom in or out. 

I did find an old issue https://github.com/almende/vis/issues/143 where there is some code showing how to implement that, but it would be great to have that built-in as a simple API we could call.

yotamberk:
closes #1955


docs/timeline/index.html
```diff
@@ -1322,9 +1322,20 @@ <h2 id="Methods">Methods</h2>
       </td>
     </tr>
 
-  </table>
-
+     <tr>
+      <td>zoomIn(percentage)</td>
+      <td>none</td>
+      <td>Zoom in the current visible window. The parameter <code>percentage</code> can be a <code>Number</code>. If the parameter value of <code>percentage</code> is null, the window will be left unchanged.
```
mojoaxel: We should make sure what a `percentage` value is. Something between [0..1] or [0..100]?

lib/timeline/Core.js
```diff
@@ -677,6 +677,46 @@ Core.prototype.getWindow = function() {
 };
 
 /**
+ * Zoom in the window such that given time is centered on screen.
+ * @param {String} time
```
mojoaxel: should be something like this:
```js
@param {Number} percentage Percentage of the min-max-timespan you want to zoom [0..1]
```
should be 
```js
@param {Number} percentage Percentage of the min-max-timespan you want to zoom [0..1]
```

lib/timeline/Core.js
```diff
@@ -677,6 +677,46 @@ Core.prototype.getWindow = function() {
...
+ */
+Core.prototype.zoomIn = function(percentage) {
+  if (!percentage) return
```
mojoaxel: We also should check if `percentage` is [0..1] otherwise throw an exception.

lib/timeline/Core.js
```diff
@@ -677,6 +677,46 @@ Core.prototype.getWindow = function() {
 };
 
 /**
+ * Zoom in the window such that given time is centered on screen.
+ * @param {Number} percentage
```
mojoaxel: maybe an additional explanation? That the percentage of the min-max-timespan. Percentage must be be [0..1]

lib/timeline/Core.js
```diff
@@ -677,6 +677,46 @@ Core.prototype.getWindow = function() {
...
+ */
+Core.prototype.zoomIn = function(percentage) {
+  if (!percentage) return
```
mojoaxel: we should also check if percentage is [0..1]. Otherwise throw exception.

docs/timeline/index.html
yotamberk: I've enforced the percentage to be between 0 and 1 now

lib/timeline/Core.js
```diff
@@ -677,6 +677,46 @@ Core.prototype.getWindow = function() {
 };
 
 /**
+ * Zoom in the window such that given time is centered on screen.
+ * @param {Number} percentage - must be between [0..1]
+ */
+Core.prototype.zoomIn = function(percentage) {
+  if (!percentage || percentage < 0 || percentage > 1) return
```
mojoaxel: I don't like the (non existing) error handling. But as far as I it is not better in all the other functions so that's ok for now.

PR #2246, opened 2016-10-29
generate details source maps of webpack modules

mojoaxel:
This sourcemaps really help during development. Here a screenshot how it looks in the chrome dev-tools:

image

But we should talk about if we want to remove the big map file from the distribution ZIP when building a release. What do you think?


yotamberk:
I think it's useful only for dev so leave it in `develop` and when we put out a release, remove it from the master

PR #2254, opened 2016-11-02
Timeline: options.editable to override item.editable

simo9000:
Removed redundant checks for item.editable. These checks cause the editable attribute of each item to override the timeline editable options. Applications that use timeline editable options to toggle between view and edit "modes" while maintaining partial edibility determined by item.editable need the converse to be true.


mojoaxel:
Thanks, but we only can accept pull-request against the `develop` branch. Read [this](https://github.com/almende/vis/blob/develop/misc/how_to_help.md#creating-pull-requests) for details.

simo9000:
@mojoaxel sorry about that. I didn't look hard enough for contribution guidelines. I will resubmit as directed.

PR #2256, opened 2016-11-03
Timeline: options.editable to override item.editable

simo9000:
Removed redundant checks for item.editable. These checks cause the editable attribute of each item to override the timeline editable options. Applications that use timeline editable options to toggle between view and edit "modes" while maintaining partial edibility determined by item.editable need the converse to be true. (Resubmit of #2254 to develop)

### DIFF
--- a/lib/timeline/component/ItemSet.js
+++ b/lib/timeline/component/ItemSet.js
@@ -1211,8 +1211,7 @@ ItemSet.prototype._onDragStart = function (event) {
   if (item && (item.selected || this.options.itemsAlwaysDraggable)) {
 
     if (!this.options.editable.updateTime &&
-        !this.options.editable.updateGroup &&
-        !item.editable) {
+        !this.options.editable.updateGroup) {
       return;
     }
 
--- a/lib/timeline/component/item/Item.js
+++ b/lib/timeline/component/item/Item.js
@@ -145,8 +145,7 @@ Item.prototype.repositionY = function() {
  * @protected
  */
 Item.prototype._repaintDeleteButton = function (anchor) {
-  var editable = (this.options.editable.remove || 
-                  this.data.editable === true) &&
+  var editable = this.options.editable.remove &&
                  this.data.editable !== false;
 
   if (this.selected && editable && !this.dom.deleteButton) {


brichbe:
I don't use the `item.editable` property, so I'm not directly affected by this, but looking at these diffs and the description provided by the author, it appears to me this alters the design in a way that will likely adversely affect anyone who does use these properties.

The behavior of these `editable` properties was purposefully designed so that the timeline's `editable` setting behaves as the default, but also so that default can by overridden on a per-item basis.  As an example, see ItemSet.js line 1218 - the comment and conditional clause below it confirms this is the intentional design.

So, this change proposes to modify that design so it behaves in a "converse" way, because doing so is useful to this author.  How is that accepted?  If someone using v4.16 has their code setup with timeline.editable = false, but some of their items are set as editable = true, they expect only those items to be editable -- so if they update to v4.17 all those items will be read-only because the timeline setting takes precedence?

Maybe I'm misunderstanding the intention and side affects of this merge request, but at a minimum, it appears to me some additional changes are needed to go along with this merge request, such as updating the timeline documentation to precisely explain how this new design behaves so users know which setting overrides the other setting, and under what circumstances that occurs.

yotamberk:
Hi @brichbe, reading this again I see your point. We might revert. The behavior is indeed weird now.

simo9000:
Thanks for the thoughts @brichbe. I recently updated an application from v3.12 which behaved with timeline options overriding individual item.editable values. I submitted the PR because the only alternative I can think of to accommodate v4.16 behavior and still be able to toggle editable on a subset of items in response to a user event is to iterate through all items and flip the item property* which seems inefficient. 

If the PR, as is, is unacceptable, then I would propose the addition of another timeline option to toggle between the two behaviors. This will make the predicates I modified in the current PR more cumbersome but at least both conventions will be possible. I would even set the default to be the v4.16 convention so it is a non-breaking change to v4.17.

Thoughts? I appreciate the discussion @brichbe.

*A closure might work as well but would make the code convoluted and prone to regressions.

mojoaxel:
Ok, lets resolve this!
After reading what @brichbe wrote it is clear to me that these changes would break the expected behavior. I see now were @simo9000 is going with this but I also don't think that this change is the right way, especially hours before the v4.17 minor release.
It was probably my and @yotamberk mistake not to test an fully comprehend the changes before merging, sorry for that. A special thanks to @brichbe for clearing things up. You can call this nice community team-work I guess...

@simo9000 We are still thankful for your contribution and hope you'll understand why we must revert this for now. I hope we can work together in the next weeks to find a better solution for you.

simo9000:
@mojoaxel, I fully agree with this decision. I am in no hurry to have this merged (I also was not aware of how close the minor release was, I did not mean to give anyone any concern with the timing of this PR). I have the aforementioned project running on a custom build which is sufficient for now. I would rather take more time with this to get it right and not affect the community of users. 

Going forward should we continue discussing the change here or open a new PR or issue?

mojoaxel:
> I would rather take more time with this to get it right and not affect the community of users.

:+1: Thanks for your understanding!

> Going forward should we continue discussing the change here or open a new PR or issue?

Feel free to open a new issue describing in detail what changes you propose. We can discuss the best way to handle the matter there. Just refer to this PR.

brichbe:
Sorry for the delay in responding but I'm glad to see there's agreement on this :+1: 

> then I would propose the addition of another timeline option to toggle between the two behaviors

@simo9000 yep, this was my initial thought too.  Understanding your concern (and the drawbacks) with needing to iterate all your items to flip the property, I do think the better way to add the flexibility you need would be adding a new sub-property (e.g. `overrideItems`) to the timeline's `editable` structure.  This of course would default to `false` for compatibility, but when set to `true` ensures these timeline settings override the item's settings.

simo9000:
> adding a new sub-property (e.g. `overrideItems`) to the timeline's editable structure. This of course would default to `false` for compatibility,

This is exactly what i was thinking as well. It would be backwards compatible and provide the simple handle to switch between behaviors.

@mojoaxel, @brichbe and I are in agreement in how to proceed with a better solution. Do you have any issue with me foregoing an new issue and instead starting a new PR?

mojoaxel:
> Do you have any issue with me foregoing an new issue and instead starting a new PR? I'll also update the documentation and add/update an example page.

Why not, but expect some discussion there :wink:

simo9000:
> expect some discussion there :wink:

I look forward to it.

PR #2262, opened 2016-11-05
Fix #2261

### LINKED ISSUE #2261: Issue with mousewheel zoom
I noticed problems with mouse wheel zoom in current develop branch.
Zoom is not possible at all (examples/timeline/interaction/limitMoveAndZoom.html) or seems to loose events (examples/timeline/other/localization.html).

This must be fixed before the v4.17 release because this was introduced since v4.16.1 :cry:

yotamberk:
closes #2261
(The mistake was adding the mousewheel event listener on the wrong container...)

### DIFF
--- a/lib/timeline/Core.js
+++ b/lib/timeline/Core.js
@@ -207,14 +207,14 @@ Core.prototype._create = function (container) {
     event.preventDefault();
   }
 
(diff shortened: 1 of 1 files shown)


mojoaxel:
@yotamberk Thank you so much!

PR #2275, opened 2016-11-07
Added option 'selectEvent' for Timeline.prototype.setSelection

itfourp:
When using multiple timelines with JQuery drag&drop in IE11 there occurs a problem that prevents the drag&drop events to be fired. To make JQuery drag&drop work, the 'pointerdown' event must be stopped for each item in the timeline. When calling 'event.stopPropagation()' JQuery drag&drop works, but the selection of items does not work any more. 
This PR adds an option to the setSelection-function of the Timeline-prototype that allows to call the select-event manually when the pointerdown-event-propagation was stopped.

The following code shows how to 'disable' the pointerdown-event and use the setSelection-function with the added option.

```javascript
for (var prop in myTimeline.itemSet.items) {
  if (myTimeline.itemSet.items[prop].dom &&
    myTimeline.itemSet.items[prop].dom.box) {
      var item = $(myTimeline.itemSet.items[prop].dom.box);
      var itemData = myTimeline.itemSet.items[prop].data;
      //  IE11-fix: prevent the timeline from canceling the event
      if (IsMicrosoftInternetExplorer) {
        item.on("pointerdown", function (event) {
          event.stopPropagation();
        });
        //  this re-enables selecting items
        item.on("click", function (event) {
          var clicked = $(event.currentTarget);
          var id = clicked.data("id");
          myTimeline.setSelection([id], { selectEvent: event });
        });
      }
      //  enable drag for items
      if (item.is("your selector here")) {
        item.draggable({
          axis: "y",
          revert: "invalid",
          revertDuration: 150,
          scroll: false,
          helper: "clone",
          appendTo: "body",
          drag: function (event, ui) {
            event.stopImmediatePropagation();
          },
          stop: function (event, ui) {
            event.stopImmediatePropagation();
          }
        });
      }
    }
}
//  create dropzone for another timeline
$(myTimeline2.body.dom.centerContainer).droppable({
  tolerance: "pointer",
  accept: "your selector here",
  hoverClass: "...",
  greedy: true,
  drop: function (event, ui) {
    //  your drop function
  }
});
```

Please note this is a very specific solution for a very specific problem. It does not solve the problem of dragging items of one timeline onto another but only fixes a side-effect that occurs when using JQuery drag&drop with the timeline.

### DIFF
--- a/docs/timeline/index.html
+++ b/docs/timeline/index.html
@@ -1322,7 +1322,7 @@ document.getElementById('myTimeline').onclick = function (event) {
           <li><code>focus: boolean</code><br>If true, focus will be set to the selected item(s)</li>
           <li><code>animation: boolean or {duration: number, easingFunction: string}</code><br>If true (default) or an Object, the range is animated smoothly to the new window. An object can be provided to specify duration and easing function. Default duration is 500 ms, and default easing function is <code>'easeInOutQuad'</code>. Only applicable when option focus is true. Available easing functions: <code>"linear"</code>, <code>"easeInQuad"</code>, <code>"easeOutQuad"</code>, <code>"easeInOutQuad"</code>, <code>"easeInCubic"</code>, <code>"easeOutCubic"</code>, <code>"easeInOutCubic"</code>, <code>"easeInQuart"</code>, <code>"easeOutQuart"</code>, <code>"easeInOutQuart"</code>, <code>"easeInQuint"</code>, <code>"easeOutQuint"</code>, <code>"easeInOutQuint"</code>.</li>
           <li><code>emitSelectEvent: Event</code><br>Call the select event manually to make sure the selected items will be highlighted in the timeline</li>
-		</ul>
+        </ul>
       </td>
     </tr>
 
--- a/docs/timeline/index.html
+++ b/docs/timeline/index.html
@@ -1321,7 +1321,8 @@ document.getElementById('myTimeline').onclick = function (event) {
         <ul>
           <li><code>focus: boolean</code><br>If true, focus will be set to the selected item(s)</li>
           <li><code>animation: boolean or {duration: number, easingFunction: string}</code><br>If true (default) or an Object, the range is animated smoothly to the new window. An object can be provided to specify duration and easing function. Default duration is 500 ms, and default easing function is <code>'easeInOutQuad'</code>. Only applicable when option focus is true. Available easing functions: <code>"linear"</code>, <code>"easeInQuad"</code>, <code>"easeOutQuad"</code>, <code>"easeInOutQuad"</code>, <code>"easeInCubic"</code>, <code>"easeOutCubic"</code>, <code>"easeInOutCubic"</code>, <code>"easeInQuart"</code>, <code>"easeOutQuart"</code>, <code>"easeInOutQuart"</code>, <code>"easeInQuint"</code>, <code>"easeOutQuint"</code>, <code>"easeInOutQuint"</code>.</li>
-        </ul>
+		  <li><code>emitSelectEvent: Event</code><br>Call the select event manually to make sure the selected items will be highlighted in the timeline</li>
+		</ul>
       </td>
     </tr>
 
--- a/docs/timeline/index.html
+++ b/docs/timeline/index.html
@@ -1321,7 +1321,7 @@ document.getElementById('myTimeline').onclick = function (event) {
         <ul>
           <li><code>focus: boolean</code><br>If true, focus will be set to the selected item(s)</li>
           <li><code>animation: boolean or {duration: number, easingFunction: string}</code><br>If true (default) or an Object, the range is animated smoothly to the new window. An object can be provided to specify duration and easing function. Default duration is 500 ms, and default easing function is <code>'easeInOutQuad'</code>. Only applicable when option focus is true. Available easing functions: <code>"linear"</code>, <code>"easeInQuad"</code>, <code>"easeOutQuad"</code>, <code>"easeInOutQuad"</code>, <code>"easeInCubic"</code>, <code>"easeOutCubic"</code>, <code>"easeInOutCubic"</code>, <code>"easeInQuart"</code>, <code>"easeOutQuart"</code>, <code>"easeInOutQuart"</code>, <code>"easeInQuint"</code>, <code>"easeOutQuint"</code>, <code>"easeInOutQuint"</code>.</li>
-		  <li><code>emitSelectEvent: Event</code><br>Call the select event manually to make sure the selected items will be highlighted in the timeline</li>
+          <li><code>emitSelectEvent: Event</code><br>Call the select event manually to make sure the selected items will be highlighted in the timeline</li>
 		</ul>
       </td>
     </tr>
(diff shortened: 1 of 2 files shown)


lib/timeline/Timeline.js
```diff
@@ -301,12 +301,27 @@ Timeline.prototype.setData = function (data) {
  *                                    Default duration is 500 ms, and default easing
  *                                    function is 'easeInOutQuad'.
  *                                    Only applicable when option focus is true.
+ *                                `selectEvent: Event`
+ *                                    Call the select event manually to make sure the selected items
+ *                                    will be highlighted in the timeline. Specific use for JQuery-
```
yotamberk: The comment is too specific for me. I see why we can use this addition, but the comment doesn't need the jQuery D&D mention.

yotamberk:
Thanks for this addition =)

lib/timeline/Timeline.js
```diff
@@ -301,12 +301,27 @@ Timeline.prototype.setData = function (data) {
...
+ *                                    drag&drop of timelime items when using IE11.
  */
 Timeline.prototype.setSelection = function(ids, options) {
-  this.itemSet && this.itemSet.setSelection(ids);
+  var oldSelection = [];
+  this.itemSet && (oldSelection = this.itemSet.getSelection()) && this.itemSet.setSelection(ids);
 
   if (options && options.focus) {
     this.focus(ids, options);
+  } else if (options && options.selectEvent) {
```
yotamberk: Missing documentation in `docs/timeline/index.html`. You should mention this option in the docs to allow people to know of it's existence. 

I also don't love the option name. I kind of like `options.emitSelectEvent` better

mojoaxel:
@itfourp Fist of all: Thanks for taking the time to create a pull request!
I really would feel better, if we had a simple example to reproduce and test this issue. Is it only in IE11? Introducing a new option looks like overkill to me, but I do not understand the problem fully yet, in the first place.

lib/timeline/Timeline.js
itfourp: Okay, I'll update the comment.

lib/timeline/Timeline.js
itfourp: Will be changed. I'll also add the option to the docs.

itfourp:
I'll try to set up a demo for the problem.

lib/timeline/Timeline.js
```diff
@@ -301,12 +301,26 @@ Timeline.prototype.setData = function (data) {
...
   if (options && options.focus) {
     this.focus(ids, options);
+  } else if (options && options.emitSelectEvent) {
+    var newSelection = this.itemSet.getSelection();
+	// either the new selection has items or the old one has
+    if (newSelection.length > 0 || oldSelection.length > 0) {
+      // emit select event
+      this.body.emitter.emit('select', {
+        items: newSelection,
+        event: emitSelectEvent
```
yotamberk: `emitSelectEvent` will be undefined. It should be `options.emitSelectEvent`

itfourp: Thank you, I took care of that.

yotamberk:
Ok, so i've checked your demo (You have a mistake calling the `.on('click')` on `myTimeline` instead of `draggableTimeline`). I'm having a problem with the Plunker (some weird uknown issue). Please fix this mistake. In any case, something in this proposal just feels a bit of a hack... Adding an option feels like a real overkill but if it's needed, I'll approve.

itfourp:
You're right, it's a little bit hackish. Adding an option for that very specific option was the only way for me to solve that problem. Maybe there is a better alternative?

Updated demo: http://jsbin.com/vapubaj/1/edit?html,output
Please note the vis-js version used in this demo does not yet have the emitSelectEvent-option - so selecting items in IE11/Edge does not work.

yotamberk:
Can you try emiting a `select` event on clicking an item? instead of being internal, you might be able to hack a way to do it externally. If not, we'll add this option.

yotamberk:
The example is not working for me on IE11

itfourp:
Thanks, I'll try out your approach later.
About the problem, could you please be more specific? Selecting items on the timeline cannot work yet in IE11 since the version of vis-js used in this demo does not yet have the patch.

yotamberk:
OHHHH I was wondering why it wasn't working... Duh... Let me try doing it without the hack. Give me a day and if I don;t manage, I'll approve and accept this

yotamberk:
Ok, so after trying, I've concluded there is something weird with this issue... I can't solve this externally. I approve this PR although not my favorite solution...

yotamberk:
@mojoaxel Please merge it if you agree

mojoaxel:
I have no access to an IE11 at the moment. Maybe somebody else can have a look!?

yotamberk:
I checked this out on IE11. solves the problem

docs/timeline/index.html
```diff
@@ -1321,7 +1321,8 @@ <h2 id="Methods">Methods</h2>
         <ul>
           <li><code>focus: boolean</code><br>If true, focus will be set to the selected item(s)</li>
           <li><code>animation: boolean or {duration: number, easingFunction: string}</code><br>If true (default) or an Object, the range is animated smoothly to the new window. An object can be provided to specify duration and easing function. Default duration is 500 ms, and default easing function is <code>'easeInOutQuad'</code>. Only applicable when option focus is true. Available easing functions: <code>"linear"</code>, <code>"easeInQuad"</code>, <code>"easeOutQuad"</code>, <code>"easeInOutQuad"</code>, <code>"easeInCubic"</code>, <code>"easeOutCubic"</code>, <code>"easeInOutCubic"</code>, <code>"easeInQuart"</code>, <code>"easeOutQuart"</code>, <code>"easeInOutQuart"</code>, <code>"easeInQuint"</code>, <code>"easeOutQuint"</code>, <code>"easeInOutQuint"</code>.</li>
-        </ul>
+		  <li><code>emitSelectEvent: Event</code><br>Call the select event manually to make sure the selected items will be highlighted in the timeline</li>
```
mojoaxel: Can you please fix the indent.

lib/timeline/Timeline.js
```diff
@@ -301,12 +301,26 @@ Timeline.prototype.setData = function (data) {
...
 Timeline.prototype.setSelection = function(ids, options) {
-  this.itemSet && this.itemSet.setSelection(ids);
+  var oldSelection = [];
+  this.itemSet && (oldSelection = this.itemSet.getSelection()) && this.itemSet.setSelection(ids);
 
   if (options && options.focus) {
     this.focus(ids, options);
+  } else if (options && options.emitSelectEvent) {
+    var newSelection = this.itemSet.getSelection();
+	// either the new selection has items or the old one has
```
mojoaxel: Can you please fix the indent here, too.

mojoaxel:
> I checked this out on IE11. solves the problem

I'm sorry that I have to say it over and over but I still don't understand the problem!
If this is only an issue in IE a new option is definitively not the right way to solve this problem! If it is a general problem maybe we always should emit the event? I don't want to introduce a new option for a very specific edge-case. As I said: I don't understand the problem to be honest.

mojoaxel:
> fix indent again

@itfourp Thx, that was fast! 
Maybe you can join [the gitter chanel](https://gitter.im/vis-js/Lobby) and we can talk about this issue!?

mojoaxel:
After talking to @itfourp I decided that a new option is not the right way to fix this issue. There seems to be problems with event propagation in MS browsers. @itfourp Is going to open a new issue and then we try to fix this problem in general without introducing a specific option.

@itfourp Thanks for bringing this up in the first place!

PR #2281, opened 2016-11-08
Graph2d performance enhancement

ludost:
Important performance enhancement supporting incremental changes to large graphs, these were too slow due to doing a full item conversion on each add().

Adding the id field to the DataSet output is a relative high impact change.

Note: the data handling side of Graph2d is currently somewhat over-engineered (=too complex) It can use some simplification....

### DIFF
--- a/lib/timeline/component/LineGraph.js
+++ b/lib/timeline/component/LineGraph.js
@@ -332,7 +332,7 @@ LineGraph.prototype.setGroups = function (groups) {
 };
 
 LineGraph.prototype._onUpdate = function (ids) {
-  this._updateAllGroupData();
+  this._updateAllGroupData(ids);
 };
 LineGraph.prototype._onAdd = function (ids) {
   this._onUpdate(ids);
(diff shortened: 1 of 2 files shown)


lib/timeline/component/LineGraph.js
```diff
@@ -341,7 +341,7 @@ LineGraph.prototype._onRemove = function (ids) {
   this._onUpdate(ids);
 };
 LineGraph.prototype._onUpdateGroups = function (groupIds) {
-  this._updateAllGroupData();
+  this._updateAllGroupData(null,groupIds);
```
mojoaxel: space missing ;-)

lib/timeline/component/LineGraph.js
```diff
@@ -427,10 +427,18 @@ LineGraph.prototype._updateGroup = function (group, groupId) {
  *
```
mojoaxel: jsDocs for param `ids`and param `groupIds` is missing.

lib/timeline/component/LineGraph.js
ludost: I was wondering if we should run a code style formatter over all files, as it's currently a bit of a mess.

mojoaxel: Yes, we are thinking about it. But a code formatter would destroy the "blame" History. It's maybe something for the next major release..

mojoaxel:
@ludost Thanks again!

PR #2286, opened 2016-11-09
Update index.html of Graph3D

Loreno10:
dataColor.fill and dataFill.stroke had wrong descriptions.


yotamberk:
You can call yourself a master developer now ;-)

PR #2290, opened 2016-11-09
Fix axis not displayed after daylight saving

### LINKED ISSUE #2251: Timeline axis labels disappearing
Hi, 
This might be a duplicate of #1393, but I can still reproduce the issue (v. 4.16.1). I've created a [bin here](http://jsbin.com/sezatos/edit?html,output).

You have to do a bit of fiddling to reproduce the issue:
1. Resize the width of the output window to make it as wide as possible. 
2. Click on "Run with JS"
3. Grab the far left of the timeline axis and drag it to the right.
At some point all the minor labels disappear.

This is also reproducible in the current examples page, e.g. on http://visjs.org/examples/timeline/other/localization.html. If you grab the far left of the axis and drag to the right as far as it will go, at some point the minor labels will disappear

Couple of screenshots:
Initial state:
image

Same screen, no minor labels:
image

Update: I was not able to reproduce the issue in my fork, but after fetching from upstream the issue has appeared. So, probably a regression

josdejong:
@mojoaxel can you have a look at this PR?

It fixes #2251. 

It's a bit of a workaround, it should be able to solve this by using moment.duration (see moment.js docs: http://momentjs.com/docs/#special-considerations-for-months-and-years), but I couldn't get that working so far.

With this fix, the axis will be displayed again after daylight saving switch:

image

What's not yet very nice is that the hour 2 o clock is displayed twice (empty space would be more neat I think):

image

### DIFF
--- a/lib/timeline/TimeStep.js
+++ b/lib/timeline/TimeStep.js
@@ -216,7 +216,17 @@ TimeStep.prototype.next = function() {
       case 'millisecond':  this.current.add(this.step, 'millisecond'); break;
       case 'second':       this.current.add(this.step, 'second'); break;
       case 'minute':       this.current.add(this.step, 'minute'); break;
-      case 'hour':         this.current.add(this.step, 'hour'); break;
+      case 'hour':
+        this.current.add(this.moment.duration(this.step, 'hour'));
+
+        // correct for daylight saving
+        // FIXME: use this.current.add(moment.duration(this.step, 'hour'))
+        // see http://momentjs.com/docs/#special-considerations-for-months-and-years
+        if (this.current.hours() % this.step !== 0) {
+          this.current.add(this.step - this.current.hours() % this.step, 'hour');
+        }
+
+        break;
       case 'weekday':      // intentional fall through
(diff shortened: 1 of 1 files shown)


itfourp:
Look like a good solution to me. I'd personally prefer to have 2 o clock twice over an empty space but that's not for me to decide.

lib/timeline/TimeStep.js
```diff
@@ -216,7 +216,17 @@ TimeStep.prototype.next = function() {
...
       case 'second':       this.current.add(this.step, 'second'); break;
       case 'minute':       this.current.add(this.step, 'minute'); break;
-      case 'hour':         this.current.add(this.step, 'hour'); break;
+      case 'hour':
+        this.current.add(this.moment.duration(this.step, 'hour'));
+
+        // correct for daylight saving
+        // FIXME: use this.current.add(moment.duration(this.step, 'hour'))
+        // see http://momentjs.com/docs/#special-considerations-for-months-and-years
+        if (this.current.hours() % this.step !== 0) {
```
mojoaxel: To be honest, I don't understand this part. Can you explain in a few words what you are doing here?

Issue #257, opened 2014-08-21
Toggle the visibility of groups and subgroups


josdejong:
It would be useful to implement a property `visible` for groups in the Timeline, similar to that of Graph2D, so you can toggle visibility of individual groups.

AlexDM0:
Hi,

Over the last year a lot of feature requests have been made. We have just introduced our new website which has a list of the requested features. We have placed this request on that list. 

The list can be found here:
http://visjs.org/featureRequests.html

An explaination of the new system can be found here:
http://visjs.org/blog.html#New\ website\ for\ vis\.js\!

I would like to stress that this does not mean we abandon this request. Discussion here will continue if needed on this feature but we will close it to keep our Github issue page more of a bug-todo list.

Future feature requests will still be made here and then added to the website by us.

Regards,

Alex

mojoaxel:
Reopening as Feature-Request issue (see #2114). 
Everybody: Please feel free to implement this!

simo9000:
@mojoaxel I just ran into the need for this feature. If you don't mind i'd like to take a shot at implementing this.

yotamberk:
closed with #2315

PR #2303, opened 2016-11-11
Toggle Group Visibility

simo9000:
This is an implementation of the feature requested in #257. This only toggles group visibility not subgroup visibility. This also includes and update to the documentation and an example page to highlight the new feature.

### DIFF
--- a/examples/timeline/groups/groupsEditable.html
+++ b/examples/timeline/groups/groupsEditable.html
@@ -28,7 +28,8 @@
 </head>
 <body>
 <p>
-  This example demonstrates editable groups (for now only reordering).
+  This example demonstrates editable groups (reordering and hiding).
+  <button onclick="showAllGroups()">Restore Hidden</button>
 </p>
 <div id="visualization"></div>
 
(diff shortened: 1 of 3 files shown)


examples/timeline/groups/groupsEditable.html
```diff
@@ -55,7 +56,30 @@
 	{"content": "WEC", "id": "WEC", "value": 18, className:'endurance'},
 	{"content": "GP2", "id": "GP2", "value": 19, className:'openwheel'}
   ]);
-  
+
+  // loop through groups and convert content strings into elements with hide buttons.
+  groups.forEach(function(group){
```
yotamberk: You're better off using the `options.groupTemplate` to achieve this. This seems kind of hacky and unclear...

simo9000:
I tried your idea with the dataset.filter and it works just as well as the PR. It is probably a better solution since it likely has less overhead for hidden groups. What i'll do then is break up the PR to remove the changes to the library and instead i'll modify the example to use the dataset.filter. Then #257 can reference the updated example and be closed.

yotamberk:
So I'll close this PR and when you're ready with a new one, please resubmit =)

PR #2304, opened 2016-11-11
Fix #2285

### LINKED ISSUE #2285: Timeline onUpdate no longer working in v4.17.0
I just upgraded to v4.17.0 and noticed `onUpdate` callback for the Timeline is not being triggered. Its not working in the example demo either: http://visjs.org/examples/timeline/editing/editingItemsCallbacks.html

yotamberk:
closes #2285

### DIFF
--- a/lib/timeline/component/item/Item.js
+++ b/lib/timeline/component/item/Item.js
@@ -141,6 +141,42 @@ Item.prototype.repositionY = function() {
 };
 
 /**
+ * Repaint a drag area on the center of the item when the item is selected
+ * @protected
+ */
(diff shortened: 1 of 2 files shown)


lib/timeline/component/item/Item.js
```diff
@@ -141,6 +141,42 @@ Item.prototype.repositionY = function() {
 };
 
 /**
+ * Repaint a drag area on the center of the item when the item is selected
```
mojoaxel: Why did you mode the function in the file? Its confusing... :-)

PR #2284, opened 2016-11-08
Timeline override items

simo9000:
This is a continuation of #2256. I have implemented a sub-option to `timeline.editable` called `overrrideItems` which provides a means to switch between two precedence conventions, 1. `item.editable` and 2. `timeline.editable`.

1. `item.editable`; (current behavior*) - this convention allows the `item.editable` option to override the values of `timeline.editable`. 

2. `timeline.editable` - this convention allows the timeline options to override the item options when 'overrideItems' is set to true. This is useful when switching an entire timeline between "editing" and "read-only" modes while still enforcing individual `item.editable` options during "editing" mode.

To preserve backward capability `timeline.editable.overrideItems` defaults to false which enables convention 1. This PR includes documentation updates and a new example page where `timeline.editable` options can be toggled to easily test different configurations. A [test matrix](https://docs.google.com/spreadsheets/d/1p3ldpknj067XwmCbv4sEPihogsXTx4qdGBpnTxyWunY/edit?usp=sharing) was created and used to confirm the behavior. Bugs fixed in [8ae5d2a](https://github.com/almende/vis/commit/8ae5d2acdb14de86660a21b8f190b7926dac8478) and [d2899c1](https://github.com/almende/vis/commit/d2899c1bed1eb581a8d157cf738e6712dafed502) were found as a result of this exhaustive testing.

*currently the setting `item.editable = true` overrides `timeline.editable.remove` and `timeline.editable.updateTime` but does not override `timeline.editable.updateGroup` which is inconsistent with the documentation. The bugs found in matrix testing were split into two commits to separate fixes that do not impact this behavior ([8ae5d2a](https://github.com/almende/vis/commit/8ae5d2acdb14de86660a21b8f190b7926dac8478)) and an update to make `item.editable = true` override all behaviors ([d2899c1](https://github.com/almende/vis/commit/d2899c1bed1eb581a8d157cf738e6712dafed502)). This makes it easy to revert [d2899c1](https://github.com/almende/vis/commit/d2899c1bed1eb581a8d157cf738e6712dafed502) if needed as this commit has the potential to impact backward compatibility.

PR #2305, opened 2016-11-12
Timeline: More editable control

simo9000:
I resisted the urge to do this but I didn't see a way around it. #2284 provided a clear means of decided when level of options was in control of the editable behavior of items. However it fell short in that when individual items are determined to be editable (and `overrideItems` is false), it enabled all three edit behaviors. This enables further control over individual item behavior but allowing `updateTime`, `updateGroup`, and `remove` to be toggled at the item level. I am glad I chose to pursue this as I found a bug which was introduced in #2284 where `timeline.editable == true` would also set `overrideItems` to true.

### DIFF
--- a/lib/timeline/component/ItemSet.js
+++ b/lib/timeline/component/ItemSet.js
@@ -361,7 +361,7 @@ ItemSet.prototype.setOptions = function(options) {
         this.options.editable.updateGroup   = options.editable;
         this.options.editable.add           = options.editable;
         this.options.editable.remove        = options.editable;
-        this.options.editable.overrideItems = options.editable;
+        this.options.editable.overrideItems = false;
       }
       else if (typeof options.editable === 'object') {
         util.selectiveExtend(['updateTime', 'updateGroup', 'add', 'remove', 'overrideItems'], this.options.editable, options.editable);
@@ -1225,8 +1225,8 @@ ItemSet.prototype._onDragStart = function (event) {
     }
 
     // override options.editable
-    if ((item.editable === false || !item.editable)
-    && !this.options.editable.overrideItems) {
+    if ((item.editable != null && !item.editable.updateTime && !item.editable.updateGroup)
+        && !this.options.editable.overrideItems) {
       return;
     }
 
@@ -1361,8 +1361,8 @@ ItemSet.prototype._onDrag = function (event) {
 
     //only calculate the new group for the item that's actually dragged
     var selectedItem = this.touchParams.selectedItem;
-    var updateGroupAllowed = me.options.editable.updateGroup ||
-                             (!this.options.editable.overrideItems && selectedItem.editable === true);
+    var updateGroupAllowed = ((this.options.editable.overrideItems || selectedItem.editable == null) && this.options.editable.updateGroup) ||
+                             (!this.options.editable.overrideItems && selectedItem.editable != null && selectedItem.editable.updateGroup);
     var newGroupBase = null;
     if (updateGroupAllowed && selectedItem) {
       if (selectedItem.data.group != undefined) {
@@ -1388,12 +1388,15 @@ ItemSet.prototype._onDrag = function (event) {
       }
 
       var itemData = this._cloneItemData(props.item.data); // clone the data
-      if (props.item.editable === false && !me.options.editable.overrideItems) {
+      if (props.item.editable != null
+        && !props.item.editable.updateTime
+        && !props.item.editable.updateGroup
+        && !me.options.editable.overrideItems) {
         return;
       }
 
-      var updateTimeAllowed = me.options.editable.updateTime || (
-          props.item.editable === true && !me.options.editable.overrideItems);
+      var updateTimeAllowed = ((this.options.editable.overrideItems || selectedItem.editable == null) && this.options.editable.updateTime) ||
+                               (!this.options.editable.overrideItems && selectedItem.editable != null && selectedItem.editable.updateTime);
       if (updateTimeAllowed) {
         if (props.dragLeft) {
           // drag left side of a range item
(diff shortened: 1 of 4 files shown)


lib/timeline/component/ItemSet.js
```diff
@@ -1388,12 +1388,14 @@ ItemSet.prototype._onDrag = function (event) {
...
-      if (props.item.editable === false && !me.options.editable.overrideItems) {
+      if (!props.item.editable.updateTime
+        && !props.item.editable.updateGroup
+        && !me.options.editable.overrideItems) {
         return;
       }
 
-      var updateTimeAllowed = me.options.editable.updateTime || (
-          props.item.editable === true && !me.options.editable.overrideItems);
+      var updateTimeAllowed = (me.options.editable.overrideItems && me.options.editable.updateTime) || (
```
yotamberk: This overrides the previous default... what happens when:
`me.options.editable.overrideItems==false` and `me.options.editable.updateTime==true`?
updating time in this case should be allowed even if `props.item.editable.updateTime` is not defined (or false, as set in default)

lib/timeline/component/ItemSet.js
```diff
@@ -1456,8 +1458,8 @@ ItemSet.prototype._onDrag = function (event) {
         }
       }
 
-      var updateGroupAllowed = me.options.editable.updateGroup ||
-          props.item.editable === true;
+      var updateGroupAllowed = (me.options.editable.overrideItems && me.options.editable.updateGroup) ||
+          props.item.editable.updateGroup;
```
yotamberk: What about: `!me.options.editable.overrideItems && me.options.editable.updateGroup`? You should be able to update groups but in this logic, you can't...

lib/timeline/component/item/Item.js
```diff
@@ -148,7 +164,7 @@ Item.prototype.repositionY = function() {
 Item.prototype._repaintDeleteButton = function (anchor) {
   var editable = (this.options.editable.remove &&
                   this.options.editable.overrideItems)
-                  || (this.data.editable === true &&
+                  || (this.editable.remove &&
                   !this.options.editable.overrideItems);
```
yotamberk: What about: `this.data.editable.remove && !this.options.editable.overrideItems`?

lib/timeline/component/ItemSet.js
```diff
@@ -1225,8 +1225,8 @@ ItemSet.prototype._onDragStart = function (event) {
     }
 
     // override options.editable
-    if ((item.editable === false || !item.editable)
-    && !this.options.editable.overrideItems) {
+    if (!item.editable.updateTime && !item.editable.updateGroup
```
simo9000: This predicate is in the `ItemSet._onDragStart`, my understanding is that  `remove` is only concerned with `Item._repaintDeleteButton`. Shouldn't this predicate not be affected by the value of `remove`?

yotamberk: oops you're right. I didn't notice the function it was in. Ignore this comment

simo9000:
@yotamberk Prior to this PR, there was no default for `item.editable`, the behavior was actually to enable editing if `item.editable` is not set as configured by the timeline. To keep this behavior, i have removed the defaults for `item.editable` to allow it to be null and updated the predicates. This oversight was actually made when i set up the [test matrix](https://docs.google.com/spreadsheets/d/1p3ldpknj067XwmCbv4sEPihogsXTx4qdGBpnTxyWunY/edit?usp=sharing) for #2284. I have changed the anticipated behavior of default for this commit. Please reconsider for approval.

Thanks for catching the logic flaws.

lib/timeline/component/ItemSet.js
```diff
@@ -1456,8 +1459,8 @@ ItemSet.prototype._onDrag = function (event) {
         }
       }
 
-      var updateGroupAllowed = me.options.editable.updateGroup ||
-          props.item.editable === true;
+      var updateGroupAllowed = ((me.options.editable.overrideItems || !props.item.editable ) && me.options.editable.updateGroup) || (
```
yotamberk: Here too:
`(me.options.editable.overrideItems && me.options.editable.updateGroup) || 
(!me.options.editable.overrideItems && (me.options.editable==true || me.options.editable.updateGroup)`

yotamberk:
I hope I'm not discouraging you. The logic is kind of messy and I think it needs to be cleaned up a bit before being approved

simo9000:
@yotamberk It's not discouraging at all. You are right that it was still messy. While it was "working" to pass all the tests I could come up with, the predicates were not standardized even though they were doing the same basic tests. I decided to formalize this logic:

`Ov = options.editable.overrideItems`
`<property> => updateGroup || updateTime || remove`
`Tp = options.editable.<property>`
`I = item.editable`
`Ip = item.editable.<property>`

The abstract predicate that applies in all cases is 

`((Ov || I == null) && Tp) || (!Ov && I != null && Ip)`

I have changed all three predicates you suggested changes on to use this form. The predicate at line 1460 of itemset.js is actually redundant as `updateGroupAllowed` is still in scope from the predicate at line 1364 so i have removed it.

Your code suggestions included `item.editable == true`, this will not work with the edits that i have made to Item.js. Item.editable is only assigned values of null or of object. As the abstract predicate indicates, i have removed the reliance on truthiness.

yotamberk:
MUCH BETTER! Good job! 
>Item.editable is only assigned values of null or of object.

I'm still wondering about that. seems a little weird that it's different than `options.editable`. But I'll approve this. Again, very nice work! Looking forward to more PRs =)

mojoaxel:
@simo9000 Thanks! Thats a nice improvement!

PR #2329, opened 2016-11-17
Edit edge without endpoint dragging, and pass label in data

### LINKED ISSUE #1756: editEdge is not called when I select an edge and click on the 'Edit Edge' button.
I am popping up a modal dialog when I am editing a node.  I want to do the same for editing an edge.

### LINKED ISSUE #2068: data in editEdge
Hello,
please, could you help me with a little thing? I have the following in the manipulation options:

	`editEdge: function (data, callback) {
		document.getElementById('operationEdge').innerHTML = "Edit Edge";
		document.getElementById('edge-id').value = data.id;
		document.getElementById('edge-label').value = data.label;
		document.getElementById('edge-title').value = data.title;
		document.getElementById('saveEdge').onclick = saveEdgeData.bind(this, data, callback);
		document.getElementById('cancelEdge').onclick = cancelEdgeEdit.bind(this, callback);
		document.getElementById('edgePopup').style.display = 'block';
	}`

My problem is that I want the user also be able to edit the edge label and title. Unfortunately, unlike at editNode, data only consists of id, from and to. No label, no title. Is there a way to get the full edges data at this place?

Thanks a lot in advance.

Best regards!

### LINKED ISSUE #2233: Get edge's label from editEdge event
We should be able to get the `label` information for edges in the `editEdge` event.

Currently, we can get `label` information for nodes from `editNode`, so it makes sense to be able to get the label from `editEdge` as well.

This will require adding another event - or adjusting the current events -  for editing edges, so that users aren't required to arbitrarily move an edge's link to trigger the `editEdge` event. Users should be able to edit an edge's label without any other adjustments.

For current workarounds: #2068

eymiha:
fixes #2233 
fixes #2068  
fixes #1756 

This adjusts the docs to the actual old mechanism and adds new code to skip the endpoint dragging and go straight to a user-defined edge editor if given.

Adding
```
{ manipulation: { editEdge: { editWithoutDrag: function } } }
```
calls the function directly, before any endpoint dragging takes place. If the old mechanism that activates endpoint dragging before the function call is desired, just supply the function directly as the value for `edgeEdit`.

In addition, the selected edge's label is added to the data passed to the edit function.

Also, when the user edit is cancelled, the edit bar was not being redisplayed. This is also fixed.


mojoaxel:
@eymiha Could you please add (or improve an existing) example to show this new feature. That way it would be much easier to test.

eymiha:
I added a quick and dirty example manipulation that edits edge labels via a popup, without dragging first. Edge creation still allows a drag, but the label assignment is included as well.

mojoaxel:
@eymiha Thanks very much for this nice improvement!

PR #2334, opened 2016-11-19
Fix group jittering 

### LINKED ISSUE #2287: Timeline jittering when updating group heights in 4.17.0
Hi,

Timeline examples in documentation all seem to allow it to grow in height as needed, with the timeline container set to, e.g.

`#visualization {
      box-sizing: border-box;
      width: 100%;
      height: 300px;
    }`

and no 'height' or 'maxHeight' set in the Timeline object options.

This allows it to grow or shrink in height nicely when zooming in/out, with the group heights being adjusted accordingly. However, this can lead the timeline to exceed the page height and make it more complicated to navigate around vertically. All fine for the examples on the site.

But if we want to limit it to a fixed height by setting the 'height' property in the Timeline object options to '100%' and navigate by dragging vertically, however, the growing/shrinking of group heights becomes jittery. Group lanes jump very quickly and momentarily to an intermediate position before landing in the final position, which makes it uncomfortable to the eye to navigate around the limits where items become more scarce and the timeline adjusts the height of groups more often.

Here's a jsbin with some data from the editable groups example in documentation:

https://jsbin.com/reqapuh/2/edit?html,output

I've set the 'height' option in the Timeline object to 100% and limited the container height to 600px to force the contents to not grow beyond the page height and this jittering to become visible when panning and zooming close to the ends of the data. If you drag slowly around Oct - Dec pay close attention to the lane labels jumping quickly and not smoothly as in the original example.

This doesn't happen in version 4.12.0, but it does in 4.17.0. I wanted to update to the latest version which adds a lot, but can't because of this very annoying issue that makes my timeline jitter so much at the slightest of mouse drags.

yotamberk:
fixes #2287

After much thought, I see that this is the best way for now to have vertically hidden items that are in range but not in view, without jittering of the heights. This doesn;t eliminate all jittering since items show up horizontally and will affect the height of the whole centerContainer (assuming `stack: true`) 

To compliment the smoothness, I suggest adding:
``` css
.vis-background .vis-group, .vis-foreground .vis-group, .vis-labelset .vis-label {
  -webkit-transition: height 1s; 
  -moz-transition: height 1s; 
  -ms-transition: height 1s; 
  -o-transition: height 1s; 
  transition: height 1s; 
}
```

### DIFF
--- a/examples/timeline/groups/verticalItemsHide.html
+++ b/examples/timeline/groups/verticalItemsHide.html
@@ -104,16 +104,8 @@
   var options = {
     stack: true,
     maxHeight: 400,
-    verticalScroll: true,
-    zoomKey: "ctrlKey",
     start: new Date(),
     end: new Date(1000*60*60*24 + (new Date()).valueOf()),
-    editable: true,
-    margin: {
-      item: 10, // minimal margin between items
-      axis: 5   // minimal margin between items and the axis
-    },
-    // orientation: 'top'
   };
 
 
--- a/examples/timeline/groups/verticalItemsHide.html
+++ b/examples/timeline/groups/verticalItemsHide.html
@@ -104,6 +104,8 @@
   var options = {
     stack: true,
     maxHeight: 400,
+    verticalScroll: true,
+    zoomKey: "ctrlKey",
     start: new Date(),
     end: new Date(1000*60*60*24 + (new Date()).valueOf()),
     editable: true,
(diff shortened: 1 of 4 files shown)


mojoaxel:
This does not really improve the jitter and therefore does not fix #2287

peek 2016-11-20 10-53

But this adds important changes to support the new group-visibility. Therefor I think we can merge it.

@yotamberk Can you please remove the "fixes #2287" from the description.
Also the css-transition work-around should be better postet in #2287 also.

Thanks!

PR #2339, opened 2016-11-21
Improvement of camera moving for 3d graph

linuxnotes:
Improvement of camera moving for 3d graph


linuxnotes:
not correct branch

PR #2343, opened 2016-11-21
label margins for nodes

### LINKED ISSUE #1653: Feature Request: Add 'Padding' Property to Network Graph Nodes
Currently, when drawing network graphs, there is not a lot of spacing between a node's internal label and its border - the node's **internal padding** is minimal. Increasing this internal padding, however, does not seem to be a trivial matter. As a newbie, I've read through the docs, trawled through StackOverflow and Google, but have not found a way to do it.

Ideally, this should just be another node property that the user could set easily, like this:

``` javascript
var nodes = new vis.DataSet([
    {id: 1, label: 'parent', padding: 18},
    {id: 2, label: 'child 1', padding: 9},
    {id: 3, label: 'child 2', padding: 9}
]);
```

If there is an easy workaround that I've missed, I'd be extremely happy to find out!

### LINKED ISSUE #2342: Specify label margins
To improve layout inside nodes, margin values should be able to be adjusted.

eymiha:
fixes #1653 
fixes #2342 

This change allows the margin values of node labels to be set as a group or independently, and allows the default margin to be set for a network. Initially the default is 5, matching the margin used in the prior implementation; existing networks will not change appearance.

On the network options -
```
    nodes: { margin: number }
```
to apply the same value on all of the label's sides, or
```
    nodes: { margin: { top: number, right: number, bottom: number, left: number} }
```
to specify individual values.

The margin for individual nodes is changed with the same margin specifications.

`examples/network/labels/labelMargins.html` demonstrates the feature.

### DIFF
--- a/docs/network/nodes.html
+++ b/docs/network/nodes.html
@@ -459,6 +459,36 @@ network.setOptions(options);
             <td><code>undefined</code></td>
             <td>When using the hierarchical layout, the level determines where the node is going to be positioned.</td>
         </tr>
+        <tr class='toggle collapsible' onclick="toggleTable('optionTable','margin', this);">
+            <td><span parent="margin" class="right-caret"></span> margin</td>
+            <td>Object or Number</td>
+            <td><code>5</code></td>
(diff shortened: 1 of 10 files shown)


mojoaxel:
@eymiha Nice Job! Thanks!

PR #2348, opened 2016-11-22
fixed label disappearing bug for large axis values in graph3d

gotoken:
Axis labels of graph3d is positioned on the middle point of between the origin and the center of each range. e.g.,

```javascript
xText   = xRange.center() / 2;
````

For example, for xRange = [100, 200], the center of range is (100+200)/2 = 150 and the middle point is (0+150)/2 = 75 then the middle point is positioned out of range. So axis labels disappear. This patch uses points divide each range in the ratio 1:3 as axis label positions 

I've setup preview https://fiddle.jshell.net/gotoken/5sk9ka72/3/show/ . Please try to reproduce by pasting this CSV into http://visjs.org/examples/graph3d/playground/index.html then click "Draw graph".

### DIFF
--- a/lib/graph3d/Graph3d.js
+++ b/lib/graph3d/Graph3d.js
@@ -1394,7 +1394,7 @@ Graph3d.prototype._redrawAxis = function() {
   var xLabel = this.xLabel;
   if (xLabel.length > 0) {
     yOffset = 0.1 / this.scale.y;
-    xText   = xRange.min + (xRange.max - xRange.min)/2;
+    xText   = (xRange.max + 2*xRange.min)/3;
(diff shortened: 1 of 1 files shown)


mojoaxel:
@wimrijnders Can you please have a look at this.

gotoken:
@mojoaxel Thanks you noticed him. I changed my patch to preserve original positioning way, that is, an axis label divides its range in the ratio 1:3. It may well be acceptable. Apologies for the changes after PR. 

01d9ad9

wimrijnders:
@mojoaxel This fix is fine by me,

PR #2352, opened 2016-11-22
This was a necessity I had when I started using dot-color of graph3d module, if you think it will be useful for someone else, you could accept this change.

geovane-gomes:
In graph3d, 'dot-color' graph style:

- Created the 'customPointColor' (default = false) option to allow the developer to define point color (fill and border) in 'dot-color' graphics;

- To use, you must change the 'customPointColor' option to true,
  e.g. "...keepAspectRatio: false,
           verticalRatio: 0.5,
           customPointColor : true
           };";
  and pass 'fill_color' and 'border_color' as 'data' object fields in 'data.add' function call,
  e.g. "var z = 0;
        var x = 0;
        var y = 4;
        var value = 20;
        var fillColor = 'red';
        var borderColor = 'rgb(255,255,255)';
        data.add({ z: z, x: x, y: y, style: value, extra: extra_content, fill_color: fillColor, border_color: borderColor });";

- When changing the 'customPointColor' option to true, it is good to change the 'showLegend' option to false so as not to make it strange in the graph view;

- I hope I can make an improvement in the legend soon;

### DIFF
--- a/dist/vis.js
+++ b/dist/vis.js
@@ -7913,7 +7913,8 @@ return /******/ (function(modules) { // webpackBootstrap
     yStep: autoByDefault,
     zMin: autoByDefault,
     zMax: autoByDefault,
-    zStep: autoByDefault
+    zStep: autoByDefault,
+    customPointColor: false
   };
 
   // -----------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


mojoaxel:
@geovane-gomes Sorry, but we only can accept pull-requests to the `develop` branch.

PR #2385, opened 2016-11-30
Multiple fonts in lables, and width and height constraints

eymiha:
(This replaces PR #2362, adding functionality for width and height constraints)

fixes #1603
fixes #1628
fixes #1936
fixes #2298 
fixes #2384

Adds multifont rendering capability to network node and edge labels. Existing networks are not altered.

font.multi designates html or markdown. Bold, italic and code (monospaced) markup is supported.

font.bold designates the bold font properties, font.italic the italic font properties, font.bolditalic the bold-italic font properties, and font.mono the monospace font properties.

A vadjust property has also been added to the fonts that may be used to adjust any baseline mismatching between fonts.

The network/labels/labelMultifont.html example provides illustration and some finer details.

------

Adds minimum and maximum widths for nodes, maximum widths for edges, and minimum height and vertical label alignment for nodes. Existing networks are not altered.

Maximum width settings will break label lines on spaces to fit. Minimum width (and height) settings will expand the label horizontally (or vertically) to reach the minimum.

If the node is expanded vertically, the vertical alignment option will push it to the top, middle or bottom of the unfilled space.

The network/nodeStyles/widthHeight.html example provides illustration and some finer details.

------

Properties may be assigned in the network node and edge options, or on individual nodes and edges.

The documents are also updated.

PR #2394, opened 2016-12-02
add edges.font.highlight for selected edge

ZhengxianLan:
Hello,I add a config to set selected edge's color.Please review and merge it, thank you .

### DIFF
--- a/examples/network/edgeStyles/arrows.html
+++ b/examples/network/edgeStyles/arrows.html
@@ -52,7 +52,18 @@
     nodes: nodes,
     edges: edges
   };
-  var options = {};
+var options = {
+  edges: {
+    label:'label hi',
+    font: {
+      color: '#A3A9B0',
+      highlight: '#475461',
+      strokeWidth: 0,
+      align: 'top',
+      size:10
+    }
+  }
+};
   var network = new vis.Network(container, data, options);
 </script>
 
--- a/lib/network/modules/EdgesHandler.js
+++ b/lib/network/modules/EdgesHandler.js
@@ -38,6 +38,7 @@ class EdgesHandler {
       dashes: false,
       font: {
         color: '#343434',
+        highlight: '#ff0000',
         size: 14, // px
         face: 'arial',
         background: 'none',
--- a/lib/network/options.js
+++ b/lib/network/options.js
@@ -41,6 +41,7 @@ let allOptions = {
     dashes: { boolean, array },
     font: {
       color: { string },
+      highlight: { string: string },
       size: { number }, // px
       face: { string },
       background: { string },
@@ -396,6 +397,7 @@ let configureOptions = {
     dashes: false,
     font: {
       color: ['color', '#343434'],
+      highlight: ['color', '#475461'],
       size: [14, 0, 100, 1], // px
       face: ['arial', 'verdana', 'tahoma'],
       background: ['color', 'none'],
--- a/lib/network/modules/components/shared/Label.js
+++ b/lib/network/modules/components/shared/Label.js
@@ -131,7 +131,7 @@ class Label {
 
     // configure context for drawing the text
     ctx.font = (selected && this.nodeOptions.labelHighlightBold ? 'bold ' : '') + fontSize + "px " + this.fontOptions.face;
-    ctx.fillStyle = fontColor;
+    ctx.fillStyle = (selected && this.nodeOptions.font.highlight) ? this.nodeOptions.font.highlight : fontColor;
     // When the textAlign property is 'left', make label left-justified
     if ((!this.isEdgeLabel) && this.fontOptions.align === 'left') {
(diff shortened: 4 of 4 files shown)


yotamberk:
Hi @ZhengxianLan, Please pull from `develop` branch to resolve the conflicts =)

ZhengxianLan:
OK～

2016-12-02 19:16 GMT+08:00 yotamberk <notifications@github.com>:

lib/network/modules/EdgesHandler.js
```diff
@@ -38,6 +38,7 @@ class EdgesHandler {
       dashes: false,
       font: {
         color: '#343434',
+        highlight: '#ff0000',
```
mojoaxel: I would consider this a breaking change because edge labels get now colored in red if highlighted. I think we could live with something like `#000000`but not red at this point.

lib/network/options.js
```diff
@@ -478,6 +479,7 @@ let configureOptions = {
     dashes: false,
     font: {
       color: ['color', '#343434'],
+      highlight: ['color', '#475461'],
```
mojoaxel: see above. This should also be `#000000`.

examples/network/edgeStyles/arrows.html
```diff
@@ -52,7 +52,18 @@
     nodes: nodes,
     edges: edges
   };
-  var options = {};
+var options = {
+  edges: {
+    label:'label hi',
+    font: {
+      color: '#A3A9B0',
+      highlight: '#475461',
```
mojoaxel: I think the arrows example is not the best place to show the new highlighted label feature. Can you please restore `edgeStyles/arrows.html` and make your changes e.g. in `labels/labelColorAndSize.html`

examples/network/edgeStyles/arrows.html
```diff
@@ -52,7 +52,18 @@
     nodes: nodes,
     edges: edges
   };
-  var options = {};
+var options = {
+  edges: {
+    label:'label hi',
```
mojoaxel: Sorry, but I personally don't like the "hi". Maybe just "label" !?

mojoaxel:
@ZhengxianLan You can just make new commits to your branch. This pull-request gets then updated automatically.

mojoaxel:
This PullRequest is in conflict with the changes made in #2385. It should not be merged at this time. 

@eymiha Can you have a look at this. You are now the official expert ;-)

eymiha:
Sure, I'll take a look

eymiha:
Thinking about 'selected' objects and font changes...

Default, non-font.multi behavior probably shouldn't do anything but make the edge text bold and thicken the lines. The right change is probably to make a selected option some functions for a node and edge.

So, if present

*selected.label* function would be called in Label#_processLabel and Label#_drawText and be passed the canvas context and the selected state, free to make whatever changes to the context desired. If the developer wanted to change the font color, face, size, modifier, etc. then they could do whatever they'd like.

*selected.node* function would be called in Node(shape)#draw and be passed the canvas context and the selected state, free to make whatever changes to the context desired. The function would also return temporary state to override the border and shadow parameters when selected.

*selected.edge* function would be called in Edge#draw and be passed the canvas context and selected state, free to make whatever changes to the context desired. The function would also return temporary state to override the arrow parameters when selected.

This mechanism would allow group and individual behaviors, and different selection behaviors for nodes and edges. And since it's an option, we could continue to make the existing networks function as needed.

I'd prefer to see this in a pull request, offering more comprehensive behavior than a specialized action such as the font turning a 'red'. It needs to be an enabling feature offering a free path to satisfy any reasonable developer's intentions.

But that's just my opinion. I'm not the keeper of the keys to vis, so I defer to higher judgement. ;-)

eymiha:
Or I  can just fix the code with the change @ZhengxianLan suggests...

mojoaxel:
> I'd prefer to see this in a pull request, offering more comprehensive behavior than a specialized action such as the font turning a 'red'. It needs to be an enabling feature offering a free path to satisfy any reasonable developer's intentions.

Me too. Feel free to provide a solution if you find the time. Would be a nice addition to the new style possibilities you've created.

> But that's just my opinion. I'm not the keeper of the keys to vis, so I defer to higher judgement. ;-)

We are constantly searching for new members of the holy visjs-core circle. So if you want a key to the "inner halls of endless work" just keep the pull-requests coming and engage in answering issues ;-)
(_Honestly we could use your help! You can find us on gitter._)

ZhengxianLan:
@eymiha Thanks for you to review my code. I think your opinion is better than mine.
  If you could pick a time to change the feature as you've said,just do it. Thanks.

eymiha:
Thanks - I've got the changes in motion. The general mechanism works, I'm just building it out.

mojoaxel:
@eymiha is working of a new solution for this.

PR #2401, opened 2016-12-03
Fixes bug introduced by #2385

mojoaxel:
see https://github.com/almende/vis/pull/2385/files#r90764463

PR #2403, opened 2016-12-03
make sure a given element isn’t undefined before accessing properties

### LINKED ISSUE #2402: util.topMost function failing
When an element of the given 'pile' to test is undefined, the accessing the element's property fails (of course.)

eymiha:
fixes #2402 

the util.topMost functions searches a pile of candidates for the first that defines an accessor. But no check was being done for the presence of a candidate before checking. An added conditional now prevents this from happening.

### DIFF
--- a/lib/util.js
+++ b/lib/util.js
@@ -1488,17 +1488,19 @@ exports.topMost = function (pile, accessors) {
     accessors = [accessors];
   }
   for (const member of pile) {
-    candidate = member[accessors[0]];
-    for (let i = 1; i < accessors.length; i++){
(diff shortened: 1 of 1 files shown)


mojoaxel:
I think I prefer #2401 :wink:

eymiha:
take #2401, the it's all the same!  :-)

mojoaxel:
@eymiha It's your bug I'm ok with you fixing it!

eymiha:
Thanks! I've been using the constraints with multiple fonts and it's all really working nicely for me.

PR #2411, opened 2016-12-05
added code-climate checks for develop branch

mojoaxel:
Code-Climate provides a nice opportunity to get a feeling for eslint and csslint status.
I would like to integrate linting into the build process but before that we should get the major issues out of the way.
Rules that make no sense for use should be disabled.

### DIFF
--- a/.codeclimate.yml
+++ b/.codeclimate.yml
@@ -1,0 +1,26 @@
+languages:
+  JavaScript: true
+engines:
+  eslint:
+    enabled: true
+  csslint:
+    enabled: true
+  markdownlint:
+    enabled: true
+  duplication:
+    enabled: true
+    config:
+      languages:
+        javascript:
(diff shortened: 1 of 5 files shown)


yotamberk:
So how do we use this? I'm not sure how to check out this PR

mojoaxel:
@yotamberk Sorry my mistake of not being clear. You can go to [the codeclimate](https://codeclimate.com/github/almende/vis/) website and see issues found by different lint-tools. 
My change removes quite some tests so that only "real" show up.
You can not realy test this PR...

mojoaxel:
@yotamberk Thanks!

Now we can start adressing those issues: https://codeclimate.com/github/almende/vis/compare/develop

yotamberk:
I'll have to get in to it. Not sure yet the benefits. Looks cool though!

Issue #1847, opened 2016-05-12
[Network] Method "getNodeAt" is not working at click on selected node


Alexthetiger:
Method "getNodeAt" is not working properly inside "click" event at click on selected node. Thereby I haven't information about click target. Example: I click on selected node and in "click" event callback I call "this.visNetwork.getNodeAt(obj.pointer.DOM)" it returns undefined.

spurdy:
Looks like the 'unselect' and 'select' actions that fire before the getNodeAt method would be called removes the node.shape.width and node.shape.height which causes issues for a method that getNodeAt() calls. You can get around the issue by calling a redraw(). Here's a fiddle that shows both the problem and the work-around:
[JS Fiddle](https://jsfiddle.net/qwhzofqq/1/)

Alexthetiger:
Yes, it works, thanks.

wimrijnders:
It should be noted here that the parameter of the callback in `network.on('click', function(obj){...})` also contains the node id clicked on.

```
network.on('click', function(obj){ 
  var nodeId = this.getNodeAt(obj.pointer.DOM);   // returns 'undefined' due to bug

  nodeId = obj.nodes[0];      // the node id that getNodeAt() should be returning
}); 
```

This is, in my opinion, the preferred solution to getting the node id within an event listener. It will work with almost all event listeners; an exception is `showPopup`, which returns the node id directly in parameter `obj`.

bradh:
Is this now fixed pending release?

wimrijnders:
@bradh Yes, this is fixed by #2885.

PR #2434, opened 2016-12-09
Check for the need to redraw

spurdy:
Adds a check for shape.width and shape.height in the 'isOverlappingWith' method, and issues a redraw if they are undefined. I believe this fixes issue [1847](https://github.com/almende/vis/issues/1847) that was causing the 'getNodeAt' method to not work in many of the event listeners.


yotamberk:
Thanks for the PR! Can you are submit it please to the develop branch?

spurdy:
Will do!

PR #2435, opened 2016-12-09
Redraw when node shape.width and shape.height is null

spurdy:
Adds a check for shape.width and shape.height in the 'isOverlappingWith' method, and issues a redraw if they are undefined. I believe this fixes issue [1847](https://github.com/almende/vis/issues/1847) that was causing the 'getNodeAt' method to not work in many of the event listeners.


yotamberk:
I'm sorry but still no good. It seems that youve done your changed on a master branch. You will have to make your changed on your develop branch and resubmit to the develop branch so your PR doesnt include irrelenant changes...

PR #2436, opened 2016-12-09
Adding width and height check

spurdy:
Adds a check for shape.width and shape.height in the 'isOverlappingWith' method, and issues a redraw if they are undefined. I believe this fixes issue #1847 that was causing the 'getNodeAt' method to not work in many of the event listeners.


mojoaxel:
@spurdy Thanks! That really looks like a nice fix!
But to be honest I'm not sure if a redraw is the right way to solve this problem. Redraws are usually expensive (big networks). We should figure out WHY the width and the height of the shape is null in the first place. 
We should try together to figure it out for the next release. If we did not find a better sone in the next days, we should acceppt your solution! Thanks again!

mojoaxel:
Sorry, but I still don't feel comfortable merging this. I'll move it to v4.19.

bradh:
Is this overtaken by #2885?

wimrijnders:
> @wimrijnders: Is this overtaken by #2885?

Yes, that is correct. #2885 fixes the concerns that @mojoaxel had with #2436.

bradh:
Overtaken by #2885.

Issue #1746, opened 2016-03-22
[Feature Request] [Timeline] Allow for "rolling mode"


BTMorton:
Can you implement a 'rolling mode' that ticks the background instead of the current time bar. The current time bar would be fixed in place.

josdejong:
I'm not sure what you mean but I guess you can implement it yourself using a method like `moveTo`?

BTMorton:
I made a quick codepen example of what I mean. Something like this: http://codepen.io/BTMorton/pen/NNwvYW

Obviously, to extend that and make it usable, I'd have to watch for any window drag changes and then store the new time offset and moveTo(Date.now() + timeOffset). My other concern is that doing this causes extra processing. If the onTickEvent is being called after the timebar has already moved, it will in effect be being moved once, and then moved back and the whole timeline adjusted with it. Whereas, with a built-in rolling mode, you could skip the moving back and just handle it all in the initial tick calculation (if that makes any sense)

yotamberk:
Added with #2439

PR #2439, opened 2016-12-10
Rolling mode

yotamberk:
adds #1746


yotamberk:
Can be done if the feature will be requested later =)

Issue #2445, opened 2016-12-13
vis.js is not compatible with YUI Compressor due to use of reserved words


easleydp:
Trying to use vis Timeline in a Sencha ExtJS application. The Sencha Cmd build tool uses YUI Compressor by default. It barfs on vis.js due to use of reserved words, e.g. "default" in the following:

`  function _interopRequireDefault(obj) { return obj && obj.__esModule ? obj : { default: obj }; }`

PS. I also tried using the Closure Compiler and this failed for some other reason (can't recall the details).

yotamberk:
Can you submit a PR to allow the support to these tools?

yotamberk:
solved with #2452! Thanks again @easleydp !

PR #2452, opened 2016-12-14
Fix YUI Compressor incompatibilities and improve JavaScript correctness

easleydp:
Solves #2445

Reason for changes: (1) To allow the Vis distributed code to be processed by YUI Compressor (e.g. via tools such as Sencha Cmd, where it can be difficult not to use a compressor). (2) Improve general JavaScript correctness.

Summary of changes:

1) Changes to Vis application code:
1.1) Avoid using "boolean" as a variable name because it is a JavaScript reserved word. (Used "bool" instead.)
1.2) Avoid using "char" as a variable name because it is a JavaScript reserved word. (Used "ch" instead.)

2) Changes to Babel configuration:
2.1) Include the 'transform-es3-property-literals' Babel plugin to ensure that reserved words are quoted in object property keys, e.g.:
`  function _interopRequireDefault(obj) { return obj && obj.__esModule ? obj : { 'default': obj }; }`
rather than:
`  function _interopRequireDefault(obj) { return obj && obj.__esModule ? obj : { default: obj }; }`
2.2) Include the 'transform-es3-member-expression-literals' Babel plugin to ensure that reserved words are quoted in property accesses, e.g.:
`    return new _Configurator2['default'](this, this.dom.container, configureOptions);`
rather than:
`    return new _Configurator2.default(this, this.dom.container, configureOptions);`

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -51,7 +51,8 @@ var webpackModule = {
       loader: 'babel',
       query: {
         cacheDirectory: true,
-        presets: ['es2015']
+        presets: ['es2015'],
+        plugins: ['transform-es3-property-literals', 'transform-es3-member-expression-literals']
       }
     }
   ],
--- a/lib/network/modules/components/shared/Label.js
+++ b/lib/network/modules/components/shared/Label.js
@@ -519,63 +519,63 @@ class Label {
       }
     }
     while (s.position < text.length) {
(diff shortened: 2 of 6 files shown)


eymiha:
I've detailed the problem and solution for issue #2500 as a comment there.

PR #2459, opened 2016-12-15
Allow per-item disabling of limitSize

grahamj:
I have an application for the vis.js timeline that requires item DIVs to always be their full size, however due to concerns over browser DIV size limitations there is code to keep them limited in size. Accordingly this patch allows the selective disabling of this limiting by adding `limitSize: false` to an Item's data.


yotamberk:
@grahamj Can you please resubmit this PR to the `develop` branch? We cannot accept this PR like this...
It would be great if you could also add an example and to the docs.

PR #2460, opened 2016-12-15
Create .babelrc

grahamj:
When I was going through the documentation about creating custom builds I ran into the error outlined at the below link. Applying the suggested fix solved it so I'm submitting same here.

http://stackoverflow.com/questions/30386317/babelify-throws-parseerror-on-import-a-module-from-node-modules


yotamberk:
@grahamj Please resubmit this PR to `develop` branch. Thanks for the PR! We appreciate this addition, but cannot accept this PR like this.

mojoaxel:
@grahamj Thanks from me to. Please checkout https://github.com/almende/vis/blob/develop/gulpfile.js#L54 if we can integrate this into the gulpfile.js.

grahamj:
@mojoaxel I'm not sure what you mean about the gulpfile (I don't have much experience with gulp)

I noticed that the preset value is in the gulpfile; it was when running browserify through babelify via the command line that I ran into problems, and of course the gulpfile isn't involved in that case.

PR #2483, opened 2016-12-20
Graph3D 'dot-color' mode: explicitly set the dot color

ephraimberkovitch:
If the style value of dot is string, it is treated as color and directly passed to the graph;
otherwise, if it is a number, the previous algorithm is still in place

        var x = pow(random(), 2);
        var y = pow(random(), 2);
        var z = pow(random(), 2);
        var dist = "#00ff00"; //sqrt(pow(x, 2) + pow(y, 2) + pow(z, 2));


yotamberk:
Sorry but still no good. it seems that you did your changes on a wrong branch. You'll have to apply you additions on a new branch pulled from the `develop` branch. This PR is not reviewable... You've added files that should be in `develop`

PR #2493, opened 2016-12-22
add template support for minor/major labels

yotamberk:


Tooa:
Do we need some kind of documentation for this feature either via an example or mentioning it in the documentation?

yotamberk:
I think it's very rare that someone would use it, but if they do, they can. I don't think it need s documentation

ShawnMercado:
Is there any information about how to use this feature?

yotamberk:
@ShawnMercado  checkout this example http://visjs.org/examples/timeline/other/functionLabelFormats.html

PR #2495, opened 2016-12-22
Added showXAxis, showYAxis, showZAxis options on Graph3d

nmehrle:
Added showXAxis, showYAxis, showZAxis options on Graph3d
They all default to true.
Setting one option false prevents the drawing of the respective axis, axis label, and tick marks/tick mark labels. Setting all three to false and showGrid to false allows display of data only, without any supplementary lines


yotamberk:
Hey @nmehrle, please reapply your changes on a `develop` branch. We cannot accept this PR since it has chamges that are not relevant to this branch.

PR #2496, opened 2016-12-22
Added Graph3d options showXAxis,showYAxis,showZAxis

nmehrle:
Added showXAxis, showYAxis,showZAxis options to Graph 3d. Default to true. Setting one to false prevents drawing of respective axis, axis label, tick marks and tick mark labels. Setting all three to false and showGrid to false displays only data without supplementary lines

### DIFF
--- a/lib/network/options.js
+++ b/lib/network/options.js
@@ -16,29 +16,29 @@ let any = 'any';
 
 let allOptions = {
   configure: {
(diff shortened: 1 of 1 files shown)


yotamberk:
@nmehrle Hey again. Please remove the dist files. They shouldn't be added.

Issue #338, opened 2014-09-23
HTML tooltips


carueda:
This is about allowing html in token tooltips (complementing what the current 'title' does with only text), and on groups.  I know there are some "tooltip" entries already but I was not sure whether any of them is for the timeline component; feel free to mark this as duplicate if that's the case.  Thanks.

AlexDM0:
Hi,

Over the last year a lot of feature requests have been made. We have just introduced our new website which has a list of the requested features. We have placed this request on that list. 

The list can be found here:
http://visjs.org/featureRequests.html

An explaination of the new system can be found here:
http://visjs.org/blog.html#New\ website\ for\ vis\.js\!

I would like to stress that this does not mean we abandon this request. Discussion here will continue if needed on this feature but we will close it to keep our Github issue page more of a bug-todo list.

Future feature requests will still be made here and then added to the website by us.

Regards,

Alex

jasontibbitts:
I just happened to need this and ran across this closed issue.  I managed to get something that at least works for my use case.  Just pull in jquery and powerTip, do
```javascript
$('.group .item').powerTip({placement: 'se', smartPlacement: true, mouseOnToPopup: true});
```
(or whatever selector you need to get the items you want) and then the title property on every element can be any HTML you want.  I'm doing a little security camera system and the elements on the timeline are videos.  The tooltips display small versions of the videos, with controls and everything.  Seems to work fine, even with zooming and scrolling.

Maybe this is trivial or pointless, since I don't really know what I'm doing, but at least I have something going.

jasontibbitts:
Meh; this for some reason doesn't work if you call setGroups, even if you just initially set up the timeline that way.  It doesn't appear that the structure of the elements changes if you do that, and powerTip attaches the events to all of the items properly, but either the tooltip doesn't fire, or it fires _and_ the title tooltip appears.  Looks like I need to find another way to do what I wanted to do.

sekys:
Hello jason, well this is not working because you are selecting existing items. For you and all, i suggest use "bind" function. Then you are selecting existing and future-existing elements. Good luck

jasontibbitts:
Thanks for the hint.  I'm not sure that powertip code exposes a way to use bind; it appears to traverse the selected elements itself and attaches events with this.on().  Though, it does allow that to be overridden.  I'll have to see why it does appear to attach events to new elements but doesn't manage to get its tooltip magic correct.

phamtuananh791994:
Are any solution for this ?

[...]

PR #2498, opened 2016-12-23
[Timeline] Added HTML tool-tip support

lewisjb:
Adding #338 

Tested on Chrome version 55.0 and Internet Explorer version 11.576

This change will allow developers to specify a HTML 'title' (hover tool-tip) for timeline chart items.

Tried to keep code changes to a minimum, and implement it similarly to how Network does it.

The only possible issue with this that I can think of is changing the class of the tool-tip from 'vis-network-tooltip' to 'vis-tooltip' (if people were styling their network tooltip).

### DIFF
--- a/examples/timeline/items/tooltip.html
+++ b/examples/timeline/items/tooltip.html
@@ -1,0 +1,49 @@
+<!DOCTYPE HTML>
+<html>
+<head>
+  <title>Timeline | Tooltips</title>
+
+  <style type="text/css">
+    body, html {
+      font-family: sans-serif;
+      max-width: 800px;
+    }
+  </style>
+
+  <script src="../../../dist/vis.js"></script>
+  <link href="../../../dist/vis-timeline-graph2d.min.css" rel="stylesheet" type="text/css" />
+  <script src="../../googleAnalytics.js"></script>
+</head>
+<body>
+
+<h1>Tooltips</h1>
+
+<p>
+  Setting the tooltip in various ways.
+</p>
+
+<div id="tooltips"></div>
+
+<script type="text/javascript">
+  // Create a DataSet (allows two way data-binding)
+  var items = new vis.DataSet([
+    {id: 1, content: 'Item 1', start: '2016-01-01', end: '2016-01-02',
+      title: 'Normal text'},
+    {id: 2, content: 'Item 2', start: '2016-01-02', title: '<b>Bold</b>'},
+    {id: 3, content: 'Item 3', start: '2016-01-03', type: 'point',
+      title: '<span style="color: red">Red</span> text'},
+    {id: 4, content: 'Item 4', start: '2016-01-03', end: '2016-01-04',
+      title: '<table border="1"><tr><td>Cell 1</td><td>Cell 2</td></tr></table>'}
+  ]);
+
+  // Options
+  var options = {};
+
+  // Timeline object
+  var timelineTooltips = new vis.Timeline(document.getElementById('tooltips'),
+      items, options
+  );
+</script>
+
+</body>
+</html>
--- a/examples/timeline/items/tooltip.html
+++ b/examples/timeline/items/tooltip.html
@@ -32,7 +32,7 @@
     {id: 2, content: 'Item 2', start: '2016-01-02', title: '<b>Bold</b>'},
     {id: 3, content: 'Item 3', start: '2016-01-03', type: 'point',
       title: '<span style="color: red">Red</span> text'},
-    {id: 4, content: 'Item 4', start: '2016-01-03', end: '2016-01-04',
+    {id: 4, content: '<h1>HTML</h1> Item', start: '2016-01-03', end: '2016-01-04',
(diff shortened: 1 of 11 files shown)


yotamberk:
Cool! thank you! Can you add an example?

lewisjb:
Done @yotamberk.
I also tested it on Firefox and found that the `mouseover` event doesn't have a plain `.x` and `.y`, so I have fixed that aswell.
(https://developer.mozilla.org/en/docs/Web/Events/mouseover)

lewisjb:
Thanks @yotamberk!
I have just pushed a change that will fix the odd tool-tip location so that it is next to where the cursor entered the element.
The naming is unfortunate, 'title' is used so that it isn't a breaking change, and is also what is used in Network (the main reference for this implementation).

yotamberk:
@lewisjb I think there is still something weird with the location... I think I preferred it better the way it was before or with a margin of a few pixels from the item item and not exactly where the mouse is. 
There should be further work with setting for the tooltip (e.g. allowing it to follow the mouse position while hovering the item, tooltip position, trigger control... etc.) 
What do you think?

lewisjb:
@yotamberk Regarding the current way versus the previous way, I'd say that the supplied example doesn't do it justice. If the timeline chart was much larger, then I'm certain the current way would look far better (or, more accurately, the previous way would look far worse).

As for following the mouse, I can totally do that. The reason I went with it saying at it's initial location is because that is how Google Chart's Timeline does it. It might be beneficial to add it as an option to Timeline, something like `tooltipFollowMouse`. Thoughts?

yotamberk:
Okay I can see what you mean. The current way is good for now. If you can add that option that would be great =)

examples/timeline/items/tooltip.html
```diff
@@ -0,0 +1,49 @@
...
+<script type="text/javascript">
+  // Create a DataSet (allows two way data-binding)
+  var items = new vis.DataSet([
+    {id: 1, content: 'Item 1', start: '2016-01-01', end: '2016-01-02',
+      title: 'Normal text'},
+    {id: 2, content: 'Item 2', start: '2016-01-02', title: '<b>Bold</b>'},
+    {id: 3, content: 'Item 3', start: '2016-01-03', type: 'point',
+      title: '<span style="color: red">Red</span> text'},
+    {id: 4, content: 'Item 4', start: '2016-01-03', end: '2016-01-04',
+      title: '<table border="1"><tr><td>Cell 1</td><td>Cell 2</td></tr></table>'}
```
mojoaxel: Maybe you could add also an example for "HTML element".

lewisjb: @mojoaxel Do you mean adding an example that uses http://visjs.org/examples/timeline/items/htmlContents.html aswell as a HTML tool-tip?

yotamberk: I think he means a seperate example with an html tooltip

lewisjb: @yotamberk I'm sorry, I don't seem to understand... These examples have HTML tool-tips (except for the first one as it shows you don't need to use HTML)

yotamberk: True... maybe change the content of an item to html element and then i think the example is complete

lewisjb:
@yotamberk I've changed the content of an item to use HTML in the example. However this brought to light a bug.

Moving your mouse around over the HTML item moves the tool-tip weirdly. This is because the 'mouseout' and 'mouseover' events are called on children elements of the item. I personally see this as a bug with Timelines as a whole, not just HTML tool-tips*. I would be happy to fix it in this pull request, but I think it might be better to have it as a separate GitHub issue & PR (assuming it is a bug with Timeline and not just HTML tool-tips)

yotamberk:
@lewisjb Okay I think it can be merged and have the other features and issues we talked about above in later issues (It would be great if you could open these issues).
What do you mean as a problem in 
>I personally see this as a bug with Timelines as a whole, not just HTML tool-tips. 

?

lewisjb:
@yotamberk Thanks! Will do!

As for why it's a bug with Timelines as a whole, not just HTML tool-tips:
The `mouseover`/`mouseout` event is being fired on child elements. In this case it was something like
```html
<div class='vis-item'>
    <h1>HTML</h1>
    Text
</div>
```
So when you hovered over the `<h1>` element, there was a new element under the cursor, so `mouseout` is called, followed by its related `mouseover`. This is expected; however, all the code handling those events don't care about the specific element, only the 'item' which they are in. (Reference: [ItemSet.js:1874](https://github.com/almende/vis/blob/develop/lib/timeline/component/ItemSet.js#L1874) and [ItemSet.js:1896](https://github.com/almende/vis/blob/develop/lib/timeline/component/ItemSet.js#L1896))
So tool-tips will be moved, and 'itemover'/'itemout' will be emitted when the cursor is still over the same item. Which, goes against the description of 'itemover'/'itemout' in the docs. If, for some reason this is the intended behavior, then only tool-tips will need to be fixed, and I would recommend updating the docs to reflect this behavior.

PR #2491, opened 2016-12-22
Don't pass non-string values to Date.parse

jimmed:
Fixes: https://github.com/almende/vis/issues/2490

PR #2512, opened 2016-12-29
Revert "Don't pass non-string values to Date.parse"

mojoaxel:
Reverts almende/vis#2491
This commit merged the master back into develop. Must be reverted!


yotamberk:
Thanks @mojoaxel , It was my mistake merging it. I didn't see the pull from master (I still don;t see it in the original PR... Weird...) Thanks for fixing this.

PR #2517, opened 2016-12-31
added badges from isitmaintained.com

mojoaxel:
just added two new badges and changes order of the badges.


yotamberk:
Cool! I love it!

PR #2530, opened 2017-01-03
[DataSet] Fixed deleting item with id 0 (#2528)

### LINKED ISSUE #2528: Node with id=0 cannot be deleted
Hi, 

I have a very simple app created with Vis and the problem is that the node with id=0 cannot be deleted. All of the other nodes can be deleted. 

Here is the example: https://jsfiddle.net/Filip_Z/9k2s3dgj/7/

```
    // create an array with nodes
    var nodes = new vis.DataSet([
        { id: 0, label: '0' },
        { id: 1, label: '1' },
        { id: 2, label: '2' },
        { id: 3, label: '3' },
        { id: 4, label: '4' }
    ]);

    // create an array with edges
    var edges = [];

    // create a network
    var container = document.getElementById('mynetwork');

    // provide the data in the vis format:
    var data = {
        nodes: nodes,
        edges: edges
    };

    var options = {
        manipulation: {
            enabled: true,
            initiallyActive: true,
            deleteNode: function(deleteData, callback) {
                callback(deleteData)
            }
        }
    };

    // initialize your network!
    var network = new vis.Network(container, data, options);
```

As @danieldiekmeier [suggested](https://github.com/vuejs/vue/issues/4622#issuecomment-269956193), this is most likely because Vis checks for truthy-ness in the ids (because if you use a string as the id, it works, and you can remove the node):
```
nodes: [
    { id: '0', label: '0' },
    { id: 1, label: '1' },
    { id: 2, label: '2' },
    { id: 3, label: '3' },
    { id: 4, label: '4' }
]
```

lewisjb:
Fixes #2528

### DIFF
--- a/lib/DataSet.js
+++ b/lib/DataSet.js
@@ -679,7 +679,7 @@ DataSet.prototype.remove = function (id, senderId) {
     item = this._remove(ids[i]);
     if (item) {
       itemId = item[this._fieldId];
-      if (itemId) {
+      if (itemId != undefined) {
         removedIds.push(itemId);
         removedItems.push(item);
       }


Tooa:
Big thanks for you pull request. Looks good to me :)

PR #2533, opened 2017-01-03
Gh pages play ground feature

mosheDa:


mojoaxel:
To be honest, I'm not convinced that this is a necessary feature. There are multiple much more sophisticated tools like "jsbin" or "jsfiddle" out there.

Also the example-code should definitely not been hard coded, but load the code from the existing examples or something.

@yotamberk @Tooa @eymiha What do you guys think about this playground feature?

yotamberk:
We've decided to close this PR for now

PR #2535, opened 2017-01-03
Fix for Remove Events triggered by DataView Refresh()

ZacBrownBand:
If found an issue with the DataView refresh function.
When items are removed after a call to refresh the oldData property of the remove event does not contain the correct events.

The items in the set are stored in `_data._data` not `_data`. The result was that the remove event's `oldData` property was an array of `undefined`.

_Note: looks like I introduced this issue in [this pr](https://github.com/almende/vis/pull/2189)._


yotamberk:
Thanks!

Issue #2560, opened 2017-01-10
IE11 - 'Symbol' is undefined when trying to use network


dockstreet:
Hi,

In validating another issue https://github.com/almende/vis/issues/1911

discovered that snapshot in develop 4.18 works fine in Edge/Chrome/Firefox, but throws an error in  IE11 when trying to instantiate vis Network

'Symbol' is undefined is the error

Plunker created to reproduce the error 
https://plnkr.co/edit/MX1agpoRYgYbgC23NAhl?p=preview

Tooa:
Maybe one of this issues helps https://github.com/angular/angular/issues/7914 https://github.com/AngularClass/NG6-starter/issues/114 ?

dockstreet:
Looks like it could. I wonder...  Source map was added after the last 4.17 release. 
There is something here as well, seems like a polyfill could help fix this:

https://github.com/Financial-Times/polyfill-service/issues/715

dockstreet:
Ok. Update... found a fix for IE11. What I did was 

npm install babel-polyfill --save
npm install --save-dev babel-plugin-transform-runtime
npm install --save babel-runtime

and npm run build and copied the minified vis.js.min to the plunker I was testing with in IE11, it then worked fine. 

What I believe happened was the babel-loader was added in this commit, which works fine except for IE11

https://github.com/almende/vis/commit/884b75ee43065b3bae36a49a46414edb893831fb#diff-b9e12334e9eafd8341a6107dd98510c9

dockstreet:
@Tooa / @mojoaxel 
it may be of benefit to add this polyfill into the v4.18 release milestone?

perhaps for IE11  and from what I've been reading (but not tested) Safari compatibility and it looks like the solution is straightforward by adding the dependencies for babel

PR #2566, opened 2017-01-11
address IE11 issue 'Symbol' is undefined with babel-polyfill 

dockstreet:
Hi this is to address - 

https://github.com/almende/vis/issues/2560


mojoaxel:
@dockstreet Thanks a lot! It's quite heavy wait but should get rid of some brower incapabilities.

_Just an Info: The change increases the filesize of `dist/vis.js` by 22,8kB (1,34%)._

PR #2582, opened 2017-01-15
Fix #2522

### LINKED ISSUE #2522: Right button (contextmenu) while dragging item makes items uneditable
It seems to be because of 
https://github.com/almende/vis/pull/2427/files#diff-d3c0773c9b1eaa102ae263e2f400774dR1216

yotamberk:
solves #2522


yotamberk:
Clicking outside the timeline seems to not be a problem. I had a bug in one of my playgrounds. Now everything seems to work as expected (once I pull from develop to my playground)

PR #2593, opened 2017-01-17
vertically center label in network circle node

### LINKED ISSUE #2589: Problem with text inside circle ( network )
i can't get the text inside circle to center after updating vis, it look's like the following:

capture

eymiha:
fixes #2589 

vertically align the label text in a circle to center.


mojoaxel:
@eymiha Thanks so much!

PR #2595, opened 2017-01-17
self reference edge should now appear in all cases

### LINKED ISSUE #2591: [Network] self-loop edge not always show with 4.18
When network have "self-loop" edge (from = 1, to = 1), they are not always showed, depends of movement : 
net2png

eymiha:
fixes #2591 

some calls to extract circle data were passing invalid arguments.

### DIFF
--- a/examples/network/basicUsage.html
+++ b/examples/network/basicUsage.html
@@ -37,7 +37,8 @@
     {from: 1, to: 3},
     {from: 1, to: 2},
     {from: 2, to: 4},
-    {from: 2, to: 5}
+    {from: 2, to: 5},
+    {from: 3, to: 3}
   ]);
 
   // create a network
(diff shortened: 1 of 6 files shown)


Tooa:
This does not work with the example provided in https://github.com/almende/vis/issues/2591 [here](http://jsbin.com/fipaha/edit?html,console,output). Apply your fix and you will see that the self-reference edges disappear when moving the nodes.

eymiha:
On a Mac... `npm run watch` then launch Chrome (mine is _Version 55.0.2883.95 (64-bit)_) and open examples/network/basic_usage.html...

The circular link in the network is present whether a the self-referenced node is moving, or another node is moving. If there is no circular link (node 3 to itself) then you're not getting the right files.

If this works, but the example in #2591 doesn't, then there's something else amiss.

eymiha:
Got it - for some reason node.shape width is NaN, not undefined, going into _getCircleData, and then _resizeShape. Adding the checks into the pull request.

PR #2606, opened 2017-01-19
maximum width constraints must be violated by long words

### LINKED ISSUE #2604: [Network] - Setting widthConstraint.maximum to node causes error
**version: 4.18.0**
When trying to add the widthConstraint maximum attribute I receive an error in the console:
"Uncaught TypeError: Cannot read property 'width' of undefined"

Works fine with widthConstraint.minimum.
 ```
   var nodes = new vis.DataSet([
        {id: 1, label: 'Node 1', shape:'circle', widthConstraint : { maximum : 30}}
    ]);
```

eymiha:
fixes #2604 

When a word in a label is longer than the effective maximum width, the constraint must be violated to avoid an infinite loop in composition.

The network/other/widthheight example has been updated to assure working constraints. A test has been added to force the edge case handling.

### DIFF
--- a/examples/network/_tests/maximumWidthEdgeCase.html
+++ b/examples/network/_tests/maximumWidthEdgeCase.html
@@ -1,0 +1,82 @@
+<!doctype html>
+<html>
+<head>
+  <title>Maximum Width Edge Case Test</title>
+
+  <script type="text/javascript" src="../../../dist/vis.js"></script>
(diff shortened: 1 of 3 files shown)


eymiha:
I was just thinking that it might be nice to have a set of "test documents" that we can use, but aren't bundled in the release. They may help contributors diagnose problems moving forward. If not, you can just get rid of the doc and directory.

mojoaxel:
changes moved to #2650

PR #2611, opened 2017-01-20
change setCustomTimeTitle title parameter

cmolinaAlicante:
Now could be an string or a function


lib/timeline/component/CustomTime.js
```diff
@@ -143,6 +143,8 @@ CustomTime.prototype.redraw = function () {
   if (title === undefined) {
     title = locale.time + ': ' + this.options.moment(this.customTime).format('dddd, MMMM Do YYYY, H:mm:ss');
     title = title.charAt(0).toUpperCase() + title.substring(1);
+  }else if(typeof title === "function"){
```
Tooa: There is something weird with the indent and spacing of the brackets. Can you insert a space between the braces and the curly brackets i.e. `) {` and `} else if`? Also please fix the indent for the body of the if statement.

cmolinaAlicante: Fixed!
Sorry I have other rules for indent and spaces in Webstorm.

mojoaxel:
Looks good, But let's get the patch release 4.18.1 out first!

PR #2615, opened 2017-01-23
fix #2613 with if statement, and null for pointer value rather than r…

### LINKED ISSUE #2613: Network - Navigation Zoom is broken (4.18)
After #2377 commit to 4.18 Zoom events for Navigation at Network module are broken. Example of it is active at official example page - http://visjs.org/examples/network/other/navigation.html

`Uncaught ReferenceError: pointer is not defined` - when trying to zoom in or zoom out with help of bottom buttons.

CapitanMorgan:
with if statement, and passing null for pointer value rather than referencing it when undefined

### DIFF
--- a/lib/network/modules/components/NavigationHandler.js
+++ b/lib/network/modules/components/NavigationHandler.js
@@ -160,7 +160,12 @@ class NavigationHandler {
 
     this.body.view.scale = scale;
     this.body.view.translation = { x: tx, y: ty };
-    this.body.emitter.emit('zoom', { direction: '+', scale: this.body.view.scale, pointer: pointer });
+    if (typeof(pointer) != 'undefined') {
+      this.body.emitter.emit('zoom', { direction: '+', scale: this.body.view.scale, pointer: pointer });
+    } else {
+      this.body.emitter.emit('zoom', { direction: '+', scale: this.body.view.scale, pointer: null });
+    }
+
   }
   _zoomOut()  {
     var scaleOld = this.body.view.scale;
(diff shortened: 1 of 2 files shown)


lib/network/modules/components/NavigationHandler.js
```diff
@@ -160,7 +160,12 @@ class NavigationHandler {
 
     this.body.view.scale = scale;
     this.body.view.translation = { x: tx, y: ty };
-    this.body.emitter.emit('zoom', { direction: '+', scale: this.body.view.scale, pointer: pointer });
+    if (typeof(pointer) != 'undefined') {
```
Tooa: Why do we need this conditional expression here? There is no `pointer` in the scope, so this expression always evaluates to `false`, right?

lib/network/modules/components/NavigationHandler.js
```diff
@@ -172,7 +177,11 @@ class NavigationHandler {
 
     this.body.view.scale = scale;
     this.body.view.translation = { x: tx, y: ty };
-    this.body.emitter.emit('zoom', { direction: '-', scale: this.body.view.scale, pointer: pointer });
+    if (typeof(pointer) != 'undefined') {
```
Tooa: Same as above.

CapitanMorgan:
Ill remove the if, i just through it together hastily and thought the previous person who added this pointer reference, actually had a scenario were it might not be null. Ill just pass null, not sure what portion of doc needs updating.

PR #2620, opened 2017-01-23
Fix typo

ypid:

PR #2618, opened 2017-01-23
[Docs] Spelling correction

jakeboone02:
Corrected DatSet to DataSet

PR #2623, opened 2017-01-23
Develop update PR contributing

ypid:
Related to: #2618
Related to: #2620 

@yotamberk PS: Note that the way you merged https://github.com/almende/vis/pull/2622 destroyed my GnuPG signature I made over this commit and thus implicitly signing the whole git history of this project. Ref: https://github.com/almende/vis/commit/39bcadda3d2b63c9a3d46bc03a5da211e0245de0


yotamberk:
What got destroyed?

ypid:
This was my original commit https://github.com/almende/vis/pull/2622/commits/0014d909e65dfc482fcafcf1b3f9334749712d36 which has been recommitted as https://github.com/almende/vis/commit/39bcadda3d2b63c9a3d46bc03a5da211e0245de0. Note that GitHub shows a "Verified" badge for the first, while for the later it does not. Refer to https://help.github.com/articles/signing-commits-using-gpg/

mojoaxel:
I'll check it out.

mojoaxel:
> This was my original commit 0014d90 which has been recommitted as 39bcadd. Note that GitHub shows a "Verified" badge for the first, while for the later it does not.

@ypid It looks like a problem of githubs "squash" functionality that we use by default. I think in this case we could have single-merged your commit, but this is not our usual process. Sorry but it looks like we could not make use of signed commits at this time. If you see a way of doing that in the future please let us know!

mojoaxel:
@ypid This PR is going to have the same problem. We have two commits, b5037aca45aa8072173daf616492bbc657d98f1f and the branch-update merge commit dbc246ca73d932eb9d3155b7352e3a846f73d9d8. We don't want those two commits to show up in your history, therefore we "squash" them. This will again ruin your signature. As I sad: If you have an idea how to prevent this let us know!

ypid:
Thanks for taking the time to check this. I guess there are two options. Either normal merge commits or fast forward merge. The later would require me to rebase onto the latest develop commit so that a ff merge is possible. The first variant would be the more common and preferred way. Anyway, if you want you can just merge the usual way. Keep up the good work!

Issue #2494, opened 2016-12-22
Directed Hierarchy not layed out correctly 


RyanAldridge:
Hi,

When representing a fairly simple org chart as a hierarchy, nodes are laid out nicely in hubsize sort method but as soon as I switch to a directed sort method the lowest nodes in the branch suddenly don't appear under their parent node and edges criss-cross all over the place.

I'm using the following layout options and just changing the sortmethod:

layout: {
            randomSeed: undefined,
            improvedLayout:true,
            hierarchical: {
                enabled:true,
                levelSeparation: 150,
                nodeSpacing: 100,
                treeSpacing: 200,
                blockShifting: true,
                edgeMinimization: true,
                parentCentralization: true,
                direction: 'DU',        // UD, DU, LR, RL
                sortMethod: 'directed'   // hubsize, directed
            }
        },

Hubsize
![hubsize](https://cloud.githubusercontent.com/assets/24721755/21436674/0460c004-c880-11e6-92d7-395b65779aa9.png)

Directed
![directed](https://cloud.githubusercontent.com/assets/24721755/21436673/0460aac4-c880-11e6-85f7-c32f31d4d61c.png)

Any advice is appreciated.

LLTommy:
From my experience, just to change the sort method is often not enough. Also try to play with other parameters (e.g. mass of node and edges) to get better results. Unfortunately, there is not one set of options that works for all cases, so I ended up with a brute force 'try and error' approach. 

Maybe you kind find additional hints by studying old issues, e.g.: https://github.com/almende/vis/issues/1453

JoshDreamland:
Same problem, here. I get criss-crossing when one tree takes up too much horizontal space (it forces the same row of another tree into the next tree).

I've pastebinned an example, [here](http://pastebin.com/yxnvJPNY). It's fairly simple, and renders almost perfectly, with the exception of the right two trees crossing one another. Reference render [here](http://i.imgur.com/W7NPndP.png).

As with the above example, using the default "hubsize" sorting works fine. I would be able to work around this if initial x-coordinates were taken into consideration when performing the layout.

It's possible that the root of the problem is that the algorithm is incorrectly identifying independent (non-connected) networks, as setting a tree separation in the thousands does not solve this issue—it simply moves the left tree further away from the right two trees, which remain tangled.

I may take a look into what is actually going on over the weekend.

RyanAldridge:
[...]

PR #2625, opened 2017-01-23
Network: Fix tree collision in hierarchical layout

JoshDreamland:
Using the 'directed' sort method, the layout engine seemed to ignore treeSpacing (or only apply it to the first tree). This would also lead to trees colliding with one another, and having their nodes tangled in space. This complaint surfaced in a couple bugs, including #2494. This should be a step toward fixing that. For whatever reason, this code never runs when using the "hubsize" sort method.

### DIFF
--- a/lib/network/modules/LayoutEngine.js
+++ b/lib/network/modules/LayoutEngine.js
@@ -381,9 +381,11 @@ class LayoutEngine {
     // the main method to shift the trees
     let shiftTrees = () => {
       let treeSizes = getTreeSizes();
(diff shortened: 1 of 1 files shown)


joe-dirte:
Any idea when this will be merged to master?

yotamberk:
in a few weeks when i can find time to do it.
I am currently working alone maintaining the project and have my own job and projects.

joe-dirte:
Thanks Yotam , Wasn't trying to sound pushy was just curious.  Thanks for the great product!

Issue #2353, opened 2016-11-22
babel -> babel-loader ?


eymiha:
To get my builds building on my work box - a newer mac than my home box, with newer npm stuff (et al.) - I had to change gulpfile.js from `loader: 'babel',` to `loader: 'babel-loader'` to keep from getting a node error during the build:

> ... before babel was installed,
> ```
> Module not found: Error: Cannot resolve module 'babel' in ...
> ```
> after babel was installed,
> ```
> Error: The node API for 'babel' has been moved to 'babel-core'.
> ```
> after babel-core was installed,
> ```
> Error: The node API for 'babel' has been moved to 'babel-core'.
> ```
> when babel was removed,
> ```
> Module not found: Error: Cannot resolve module 'babel' in ...
> ```

During the process I ended up installing some node modules globally, and others just in the vis space. Cleaning the node_modules and installing today, with the loader set to babel, the problem from didn't recur - but it took me a bit to figure out what was giving me the problem before. I can only imagine that I have something set globally now that I hadn't before and I won't see the problem ever again (knocking-on-wood...)

I'm not sure what might be done to bridge the gap (some may break with it, others without) but it's good to know what must be done to fix the problem when it happens.

yotamberk:
@mojoaxel Can you take a look at this?

mojoaxel:
I use the following in a project of min:

### package.json
```json
"devDependencies": {
  "babel-core": "^6.11.4",
  "babel-loader": "^6.2.4",
  "babel-preset-es2015": "^6.9.0"
}
```

### webpack.config.js
```js
module: {
  loaders: [{ 
    test: /node_modules[\\\/]vis[\\\/].*\.js$/, loader: 'babel',
      query: {
        cacheDirectory: true,
        presets: ["es2015"]
      }
    }]
},
extensions: ["", ".js"],
alias: {
  visTimeline: "vis/index-timeline-graph2d"
}
```

Than I just can import `visTimeline` in my project.

mojoaxel:
@eymiha Do you still have problems with babel or can we close this issue?

eymiha:
No problems since the initial ones. The issue was mainly added in case
others had seen the same problem.

PR #2629, opened 2017-01-25
moved babel plugins from devDependencies to dependencies

mojoaxel:
If you want to integrate the vis.js sources in a project you need some babel-pugins.

see #2353


yotamberk:
Nice! Good idea and very important! I'm surprised we haven't had this option before.

PR #2631, opened 2017-01-25
Fixes instanceof Object statements for objects from other windows and iFrames.

### LINKED ISSUE #2097: DataSet.update doesn't work for objects from iframe (gwt usage issue)
https://github.com/almende/vis/blob/master/lib/DataSet.js checks data parameter in update method using data instanceof Object. It won't work if data has been created in iframe.
Here is explanation:
https://groups.google.com/forum/#!topic/google-web-toolkit/CdgvqajYeDI

Proposal if to move instanceof checks to separate class as util methods and introduce additional check using toString method, like:
Object.prototype.toString.call(o) === '[object Object]

Thanks

ZacBrownBand:
Fixes #2097

I am proposing that the use of ` o instanceof Object` be replaced with `o && type of o === 'object'`.

I have only updated the instances in DataSet, I am not sure if these changes are needed elsewhere.


lib/DataSet.js
```diff
@@ -707,7 +707,7 @@ DataSet.prototype._remove = function (id) {
   if (util.isNumber(id) || util.isString(id)) {
     ident = id;
   }
-  else if (id instanceof Object) {
+  else if (id && typeof id ==='object') {
```
Tooa: Here is an empty space missing after `===`.

ZacBrownBand:
Pesky whitespace!

There are more cases of this, but I don't think the checks will be on objects passed from external sources like in DataSet.

PR #2670, opened 2017-02-02
chore: removed google-analytics from all examples

mojoaxel:


yotamberk:
Thanks for doing this. It should've happened a long time ago...

Issue #242, opened 2014-08-13
Clustering of items.


Style87:
I'm looking to add clustering to the timeline as was in it previously. Would you be able to provide me with a starting location or brief comments on what needs to be added? This codebase is rather large and I'm quite lost.

AlexDM0:
Hi,

Over the last year a lot of feature requests have been made. We have just introduced our new website which has a list of the requested features. We have placed this request on that list. 

The list can be found here:
http://visjs.org/featureRequests.html

An explaination of the new system can be found here:
http://visjs.org/blog.html#New\ website\ for\ vis\.js\!

I would like to stress that this does not mean we abandon this request. Discussion here will continue if needed on this feature but we will close it to keep our Github issue page more of a bug-todo list.

Future feature requests will still be made here and then added to the website by us.

Regards,

Alex

antferr:
Why not make available the old clustering capabilities until the new it's ready? It was the best feature!

AlexDM0:
Because the code is completely incompatible.

Regards,

Alex

Wuber:
Hi josdejong and AlexDM0, 

We've been using the chap timeline for a while. It’s a great component, love it and good job!

We’ve recently looked into Vis because we need right-click and multiple select functionality. 
However, we are facing a major hurdle. We actually process large volume of events 
(i.e. thousands of events in very short time frame). The old chap timeline can handle this by clustering events. Since Vis doesn’t support clustering, it simply doesn’t render large events properly in short time frame. 

In summary, the performance (responsiveness and scalability) suffered without clustering. It’s become clear to us that clustering is a must have feature in order for us to move to Vis.

This said, it'd be great to have clustering capability as a high priority feature. Can you please provide  an update or estimate on this feature?

Once again, awesome work, thank you!

stevefurey:
+1 for clustering as well, been waiting for this feature for awhile and need it for all the reasons above.

An update on the progress would be appreciated.

josdejong:
@stevefurey there is no progress on this feature, and we will not be able to pick it up in the near future. Maybe someone from the community is interested in picking this up?

ghaliano:
Any progress on this?

I wrote this example in order to hack the clustering future

It's a data grouping based on timeline scale

`var clusteriser = function(timeline){
	this.timeline = timeline;

	this.getDiffDate = function(date1, date2){
		var second=1000;
		var minute = second*60;
		var hour = minute*60;
		var day = hour*24;
		var month = day*31;
[...]

PR #2673, opened 2017-02-03
feat(showcase): added showcase 'Timeline in PowerHA' (#242)

mojoaxel:


mojoaxel:
> Notice this will update all the examples as well

Yes we have to do it at some point. I actually was to lacy to create a seperate PR for this....

Issue #2602, opened 2017-01-19
this.moment.duration is not a function error after upgrading to v4.18.0 


jonathanchrisp:
I've found that when using the latest version 4.18.0, moment throws an error when setting this.moment.duration. Please see below:

```javascript
Uncaught TypeError: this.moment.duration is not a function
    at TimeStep.next (https://cdnjs.cloudflare.com/ajax/libs/vis/4.18.0/vis.js:24492:40)
    at TimeAxis._repaintLabels (https://cdnjs.cloudflare.com/ajax/libs/vis/4.18.0/vis.js:27752:12)
    at TimeAxis.redraw (https://cdnjs.cloudflare.com/ajax/libs/vis/4.18.0/vis.js:27674:10)
    at https://cdnjs.cloudflare.com/ajax/libs/vis/4.18.0/vis.js:21506:27
    at Array.forEach (native)
    at Timeline.Core._redraw (https://cdnjs.cloudflare.com/ajax/libs/vis/4.18.0/vis.js:21505:21)
    at Timeline.<anonymous> (https://cdnjs.cloudflare.com/ajax/libs/vis/4.18.0/vis.js:20741:12)
    at Timeline.Emitter.emit (https://cdnjs.cloudflare.com/ajax/libs/vis/4.18.0/vis.js:11894:22)
    at _onResize (https://cdnjs.cloudflare.com/ajax/libs/vis/4.18.0/vis.js:21697:27)
```

The line in question is:
```javascript
this.current.add(this.moment.duration(this.step, 'hour'));
```

jonathanchrisp:
Just to note we are using webpack on the project and I think this is causing the issue. The code being produced is the following on the `this.moment` instance.

```javascript
this.moment
function (date) {
  return __WEBPACK_IMPORTED_MODULE_5_moment___default()(date).tz(timezone);
}
```

jonathanchrisp:
We've decided to revert to version `4.16.1` for now.

RaekwonStorm:
We had the same error, and were able to fix it by adjusting that line to 
`this.current.add(this.step, 'hour');`

Additionally, I was wondering if there was a particular reason you had to switch blocks nested within an if/else, rather than putting the DST logic entirely in the `hour` case (lines 196-236 of TimeStep.js).

Couldn't you just do the following: 

```
switch (this.scale) {
      case 'millisecond':  this.current.add(this.step, 'millisecond'); break;
      case 'second':       this.current.add(this.step, 'second'); break;
[...]

PR #2683, opened 2017-02-06
bugfix/Timestep.prototype.next this.moment.duration bug

RaekwonStorm:
Addresses issue #2602.

I didn't include it in this PR, but I was also wondering why the many Switch blocks, as each case other than `hour` basically does the same thing.

For example, 

```switch (this.scale) {
      case 'millisecond':
        this.current.add(this.step, 'millisecond');break;
      case 'second':
        this.current.add(this.step, 'second');break;
      case 'minute':
        this.current.add(this.step, 'minute');break;
      case 'hour':
        this.current.add(this.step, 'hour');

        if (this.current.month() < 6) {
          this.current.subtract(this.current.hours() % this.step, 'hour');
        } else {
          if (this.current.hours() % this.step !== 0) {
            this.current.add(this.step - this.current.hours() % this.step, 'hour');
          }
        }
        break;
      case 'weekday': // intentional fall through
      case 'day':
        this.current.add(this.step, 'day');break;
      case 'month':
        this.current.add(this.step, 'month');break;
      case 'year':
        this.current.add(this.step, 'year');break;
      default:
        break;
    }```

could be refactored to something like:

```
if (this.scale === 'weekday') this.scale === 'day';
    
    if (this.scale === 'hour') {
      this.current.add(this.step, 'hour');

      if (this.current.month() < 6) {
        this.current.subtract(this.current.hours() % this.step, 'hour');
      } else {
        if (this.current.hours() % this.step !== 0) {
          this.current.add(this.step - this.current.hours() % this.step, 'hour');
        }
      }
    }  else {
      this.current.add(this.step, this.scale);
    }
```


RaekwonStorm:
@yotamberk Oops, I think I misunderstood your readme on submitting PRs. Let me fix that and I'll resubmit.

PR #2686, opened 2017-02-06
Added showX(YZ)Axis options to Graph3d

nmehrle:
Added options showXAxis, showYAxis, showZAxis to Graph3d. Default to true. Setting one to false disables drawing of respective axis, axis label, tick marks and tick mark labels. Setting all three to false and show grid to false allows display of data with no supplementary lines


nmehrle:
@mojoaxel @yotamberk

yotamberk:
Thanks again @nmehrle !

PR #2691, opened 2017-02-08
Timeline locales : add french

AurelienC:


yotamberk:
Hey @AurelienC , 2 problems:
1. If you want french to be added, add moment with locales. Don't add it to the core.
2. you've pushed to the `master` branch instead of `develop` so this can't be accepted.

Issue #2598, opened 2017-01-18
Timeline - Tooltip flickers when dragging item to edge


thomasbarone:
This is a minor issue that occcurs when dragging an item with a tooltip across to the edge of the timeline so that its start date is before the visible date. It causes the tooltip to flicker.

This can be seen on the vis timeline tooltip example:
[http://visjs.org/examples/timeline/editing/tooltipOnItemChange.html](http://visjs.org/examples/timeline/editing/tooltipOnItemChange.html)

Steps to reproduce:

1. Select an item from the bottom timeline (under "With groups")
2. Drag it to the left so that the group name overlaps the item and drag it back and forth
3. The tooltip will flicker on each move

It would be better if the tooltip remained in place, similar to how the item's text content remains in place in this situation

mojoaxel:
I can reproduce this:
![peek 2017-01-18 12-05](https://cloud.githubusercontent.com/assets/600565/22061693/77b99bb0-dd76-11e6-9d5c-c4fdc99e346a.gif)

It's really a minor issue!

yotamberk:
Fixed with #2702

mojoaxel:
fixed by @yotamberk with #2702

PR #2702, opened 2017-02-11
Flickering onUpdateTimeTooltip 

yotamberk:
This is a simple solution for #2598.
The full solution for the tooltip is probably to append the tooltip to the `dom.body` as a free element and position it compared to the selected updating item instead of appending the tooltip directly to the item. 
But for now, this is a good solution I think.


lib/timeline/component/css/item.css
```diff
@@ -103,6 +103,10 @@
   white-space: nowrap;
   padding: 5px;
   border-radius: 1px;
+  transition: 0.4s;
```
mojoaxel: This is a little dirty but I also see no other way to fix this.

PR #2704, opened 2017-02-11
Item events original event

### LINKED ISSUE #2672: Timeline itemover / itemout do not pass through the original event
Version 4.18.1.

Event callbacks registered for `itemover` / `itemout` do not get the original mouse event as the `event` property on the passed `props` object. Instead they seem to get an object whose value is `{isTrusted: true}`.

This stops us from implementing a more custom hover mechanism using the mouse position to place the hover as we don't have access to the original mouse position.

yotamberk:
fix #2672

The only thing I'm not managing to fix is the `eventListeners.html` example log. I cannot figure out (tried all day) to try logging to the html the `MouseEvent` in the properties. 
If you look at the `properties` in the examples that get passed to the `logEvent` function, you will see the `properties` containing `MouseEvent` with the full data for the event. I couldn't get it to a JSON string format that would be compatible to be displayed on to the html log.


yotamberk:
Fixed

examples/timeline/interaction/eventListeners.html
```diff
@@ -83,11 +86,23 @@
       logEvent(event, properties);
     });
 
+    function elementsCensor (object) {
```
mojoaxel: Maybe this function is better called "stringifyObject" or something similar. "Censor" is realy not the right word here..

yotamberk:
@mojoaxel fixed.

Issue #2647, opened 2017-01-28
[Timeline] Add support for dynamic changing of option rollingMode


seantrace:
Please could you add a feature request to support changing the rollingMode option dynamically using setOptions method.

yotamberk:
added with #2705

mojoaxel:
implemented by @yotamberk with #2705

PR #2705, opened 2017-02-11
Dynamic rolling mode option

yotamberk:
solves #2647


lib/timeline/Timeline.js
```diff
@@ -572,4 +572,20 @@ Timeline.prototype.getEventProperties = function (event) {
...
+/**
+ * Toggle Timeline rolling mode
+ */
+
+Timeline.prototype.toggleRollingMode = function () {
+  if (this.range.rolling) {
+    this.range.stopRolling();
+  } else {
+    if (this.options.rollingMode == undefined) {
+      this.setOptions(options)
```
mojoaxel: `options` is not defined.

PR #2716, opened 2017-02-14
Add initial tests for Timeline PointItem

bradh:
I'm looking to add some fancier item types, this is initial work on understanding what is already there and how it works.

### DIFF
--- a/test/TestSupport.js
+++ b/test/TestSupport.js
@@ -1,0 +1,22 @@
+
+module.exports = {
+  buildSimpleTimelineRangeBody: function () {
+    var body = {
+      dom: {
+        center: {
+          clientWidth: 1000
+        }
+      },
+      domProps: this.props,
+      emitter: {
+        on: function () {},
+        off: function () {},
+        emit: function () {}
+      },
+      hiddenDates: [],
+      util: {}
+    }
+    body.dom.rollingModeBtn = document.createElement('div')
+    return body
+  }
+}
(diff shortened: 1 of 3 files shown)


test/TestSupport.js
```diff
@@ -0,0 +1,22 @@
...
+            domProps: this.props,
+            emitter: {
+                on: function() {},
+                off: function() {},
+                emit: function() {}
+            },
+            hiddenDates: [],
+            util: {}
+            };
+            body.dom.rollingModeBtn = document.createElement('div');
```
yotamberk: Why did you add a rollingModeBtn?

bradh: I didn't (well, not in this PR). That is existing code from https://github.com/almende/vis/blob/develop/test/TimelineRange.test.js#L24 that is needed for the part of the Range constructor that adds an event listener: https://github.com/almende/vis/blob/develop/lib/timeline/Range.js#L69

test/TestSupport.js
```diff
@@ -0,0 +1,22 @@
+
```
mojoaxel: The indent in this file is a little off. Can you please change it to something consistent with the other files. Thanks!

bradh: So just two spaces?

mojoaxel: ok

test/TestSupport.js
yotamberk: But it seems weird that it's needed for the test without being even used

bradh: It is needed to create a Range object, which I use for the "isVisible" checks.

PR #2717, opened 2017-02-14
Documentation: more precision on verticalScroll

CharlesNo:
This addition of information allow people to understand that the verticalScroll works with maxHeight. If maxHeight is not set and verticalScroll is set to true, the verticalScroll will never be shown.


yotamberk:
Thanks! But please resubmit this to the `develop` branch

PR #2729, opened 2017-02-15
Improve the performance of the network layout engine

richvdh:
The performance of the network layout engine is quite poor for highly-connected networks. In the worst case of the fully-connected graph, a number of the methods are O(2^n) in the number of nodes, which is disastrous for large networks.

This PR tweaks the worst culprits to maintain a "visited" map which allows them to only visit each node once, hence limiting them to O(n).


lib/network/modules/LayoutEngine.js
```diff
@@ -532,8 +543,8 @@ class LayoutEngine {
         let diffAbs = Math.abs(pos2 - pos1);
         //console.log("NOW CHEcKING:", node1.id, node2.id, diffAbs);
         if (diffAbs > this.options.hierarchical.nodeSpacing) {
-          let branchNodes1 = {}; branchNodes1[node1.id] = true;
```
richvdh: `getBranchNodes` now requires an empty map rather than one in which the first node is populated; since `getBranchNodes` always populates the map for the first node, this was redundant anyway.

Tooa:
Can we test this somehow? Can you provide us an example showing your improvements for a very dense network?

richvdh:
Have a look at the examples at https://github.com/richvdh/vis/commit/d21fe3b930372fbe7bf0bd7af7bb166ce411f5f9. `demo.jsonp` takes 250ms to lay out here, versus 20 seconds with vis 4.18.1. I've not had the patience to wait for 4.18.1 to complete laying out `demo_big.jsonp`.

Tooa:
Nice! Thank you very much. It looks good to me. @mojoaxel can you have a look at this too? I want to make sure to not overlooking something.

mojoaxel:
@richvdh Thanks very much! 
Please give us some time to review your changes and test them against our existing examples. This is really amazing but we want we make sure not to introduce new bugs (we already have enough :wink: )

mojoaxel:
> Have a look at the examples at richvdh@d21fe3b

@richvdh Can you add these example to the pull request? It's always good to have new examples/tests for "new" features.

richvdh:
>  Can you add these example to the pull request? It's always good to have new examples/tests for "new" features.

I'm happy to just add them as-is if you're happy with that - but they're far from polished, and I'm reluctant to set off down a path of spending more time on that.

mojoaxel:
> I'm happy to just add them as-is if you're happy with that

Yeah that's fine. Just make sure the example is working.

richvdh:
Ok, done.

(don't think the travis failure is my fault!)

mojoaxel:
OK, now we have kind of a problem. 
@ludost also provides with #2759 a big change on the LayoutEngine.
Whatever pull-request gets merged first makes merging for the other one really hard.

If I had to decide, I would say we write some nice performance tests and run both pull-requests against develop. For me this is the only way to decide which pull-request to merge first. 

Of course @ludost is the official maintainer and has the last word on this. In my opinion it would be the best for @ludost to talk directly to @richvdh and figure out how to deal with this performance thing!

_Well, multiple competing pull-requests is a good thing after all!_

ludost:
Hi,

Is suggest merging this one first and let me handle the conflict, as my pull request is still too rough.

Greetings,
Ludo.

PR #2737, opened 2017-02-16
Ensure that setting {editable: true} option works.

bradh:
Includes some bonus tests.

I also considered removing this logic:
```
  this.editable = null;
  if (this.data && this.data.hasOwnProperty('editable')){
    if(typeof this.data.editable === 'boolean') {
        this.editable = {
          updateTime: this.data.editable,
          updateGroup: this.data.editable,
          remove: this.data.editable
        }
    }
    else if(typeof options.editable === 'object') {
      this.editable = {};
      util.selectiveExtend(['updateTime', 'updateGroup', 'remove'], this.editable, data.editable);
    };
  }
```
out of the superclass (https://github.com/almende/vis/blob/master/lib/timeline/component/item/Item.js#L34) but that might result in some unexpected behavioural change depending on how it was being used and for other Item subclasses. So this seemed like a minimally intrusive change.

### DIFF
--- a/lib/timeline/component/item/PointItem.js
+++ b/lib/timeline/component/item/PointItem.js
@@ -101,6 +101,7 @@ PointItem.prototype.redraw = function() {
 
     var editable = (this.options.editable.updateTime || 
                     this.options.editable.updateGroup ||
+                    this.options.editable === true ||
                     this.editable === true) &&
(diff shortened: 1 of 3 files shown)


mojoaxel:
closed in favor of #2743

PR #2757, opened 2017-02-20
ReferenceException in lib/timeline/component/item/Item.js

jangajack:
I received consistent reference exceptions in version **4.18.1** and tracked it back to two lines within _lib/timeline/component/item/Item.js_  Both instances use the same code pattern.

` 
else if(typeof options.editable === 'object') {
      this.editable = {};
      util.selectiveExtend(['updateTime', 'updateGroup', 'remove'], this.editable, data.editable);
};
`

The _options_ object in this context is out of scope and throws a reference error. The first condition of the if statement checks for the boolean form of the editable property on the _this.data_ object and so I wrote the patch assuming the second condition was meant to check the object form of the _this.data_ object.

This is my first GitHub PR, so I hope it's legible and helpful.


yotamberk:
Thanks, but this issue has been addressed in the develop branch

jangajack:
I probably don't have a lot of credibility right now, but after I submitted this I checked-out the **develop branch** to change and build it myself. The instance of this code in the _setData_ function is changed, but the instance in the constructor is still faulty.

Also, both instances should probably use the _this.data.editable_ form for clarity.

**Constructor code:**
```javascript
...
this.editable = null;
  if (this.data && this.data.hasOwnProperty('editable')){
    if(typeof this.data.editable === 'boolean') {
        this.editable = {
          updateTime: this.data.editable,
          updateGroup: this.data.editable,
          remove: this.data.editable
        }
    }
    else if(typeof options.editable === 'object') { // revised code uses this.options.editable
      this.editable = {};
      util.selectiveExtend(['updateTime', 'updateGroup', 'remove'], this.editable, options.editable); // revised with this.data.editable as last parameter
    };
  }
```

jangajack:
Additionally, why check to see if _this.options.editable_ is an object and then not extend that object. I think the patch still has a number of problems, but I'm too new to the code to know the correct solution.

I did a review of all instances where options in referenced and I believe the editable concept has recently gone through a change and the instances of it's use are now inconsistent. See the following excerpts.

This uses the nested data object instead of the parsed local _this.editable_ object
````javascript
Item.prototype._repaintOnItemUpdateTimeTooltip = function (anchor) {
  if (!this.options.tooltipOnItemUpdateTime) return;

  var editable = (this.options.editable.updateTime || 
                  this.data.editable === true) &&
                 this.data.editable !== false;
...
```

This uses the options object submitted to the constructor ignoring the _this.editable_ object. I'm not sure why.
````javascript
Item.prototype._repaintDragCenter = function () {
  if (this.selected && this.options.editable.updateTime && !this.dom.dragCenter) {
    var me = this;

    // create and show drag area
    var dragCenter = document.createElement('div');
...
```

Again, I know I'm new to the code, but I think this entire class needs additional review.

mojoaxel:
@jangajack Before trying to follow your point...You try to merge your `master` into our `develop`. Please base your changes always on the current develop `branch`.
Thanks by the ways for your details PR. This is much appreciated, but you'll need to fix the base.

jangajack:
@mojoaxel Understood, that was my mistake, and I'm happy to implement the changes I outlined above in the develop branch. I believe the editable permission structure needs review throughout the Item.js class and I'm happy to do the work if people agree. Since I sent my review of the develop branch I haven't received a response. The only thing I'm unclear about in this class is instantiation and mocha tests. The Item can be instantiated with null options and a null parent. Although this makes sense in the testing apparatus it seems out of place in the operation of the class. I would like to rewrite it so that Item includes the setData method in its instantiation and that the setData method calculates and extends the editable object from the available parent.options (allow override, etc.), supplied options, previous options, and new options. And maps that to the local this.editable object. This would make the permissions checks within the Item object easier and more predictable. Does that make sense?

mojoaxel:
>  I believe the editable permission structure needs review throughout the Item.js class and I'm happy to do the work if people agree.

Nice! Feel free to improve the existing implementation. This whole "editable" topic was implemented in #2305 as far as I can see.

As always: 
* Try to keep your pull-requests small and easily reviewable. 
* Try to write Tests (we need to improve much here!)
* Make sure the existing examples and still functional.
* Have fun :wink:

PR #2760, opened 2017-02-20
Added support to supply an end to bar charts to have them scale

Vx2gas:


lib/timeline/component/LineGraph.js
```diff
@@ -487,6 +487,7 @@ LineGraph.prototype._updateAllGroupData = function (ids, groupIds) {
       //Copy data (because of unmodifiable DataView input.
       var extended = util.bridgeObject(item);
       extended.x = util.convert(item.x, 'Date');
+      extended.x2 = util.convert(item.x2, 'Date');
```
mojoaxel: I think `x2` is better called `end`. 

We also need documentation for this new parameter `end`. Can you please extend the docs. Thanks!

In addition to that we can think about allowing also `start` as `x` and deprecate `x` in a later version. This would be more compatible with the timeline.

examples/graph2d/21_barsWithEnd.html
```diff
@@ -0,0 +1,51 @@
...
+        {x: '2014-06-13 00:00:00', x2: '2014-06-14 00:00:00', y: 15 },
+        {x: '2014-06-15 00:00:00', x2: '2014-06-16 00:00:00', y: 14 },
+        {x: '2014-06-16 00:00:00', x2: '2014-06-17 00:00:00', y: 17 },
+        {x: '2014-06-17 00:00:00', x2: '2014-06-18 00:00:00', y: 20 },
+        {x: '2014-06-18 00:00:00', x2: '2014-06-25 00:00:00', y: 25 },
+    ];
+
+    var dataset = new vis.DataSet(items);
+    var options = {
+        style:'bar',
```
mojoaxel: What happens with other chart types like line charts? I suspect to `x2` to be ignored!?

Yan you please also update this example to use the new parameter: /examples/graph2d/11_barsSideBySideGroups.html

Vx2gas: Yup.. Its only designed to work with bar charts

lib/timeline/component/LineGraph.js
Vx2gas: Will check in an update to rename x2 to end

yotamberk:
Nice job! THanks!

Issue #1982, opened 2016-07-22
Overlaped items when scrolling left right


RG72:
Overlap items
![visjsoverlap](https://cloud.githubusercontent.com/assets/1561460/17048479/10f852b4-5008-11e6-93ce-6b0cbe14e13b.png)

After zoom in and zoom out. It's OK.
![visjsafterzoominout](https://cloud.githubusercontent.com/assets/1561460/17048493/242609da-5008-11e6-8592-7964c0768f65.png)

Example at http://jsbin.com/cilidonuno/1/edit?html,output

mojoaxel:
The problem seams to occure when "new" items get rendered while moving the timeline (pan).

sk29110:
FYI I only have 2 items in the timeline view still able to reproduce the problem. Just you like mentioned when you scroll(pan) so that 2 items are not in the visible view and scroll back to show those two items they are collapsed into each other.

![2016-11-24_11-16-37](https://cloud.githubusercontent.com/assets/626182/20608554/c46711d2-b237-11e6-8b69-a532b3cf107b.png)

bradh:
I have an idea why this might be occurring. In ItemSet `redraw()`, we mark the `stackDirty` with this logic:
```
  // check whether zoomed (in that case we need to re-stack everything)
  // TODO: would be nicer to get this as a trigger from Range
  var visibleInterval = range.end - range.start;
  var zoomed = (visibleInterval != this.lastVisibleInterval) || (this.props.width != this.props.lastWidth);
  if (zoomed) this.stackDirty = true;
  this.lastVisibleInterval = visibleInterval;
  this.props.lastWidth = this.props.width;

  var restack = this.stackDirty;
```

I tried forcing the restack (set `this.stackDirty` to `true`). I could not reproduce the problem with that in place.

So I think we need to consider if some additional logic is needed here. Perhaps keep old start date (and perhaps end date) and if that has changed then restack.

Fancier logic would check if the change actually needs to restack (e.g. by checking for intersection of start / end dates on each item visible), but that might be more work than is useful.

yotamberk:
@bradh nice investigation work! Do you know to implement the new added logic needed?

bradh:
[...]

Issue #1417, opened 2015-11-05
timeline items with overlapping texts


schotman:
My timeline view on mobile shows overlapping texts when scrolling horizontally. This issues occurs in also in the http://visjs.org/examples/timeline/items/pointItems.html example under specific situation.
Steps to reproduce: 
1. resize the chart such that the view contains items on the same line (screenshot 1)
2. scroll to the left such that the group resizes in height
3. scroll back to the right and see that items are overlapping (screenshot 2)

![screenshot_20151105-160928](https://cloud.githubusercontent.com/assets/2582737/10972226/90d9d004-83d8-11e5-801b-1e8ac390f994.png)
![screenshot_20151105-160946](https://cloud.githubusercontent.com/assets/2582737/10972229/94e91038-83d8-11e5-8da9-d4976057e47b.png)

When pinching the screen redraws and everything is fine again, any idea how to invoke a redraw after a scroll?

josdejong:
Thanks for the detailed description, I can reproduce this bug.

There are event `rangechange` and `rangechanged` fired while moving and zooming, you could use that to force an extra redraw.

bradh:
Still appears to occur with current version (develop branch as of 3456acffa173ae60f06ba8a242a243560a40066a, although that particular commit is nothing to do with this bug, just the testing point).

I couldn't reproduce with the left-right sequence, but it does occasionally occur if you go right (to get the resizing) then left.

roboslone:
This is happening to me while using [react-visjs-timeline](https://github.com/Lighthouse-io/react-visjs-timeline) and custom CSS with `.vis-item .vis-item-overflow {overflow: visible;}`:
<img width="764" alt="screen shot 2017-02-10 at 07 26 16" src="https://cloud.githubusercontent.com/assets/1941275/22814161/4369bd38-ef62-11e6-93a6-2a75ce3e51bb.png">

Is there a way to make timeline consider whole `content` width?

bradh:
Perhaps this is the same issue as https://github.com/almende/vis/issues/1982 ?

yotamberk:
@roboslone The addition of `overflow: visible` is not healthy for the timeline. it is not meant to show the contents. If you want to show contents without overlapping, use point items. The stacking function of items does not consider contents but their containers. What you are requesting is a new feature.

@bradh the original issue is indeed a duplicate

mojoaxel:
> Is there a way to make timeline consider whole content width?

[...]

PR #2774, opened 2017-02-22
[Timeline] Modify redraw logic to treat scroll as needing restack.

bradh:
This addresses #1982 and #1417.

It possibly reduces performance, but correctness seems better. I tried the performance example, and it seems OK (like before the change, albeit on a VM, which might not show full impact).


yotamberk:
I can accept this logic, although It probably should be refactored later to be clearer. Nice job @bradh !

schotman:
This issue is still present on mobile (Android 712, Chrome browser), when overflow: visible. It happens in range view (but also in point view) when scrolling from left to right. Items are added from the left and the text of the item added overlaps the box of the item that is on the timeline. My items have long texts. What happens when you scroll from left to right is that only a small portion of the range is visible, but the complete title, which overlaps. When I move to point view and overflow: hidden, the issue is no longer there.

I cannot reproduce this behavior on desktop, so I suspect it is only on mobile.

Issue #2769, opened 2017-02-22
Feature request: Graph3D Tooltip Styling


p-a:
Currently the style and layout of tooltips are hard coded, and only the inner HTML can be customised.
It would be nice if there was support for some means of styling tooltips, including its line and point.

Tooa:
Thanks for your feature request. It's unlikely that someone in our team will pick this up in near future since no one really works with the `graph3D` module in our team. However, we are open for pull requests. In case you are interested, check out the `develop` branch and send us a pull request. That would be really nice.

p-a:
Created a PR for this issue. Added info to docs and provided a basic example on how to use the new feature.

Issue #626, opened 2015-02-10
Give the pop-up tooltip a class so it is styleable


felixhayashi:
Hi guys,

it would be great if you could assign a class like "network-tooltip" to the tooltip that is shown when hovering a node in the network (setting `node.title`) just like you add e.g. classes to buttons or the frame. This way the end user could style the element, for example giving it a width and `overflow: auto` to make it scroll down if a lot of content is displayed as tooltip. Especially when showing complex html inside the tooltip it would be great if some aspects could be tweaked with css by the user. This adds up well to the existing means to style the tooltip (http://visjs.org/docs/network.html#Tooltips).

Thanks :)

-Felix

josdejong:
I've added a class name `'network-tooltip'` to the tooltip, it's in the `develop` branch.

felixhayashi:
:+1: many thanks @josdejong

wimrijnders:
**Note:** `network-tooltip` only features in the history; there is no occurrence of it anywhere in the code at this time.

wimrijnders:
This issue has been labeled `Inactive`, because there has been no activity here for over two months.

You can:
- Close the issue yourself, so that we know we don't need to take it into account (thanks!).
- Post a new comment to keep it active. The Inactive label and this message will be removed.
- Ignore this comment. If the inactivity continues, the issue will be closed in two weeks.

clayton93:
In case others are looking for an answer to this, the class `.vis-tooltip` is applied to the title.

PR #2780, opened 2017-02-23
Graph3d tooltip styling

p-a:
PR for issue #2769 

Graph3D's option can now take a 'tooltipStyle' object with styling properties for 'content', 'line' and 'dot'.


Tooa:
Thanks for your pull request! The network module uses a class name `network-tooltip` in the tooltip code. See [here](https://github.com/almende/vis/issues/626). Is there any benefit providing the styling as option parameter instead of using a css based approach like in the network module?

p-a:
Well, my first thought was to just replace all hard coded styling altogether with a class, but then I realised that none of the Graph3D examples loads the bundled css file and doing such as thing would break things. Therefore I aimed for a more backwards compatible solution.

But yes, for a new major version a better solution would be to introduce a `graph3d-tooltip` class.
I can of course submit a new PR for such a change.

Any chance that the current PR could be applied for a new minor release since it doesn't break things?

Tooa:
I see - good point. Then this approach is absolute fine for a minor/major version update. Let's keep it as it is and we include it in the next minor release. 

> I can of course submit a new PR for such a change.

I would love to see another pull request, which adds the same feature with the class thing. We are then able to merge this into our `5.0` branch.

lib/graph3d/Settings.js
```diff
@@ -250,6 +250,15 @@ function setOptions(options, dst) {
   setSpecialSettings(options, dst);
 }
 
+/**
+ * Merge fields from src to dst
+ */
+function mergeFields(src, dst, fields) {
```
Tooa: I think something like this already exists in the utils [here](https://github.com/p-a/vis/blob/d4a8e878a36f26ea539a2bd15060eecb542a5bfa/lib/util.js#L198).

examples/graph3d/13_tooltip_styling.html
```diff
@@ -0,0 +1,132 @@
...
+      // create our graph
+      var container = document.getElementById('mygraph');
+      graph = new vis.Graph3d(container, data, options);
+
+      if (camera) graph.setCameraPosition(camera); // restore camera position
+
+      document.getElementById('style').onchange = drawVisualization;
+    }
+  </script>
+  <script src="../googleAnalytics.js"></script>
```
Tooa: As far as I know, we don't want `googleAnalytics` anymore ([reference](https://github.com/almende/vis/pull/2670)). Can you remove it?

examples/graph3d/13_tooltip_styling.html
```diff
@@ -0,0 +1,132 @@
...
+        width:  '600px',
+        height: '600px',
+        style: style,
+        showPerspective: true,
+        showLegend: true,
+        showGrid: true,
+        showShadow: false,
+
+        // Option tooltip can be true, false, or a function returning a string with HTML contents
+        //tooltip: true,
```
Tooa: Remove `tooltip: true,`

examples/graph3d/13_tooltip_styling.html
```diff
@@ -0,0 +1,132 @@
+<!doctype html>
```
Tooa: This example is basically just a copy of [this](https://github.com/almende/vis/blob/develop/examples/graph3d/11_tooltips.html) one. Maybe it is better to adapt the other example for such a small change. What do you think?

docs/graph3d/index.html
```diff
@@ -499,6 +499,35 @@ <h2 id="Configuration_Options">Configuration Options</h2>
     </tr>
 
     <tr>
+      <td>tooltipStyle</td>
+      <td>Object</td>
+      <td>{<br>
+        content&nbsp;:<br>{
```
Tooa: Do we really need No-break spaces here?

mojoaxel: @p-a Better use a <pre> block here.

examples/graph3d/13_tooltip_styling.html
mojoaxel: I think so too. Feel free to adopt the existing example to your needs.

examples/graph3d/13_tooltip_styling.html
mojoaxel: Yes please remove the analytics.

lib/graph3d/Graph3d.js
```diff
@@ -67,6 +67,29 @@ var DEFAULTS = {
 
   style            : Graph3d.STYLE.DOT,
   tooltip          : false,
+
+  tooltipStyle     : {
```
mojoaxel: I haven't tested it yet. Please make sure the existing tooltips are basically not changed, or we have to consider this as a breaking change.

examples/graph3d/13_tooltip_styling.html
p-a: ok. removed "my" example and altered the original a bit

lib/graph3d/Settings.js
p-a: Done!. Used the util's methods instead

examples/graph3d/13_tooltip_styling.html
p-a: done!

examples/graph3d/13_tooltip_styling.html
p-a: Sorry, this was probably due to me branching of master initially...  But this file is now remoed altogether.

docs/graph3d/index.html
p-a: Done!

lib/graph3d/Graph3d.js
p-a: Tested all combos, if no `tooltipStyle` is provided the tooltips are not changed and are exactly as before.

p-a: @mojoaxel changes ok?

PR #2799, opened 2017-02-27
Add edited data as argument to the template function

RealRegatta:
Allow to display edited data with the template function.
See the sample [here](https://goo.gl/Hp33Kz)  (in browser tab enter : http://localhost:17305/index.html)


mojoaxel:
I was too fast. We also should extend the documentation!
@RealRegatta Can you please create a new pull request with an extension of the docs.Thx!

RealRegatta:
#2802

PR #2802, opened 2017-02-27
Add edited data as argument to the template function with update documentation

RealRegatta:


bradh:
Does this break existing users of the template functions (e.g. in Typescript or ES6)?

RealRegatta:
I don't know TypeScript and can not tell if add an argument to a function is a breaking change

bradh:
The code is merged, so we should do the docs.

Issue #2536, opened 2017-01-03
3d bar graph data array unsorted


hawkeyeaz1:
If you pass Graph3d an array (for data) and the array is not sorted the graphed result breaks visual usability.

Ex: graph = new vis.Graph3d(container, [ {x: 1, y: 3, z: 0}, {x: 0, y: 0, z: 0} ], { style: 'bar' });

Tooa:
This looks like a bug to me. An example to produce can be found [here](http://jsbin.com/moyazaluqe/1/edit?html,output).I'm not really used to the `Graph3D` component. So, how does this order should be defined? Is it enough to sort by the first coordinate? Can you show some more examples? Maybe this one is special?

hawkeyeaz1:
graph = new vis.Graph3d(container, [ {x: 1, y: 1, z: 0}, {x: 0, y: 0, z: 0} ], { style: 'bar' });
graph = new vis.Graph3d(container, [ {x: 1, y: 1, z: 1}, {x: 0, y: 0, z: 0} ], { style: 'bar' });
graph = new vis.Graph3d(container, [ {x: 0, y: 1, z: 1}, {x: 0, y: 0, z: 0} ], { style: 'bar' });
graph = new vis.Graph3d(container, [ {x: 0, y: 0, z: 1}, {x: 0, y: 0, z: 0} ], { style: 'bar' }); // Really interesting

wimrijnders:
Confirmed. The initialization code of graph3d assumes ordering of data. So in the first example, you get a bar width of -1 for the x-coordinate.

There is another hidden assumption here, namely that all bars have the same width. If the bar dimensions are initialized from data, this assumption may explode.

Preparing a PR to tackle them both.

wimrijnders:
To circumvent this with the current version, you can set the default bar widths in the options hash:

    var options = {
      // ...
      xBarWidth: 1,
      yBarWidth: 1,
    };
    // Do further magic here
    
    // ... and pass in the bar defaults in the options:
    graph = new vis.Graph3D(container, data, options);

wimrijnders:
I've added PR #2803 for this.

Tooa:
Thank you @wimrijnders for the fix and welcome back :)

PR #2803, opened 2017-02-28
Fix for issue #2536

wimrijnders:
Fix for initialization of bar graphs from data. When determining bar widths:

* Unsorted data now dealt with properly
* Data not assumed to be evenly spaced any more

### DIFF
--- a/lib/graph3d/Graph3d.js
+++ b/lib/graph3d/Graph3d.js
@@ -331,21 +331,30 @@ Graph3d.prototype.getDistinctValues = function(data, column) {
 }
 
 
+/**
+ * Determine the smallest difference between the values for given
+ * column in the passed data set.
+ *
+ * @returns {Number|null} Smallest difference value or
+ *                        null, if it can't be determined.
+ */
 Graph3d.prototype.getSmallestDifference = function(data, column) {
   var values = this.getDistinctValues(data, column);
   var diffs  = [];
 
   // Get all the distinct diffs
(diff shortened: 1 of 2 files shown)


lib/graph3d/Graph3d.js
```diff
@@ -327,7 +327,25 @@ Graph3d.prototype.getDistinctValues = function(data, column) {
       distinctValues.push(data[i][column]);
     }
   }
-  return distinctValues;
+  return distinctValues.sort(function(a,b) { return a - b; });
+}
+
+
+Graph3d.prototype.getSmallestDifference = function(data, column) {
```
Tooa: Open for discussion: 

Is space complexity a matter here? I assume `data` isn't usually that large right? Otherwise, I might prefer something like:

`Sort data`
`Calc minDiff <- data[1] - data[0]`
`for (i <- 2 to len(a)) do minDiff <- min(minDiff, data[i] - data[i-i])`

Also maybe mention in a comment that when you pass an empty `data` value the function returns `undefined`.

wimrijnders: OK fair enough. Will adjust the PR.

wimrijnders: > Also maybe mention in a comment that when you pass an empty data value the function returns undefined.

Also for a single point graph. I went with `null` as return value instead of `undefined`.

Comment added.

Issue #2810, opened 2017-03-01
Timeline - Nested groups do not use `groupOrder`. 


frankisans:
`ItemSet.prototype._orderGroups` use `grupOrder` option to sort all the groups (top groups and nested groups) but then `_orderNestedGroups` get the nested groups without applying order.

Maybe its enough to change:

```
       var nestedGroups = this.groupsData.get({
          filter: function filter(nestedGroup) {
            return nestedGroup.nestedInGroup == groupId;
          }
        });
```

to force the order again when geting nested groups:

```
       var nestedGroups = this.groupsData.get({
          filter: function filter(nestedGroup) {
            return nestedGroup.nestedInGroup == groupId;
          },
          order: this.optios.groupOrder
        });
```

We can even define a specific option `nestedGroupOrder` for sorting nested groups and use `groupOrder` if it is not defined.

In terms of performance, the first ordering of `_orderGroups` could be optimized to not include nested groups.

eanugent:
Thanks for posting this issue. I'm not able to get my nested groups to sort in any order other than by the id

brichbe:
Relates to #2676

PR #2817, opened 2017-03-03
Fix to sort nested groups - Related to #2810

frankisans:


Tooa:
Can you please add some words about what your patch fixes?

Issue #2594, opened 2017-01-17
Timeline - Uncaught TypeError: Cannot read property 'height' of undefined


jtissier:
Hi all,

First I want to say how much I love vis.js library, it saved me months of time to developp a timeline like yours.

I got an error Error on visjs 4.16.1 and 4.18.0 when i try to move a subgrouped item in another group :

```
vis.js:21715 Uncaught TypeError: Cannot read property 'height' of undefined
    at http://localhost:63342/rms/bower_components/vis/dist/vis.js:21715:94
    at Object.exports.forEach (http://localhost:63342/rms/bower_components/vis/dist/vis.js:718:9)
    at Group._calculateSubGroupHeights (http://localhost:63342/rms/bower_components/vis/dist/vis.js:21713:12)
    at Group.redraw (http://localhost:63342/rms/bower_components/vis/dist/vis.js:21635:10)
    at http://localhost:63342/rms/bower_components/vis/dist/vis.js:19374:32
    at Object.exports.forEach (http://localhost:63342/rms/bower_components/vis/dist/vis.js:724:11)
    at ItemSet.redraw (http://localhost:63342/rms/bower_components/vis/dist/vis.js:19372:10)
    at http://localhost:63342/rms/bower_components/vis/dist/vis.js:18495:27
    at Array.forEach (native)
    at Timeline.Core._redraw (http://localhost:63342/rms/bower_components/vis/dist/vis.js:18494:21)
```
If I catch this error I got another one on line 22179 and when I catch the second (in a console.warn) all looks good.

How can i fix it ?

blin4444:
I'm receiving similar errors after upgrading from 4.17.0 to 4.18.0. Needed to downgrade back to 4.17.0.

yotamberk:
@jtissier @blin4444  can you please supply a jsbin with an example with this error popping up?

mojoaxel:
Looking at [the subgroups example](http://visjs.org/examples/timeline/groups/subgroups.html) I can not reproduce this, by tracking any item to another group.
@@jtissier @blin4444 Please provide a way for us to reproduce this. Thx!

jtissier:
I will try to provide you a jsbin soon

blin4444:
It will be a bit difficult for me to do because I am using Vis commercially.

yotamberk:
@jtissier @blin4444 any progress? I want to help but I can't reproduce it. I need your examples

blin4444:
[...]

PR #2821, opened 2017-03-03
Fixes for removing and adding items to subgroups

BTMorton:
Added methods to remove and add items to a subgroup.
Added a method to change the subgroup an item is in, rather than removing it from the group entirely and then re-adding.
No longer decrement subgroup ID when removing a subgroup, as this can cause collisions if you remove the last item from a group and then add another one to it.
This should fix the issues found in #2594


bradh:
Seems like a good candidate for some unit tests.

bradh:
@BTMorton do you need a hand with the tests?

BTMorton:
Oh, sorry, I didn't realise you actually had a test framework yet, or that you wanted me to do them :P I can take a look at it, but if you wanted to suggest some test cases, that wouldn't be unhelpful :D

yotamberk:
@BTMorton What's the status of the tests? 
@bradh I think we should accept this PR and get the tests in in a separate PR. What do you think?

BTMorton:
Sorry, I haven't had the chance yet. As you can probably imagine, the fix is more important to the company than writing the tests...

yotamberk:
Merging this PR. Tests will be added in future PRs

Issue #2795, opened 2017-02-26
Custom format function receives Date instead of Moment


marcortw:
A custom minorLabels and a custom majorLabels function don't seem to receive the same object type (expected: Moment for both, but majorLabels gives back a Date).
There seems to be a minor bug in how the custom format function works, see the fiddle on https://jsfiddle.net/bkfj4L3h/

Cheers
Marco

yotamberk:
Closed with #2826

PR #2826, opened 2017-03-04
Cheap fix for bug #2795

marcortw:
This fix ensures that the getLabelMinor and getLabelMajor always returns a Moment object. Especially useful if you use a custom function for the timeline labels (https://github.com/almende/vis/issues/2795).

### DIFF
--- a/lib/timeline/TimeStep.js
+++ b/lib/timeline/TimeStep.js
@@ -521,7 +521,9 @@ TimeStep.prototype.getLabelMinor = function(date) {
   if (date == undefined) {
     date = this.current;
   }
-  if (date instanceof Date) date = this.moment(date);
+  if (date instanceof Date) {
+    date = this.moment(date)
+  }
 
   if (typeof(this.format.minorLabels) === "function") {
     return this.format.minorLabels(date, this.scale, this.step);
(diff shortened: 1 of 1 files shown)


lib/timeline/TimeStep.js
```diff
@@ -521,6 +521,7 @@ TimeStep.prototype.getLabelMinor = function(date) {
   if (date == undefined) {
     date = this.current;
   }
+  if (date instanceof Date) date = this.moment(date);
```
yotamberk: For consistency of code, I think it should be written in a newline format:
``` javascript
if (date instanceof Date) {
  date = this.moment(date)
}
```

yotamberk: For consistency of style in code, please change this to a newline format:
``` javascript 
if () {
...
}
```

lib/timeline/TimeStep.js
```diff
@@ -540,6 +541,7 @@ TimeStep.prototype.getLabelMajor = function(date) {
...
```
yotamberk: Same comment as above

marcortw:
See my updated PR - I hope i'm doing this the right way.

Issue #2835, opened 2017-03-06
Timeline vertical visibility performance optimization is broken


thomasbarone:
The performance optimization where groups/items that are not visible in the timeline window are not loaded to the DOM is currently broken.

See issue #1383 for a discussion of the original feature request, I feel it is best summarised in this sketch:

```
     |                   |
     | "invisible" items | 
     |                   |
 ------------------------------
 |                            |
 |       visible area         | 
 |                            |
 ------------------------------
     |                   |
     | "invisible" items | 
     |                   |
```

The problem can be seen in this [example](http://visjs.org/examples/timeline/groups/verticalItemsHide.html). Zoom in and click _Show current visible items_. It will show the visible items as well as all the "invisible" items (in groups that are above / below the current window).

yotamberk:
The performance  doesn't have an issue. It's a problem of the method `getVisibleItems`

mojoaxel:
This tests from @nathanbennett look like there is room for improvement here:

![fixcomparison](https://cloud.githubusercontent.com/assets/11827402/23735638/ba3d7434-04d6-11e7-8ab0-a99819058356.png)

simpleclicks:
Hi is there a work around for this till this bug is fixed?

PR #2848, opened 2017-03-09
Improve redraw performance with non-visible groups (related to #2835)

nathanbennett:
Related to issue #2835. I found that having a large number of non-visible groups was causing performance issues as the items for these groups are being constantly added then removed from the DOM (so they can be restacked).

The fix I made was to only restack visible groups, and to always restack a group when it becomes visible (as it may have missed out previously). Tested these changes with Chrome DevTools and it reduces a lot of "forced reflow" hammering when dragging a timeline item.

fixcomparison

I modified an existing example to demo this (https://jsfiddle.net/nathanbennett/cbprq1vt/). This isn't a very intensive case, so you might need to use CPU throttling to see an appreciable difference :).


yotamberk:
Another note: 
The verticalScroll is kind of glitchy too...

nathanbennett:
Thanks @yotamberk . I thought it was too simple :).

Unfortunately, there doesn't seem to be a way to prevent the vertical scroll jumping without restacking non-visible groups. I had another look at the issue, and decided the problem might not be "non-visible groups are restacking" but that "groups which haven't been changed are restacking".

All groups were being forced to restack when any item in the ItemSet changed (as this was tracked in ItemSet.stackDirty). I've had a go at changing this to be tracked at a group level, ie. Group.stackDirty, so that only the necessary groups are restacked.

The assumptions I've had to make are:
- The parent of an item is always a Group (seems true by current code)
- Groups only need restacking when they first appear, when an item is added/removed, when Item.setData() is called, or when the timeline is zoomed/scrolled. (I based this on the way ItemSet.stackDirty was handled)


This isn't as strong an improvement (zooming and scrolling will be as slow as before), but is still as effective for the case where an item is being dragged around.

yotamberk:
@nathanbennett Can you supply a new jsbin to show the improvements?

nathanbennett:
@yotamberk No worries. See [here](https://jsfiddle.net/nathanbennett/y8tdfoz0/).

yotamberk:
Thanks! Looks good to me

SupineSnail:
Thank you for this nathan!  I merged this in locally with develop and this fixes most of my performance issues I was having with having hundreds of groups!  Will be looking forward to this being merged in to the main branch.

yotamberk:
@nathanbennett can you pull from the develop branch and push again? It won;t let me merge this PR for some odd reason

lib/timeline/component/BackgroundGroup.js
```diff
@@ -22,10 +22,10 @@ BackgroundGroup.prototype = Object.create(Group.prototype);
  * Repaint this group
  * @param {{start: number, end: number}} range
  * @param {{item: {horizontal: number, vertical: number}, axis: number}} margin
- * @param {boolean} [restack=false]  Force restacking of all items
+ * @param {boolean} [forceRestack=false]  Force restacking of all items
  * @return {boolean} Returns true if the group is resized
  */
-BackgroundGroup.prototype.redraw = function(range, margin, restack) {
+BackgroundGroup.prototype.redraw = function(range, margin, forceRestack) {
```
yotamberk: Why is `forceRestack` defined as a parameter but not used?

yotamberk: Why is this parameter defined but never used?

nathanbennett: To me it looked like the idea was to keep its signature consistent with Group.prototype.redraw, so I left this as it was (other than renaming the parameter in line with my changes).

qwe852147:
Hi, what time the new version will be released which fix redraw performance?

yotamberk:
@qwe852147 We are having trouble merging this PR. I've asked the submitter @nathanbennett to resubmit this, but yet have had a response

yotamberk:
Closed for #3078 (due to merging problems)

PR #2849, opened 2017-03-09
Configurable minimum and maximum sizes for dot-size graphs

wimrijnders:
For graph type `dot-size` in `graph3d`, the difference between the smallest dot size and the largest is a factor of 5. I would like to have minimum and maximum dot sizes to be configurable for this graph type.

As an example, in the following diagram the values for the dots are, left to right, 1, 10 and 100. Due to the current scaling limitations, the proportions of the dot radii do not reflect this.

dotsize

I would appreciate an option in `graph3d` for extending the range of the minimum and maximum radius; this PR is an implementation of this option.

I have added two new settings:
*  `dotSizeMinFraction` - size of min-value dot as a fraction of dotSizeRatio; default `0.5`
*  `dotSizeMaxFraction` - size of max-value dot as a fraction of dotSizeRatio; default `2.5`

The default values correspond to the values in the current version, so these settings are backward-compatible. I have also added them to code of the example page `graph3d/08_dot_cloud_size.html`. I propose that you use this example to try out the settings.


wimrijnders:
For comparison, this is what the previous example looks like with use of the new fields:

dotsize2

Getting the proportions right definitely has its uses; however, I'm not too happy with how this is exposed in the interface. Perhaps someone has a better suggestion.

mojoaxel:
Is this a breaking change? It looks to me just like an improvement. I think we can release this with 4.20 !?

Tooa:
@mojoaxel No breaking change. However, I asked @wimrijnders about this via `gitter` and he replayed that we can (should?) delay it to `v5`. @wimrijnders Do you want to have this merged into `v4.20`?

wimrijnders:
No, not a breaking change. I advised @Tooa to skip it for the upcoming release because it's new functionality, i.e. it doesn't solve any current problems but adds something new.

Was this a wrong thought? You are welcome to add it as long as it's properly reviewed. But I honestly had the impression that you guys are busy enough already without having to give priority to these personal gripes of mine.

PR #2852, opened 2017-03-10
Small fixes for nested groups

BTMorton:
These are fixes implemented by our team to get the desired, dynamic nested group functionality:
* Allow nested groups to be removed by nulling the nestedGroups content.
* Don't update the showNested property if nothing has been set in the data object, unless it is undefined.
* Verify group object exists during `onGroupClick`


Tooa:
Thanks for your pull request :) Does this tackle any open issue in our tracker?

BTMorton:
Apologies, used to merging against master. Should be OK now.

It's not against any open issues that I've seen, it's just a set of fixes that were required for our particular use case. I can open issues with examples of the things this fixes, if it helps?

bradh:
Seems like another change that would benefit from unit tests

mojoaxel:
> I can open issues with examples of the things this fixes, if it helps?

@BTMorton Thanks, that would be helpful.

BTMorton:
Next included issue: When a group with nested groups is added and updated in the dataset without a showNested property, and the nested groups are currently hidden, triggering the update puts the parent group into un-collapsed mode whilst still not showing the nested groups.

Example can be seen at https://plnkr.co/edit/uobyLYuTGv9zcljkpt0a?p=preview. To reproduce, hide the nested group and click trigger update.

yotamberk:
Thanks @BTMorton ! Looking forward to more PRs from you!

Issue #2805, opened 2017-02-28
Callback for animations


henev:
Hello,

Is it possible to integrate callbacks for functions like `.setWindow()` that have animations? I would like to execute code after the animation ends. Is `setTimeout` the only possible way for now?

Thank you in advice :)

bradh:
Could you just use the rangeChanged event? See http://visjs.org/docs/timeline/#Events

henev:
Yes, I am concern with timeline.

`rangeChange` is also possible but for some things I think it will be better to use a callback which will be executed once.

bradh:
There are two events - `rangeChange` (Fired repeatedly when the timeline window is being changed) and `rangeChanged` (Fired once after the timeline window has been changed).

If you don't want `rangeChanged` behaviour, can you describe how what you want is different from that event?

henev:
I have an event handler that is attached on the `rangeChanged` event. I don't want to use this handler and to do checks there in order to execute code that is relevant only for the context I am calling `.setWindow` from. It will be much easier if I have a callback for that purpose.

bradh:
How would the callback you want differ from `rangeChanged`? I'm not trying to be difficult, I'm trying to understand the requirement.

henev:
If I call `.setWindow` from somewhere and I want after the animation ends to `.focus()` or to `.setSelection` with an item id, how would you achieve this?

bradh:
Can you set up a jsfiddle / jsbin that shows your situation?

yotamberk:
@henev checkout this PR: #2870

mojoaxel:
@henev @yotamberk Can this be closed?

PR #2870, opened 2017-03-18
Add callback functions to moveTo, zoomIn, zoomOut and setWindow

yotamberk:
checkout #2805.
I've realized that @henev is right and there should be callback functions for timeline methods using animations. So I added this PR to have the ability to call a callback function right after the `setRange` finishes (after the animation is done). This is in addition to `rangechange` event that gets called after these methods.


lib/timeline/Core.js
```diff
@@ -657,17 +658,28 @@ Core.prototype.getDataRange = function() {
...
+  else if (arguments.length == 2 && typeof arguments[1] == "function") {
+    var range = arguments[0];
+    callback = arguments[1];
+    animation = (range.animation !== undefined) ? range.animation : true;
+    this.range.setRange(range.start, range.end, { animation: animation }, callback);
   }
   else {
     animation = (options && options.animation !== undefined) ? options.animation : true;
-    this.range.setRange(start, end, animation);
+    this.range.setRange(start, end, { animation: animation }, callback);
```
bradh: This optional arguments / callback handling is starting to get pretty messy. Not sure of a better way  - perhaps treating the callback as a option?

yotamberk: Short answer: No.
I really think the callback function should be separated from the options. 
It's not exactly an option and should be passed through in an object.

PR #2907, opened 2017-03-26
[Timeline fix #2814] Do not corrupt class names at high zoom levels.

### LINKED ISSUE #2814: Css mangled at high zoom levels on timebar
The odd and day classes appear to get mangled together:  
" vis-yesterdayvis-odd"  

I beleive this is because of the trims on the functions:   https://github.com/almende/vis/blob/master/lib/timeline/TimeStep.js#L616  

As when running it in dev console in chrome:  
today(current) + even(current.minutes()).trim() gives me " vis-yesterdayvis-odd"  
but:  
today(current) + even(current.minutes()) gives me " vis-yesterday vis-odd"  
which would appear to be correct to me

I am happy to put in a pull request for this but do not know enough about the library to tell if just removing the trim function is what is needed.

bradh:
This takes a more robust approach to produce the list of class names.

I tried to produce tests for the CSS, but the concept of "now" throughout the code defeated me. So just fixes the bug for now.

### DIFF
--- a/lib/timeline/TimeStep.js
+++ b/lib/timeline/TimeStep.js
@@ -560,6 +560,7 @@ TimeStep.prototype.getClassName = function() {
   var m = this.moment(this.current);
   var current = m.locale ? m.locale('en') : m.lang('en'); // old versions of moment have .lang() function
   var step = this.step;
+  var classNames = [];
 
   function even(value) {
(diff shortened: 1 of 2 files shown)


bradh:
Replaced by PR #2909

PR #2937, opened 2017-03-30
Fix #2936: 0 nodeId not shown in getConnectedNodes(edgeId) in EdgeHalnder.js

### LINKED ISSUE #2936: [network] getConnectedNodes(edgeId) doesn't return nodeId with 0 value
Suppose we have a network containing a node with id equals 0.
Let edgeId be an id of connected to the node edge.
Executing network.getConnected(edgeId) returns list with 1 value instead of 2, consumes 0 nodeId.

Code example: 
http://jsbin.com/voyaludiwo/1/edit?html,js,console

Screenshots:
image
console

KarKarbI4:
fixes #2936


bradh:
The pull needs to go into develop, not into master.

Issue #2940, opened 2017-04-01
minor (probably trivial) issue in gephiParser.js


rretter:
Lines 48-52 are:
```
    node['title'] = gNode.title;
    node['x'] = gNode.x;
    node['y'] = gNode.y;
    node['label'] = gNode.label;
    node['title'] = gNode.attributes !== undefined ? gNode.attributes.title : undefined;
```
Note that the last line undoes the first one.  This is bad if the first line has meaning, or unimportant if the first line doesn't matter (but probably ought to be deleted just for neatness).

wimrijnders:
I can see the issue, but I'm having difficulty resolving it:

* I can't run the `importingFromGephi` example; applying the googled solutions doesn't help me
* I can't find specifications on any `Gephi JSON` format

So I'm afraid I'm not much use here.
Having said that, I can confirm from earlier issue #1189 that the `gNode.attributes.title` path is valid. I'm in the dark about `gNode.title`.

I propose to use `gNode.title` as the fallback option in the second assignment:

```
    node['x'] = gNode.x;
    node['y'] = gNode.y;
    node['label'] = gNode.label;
    node['title'] = gNode.attributes !== undefined ? gNode.attributes.title : gNode.title;
```

If it doesn't exist, no harm done. Does this sound OK to you?

rretter:
This sounds like a good approach.  Thanks!

wimrijnders:
On closer inspection, this has already been fixed in the `develop` branch. The first assignment has been removed already. No need to make a PR for the change as above.

wimrijnders:
Previous PR was faulty, redoing.

PR #2961, opened 2017-04-10
Fix for #2940 - Gephi combine assignments of node title

wimrijnders:


wimrijnders:
It appears that someone already fixed this. No need for this PR.

PR #2964, opened 2017-04-10
Fixes for loading images into image nodes

wimrijnders:
Fixes for issues which became apparent due to #2894.

This affects the loading of images for node types `Image` and `CircularImage`.

- Added extra checks, warnings and errors on image loading
- Prevent double calls of `setOptions()` on initialization of nodes with images
- Cleaned up usage of `switchImages()` in classes `Image` and `CircleImage`
- Added commenting where necessary

### DIFF
--- a/lib/network/Images.js
+++ b/lib/network/Images.js
@@ -31,8 +31,12 @@ class Images{
      * @return {Image} imageToLoadBrokenUrlOn   The image object
      */    
     _tryloadBrokenUrl (url, brokenUrl, imageToLoadBrokenUrlOn) {
-        //If any of the parameters aren't specified then exit the function because nothing constructive can be done
-        if (url === undefined || brokenUrl === undefined || imageToLoadBrokenUrlOn === undefined)  return;
+        //If these parameters aren't specified then exit the function because nothing constructive can be done
+        if (url === undefined || imageToLoadBrokenUrlOn === undefined)  return;
+        if (brokenUrl === undefined) {
+          console.warn("No broken url image defined");
+          return;
+        }
     
         //Clear the old subscription to the error event and put a new in place that only handle errors in loading the brokenImageUrl
         imageToLoadBrokenUrlOn.onerror = () => {
--- a/lib/network/Images.js
+++ b/lib/network/Images.js
@@ -1,29 +1,17 @@
+import CachedImage from './CachedImage';
+
+
 /**
  * @class Images
  * This class loads images and keeps them stored.
  */
-class Images{
+class Images {
     constructor(callback){
         this.images = {};
         this.imageBroken = {};
         this.callback = callback;
     }
     
-    /**
-     * @param {string} url                      The Url to cache the image as 
-      * @return {Image} imageToLoadBrokenUrlOn  The image object
-     */    
-    _addImageToCache (url, imageToCache) {
-        // IE11 fix -- thanks dponch!
-        if (imageToCache.width === 0) {
-            document.body.appendChild(imageToCache);
-            imageToCache.width = imageToCache.offsetWidth;
-            imageToCache.height = imageToCache.offsetHeight;
-            document.body.removeChild(imageToCache);
-        }
-    
-        this.images[url] = imageToCache;
-    }  
     
     /**
      * @param {string} url                      The original Url that failed to load, if the broken image is successfully loaded it will be added to the cache using this Url as the key so that subsequent requests for this Url will return the broken image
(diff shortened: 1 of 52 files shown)


lib/network/modules/components/nodes/util/CircleImageBase.js
```diff
@@ -24,17 +58,21 @@ class CircleImageBase extends NodeBase {
...
 
   /**
-   * Switch between the base and the selected image.
+   * Set selection and switch between the base and the selected image.
+   *
+   * Do the switch only if imageObjAlt exists.
+   *
+   * @param {true|false} selected value of new selected state for current node
    */
   switchImages(selected) {
```
rasmusblockzero: Maybe this should now be changed to choseImage or determineImage as its called all the time?

Or, as the this.selected is set here perhaps  handleImageForSelectStatus

lib/network/modules/components/nodes/util/CircleImageBase.js
```diff
@@ -10,10 +10,44 @@ class CircleImageBase extends NodeBase {
...
+   * therefore, this method needs to be reentrant. 
+   *
+   * For correct working in error cases, it is necessary to properly set
+   * field 'nodes.brokenImage' in the options.
+   *
+   * @param {Image} imageObj  required; main image to show for this node
+   * @param {Image|undefined} optional; image to show when node is selected
+   */
   setImages(imageObj, imageObjAlt) {
+    if (imageObj === undefined) {
```
rasmusblockzero: This wouldn't mean anything to the user - but should never occur because of user Error, correct?

Is there a standard in Visjs to separate things that users might see as 
throw "Option image must be defined for node type '" + this.options.shape + "'";
and pure internal debugging errors?

lib/network/modules/components/nodes/util/CircleImageBase.js
```diff
@@ -10,10 +10,44 @@ class CircleImageBase extends NodeBase {
...
+    //
+    // As an extra complication, this setOptions() is also called
+    // on an update to the node, where the two images are passed
+    // for Image and CircleImage (only).
+    //
+    // In the first case, imageObj and ImageObjAlt are always undefined;
+    // we use this to differentiate the calls.
+    // TODO: this is hacky, find a better solution
+    //
+    if (!(imageObj === undefined && imageObjAlt === undefined)) {
```
rasmusblockzero: Are there any risks that this blocks updates if the image url is reset later on?

lib/network/Images.js
```diff
@@ -31,8 +31,12 @@ class Images{
      * @return {Image} imageToLoadBrokenUrlOn   The image object
      */    
     _tryloadBrokenUrl (url, brokenUrl, imageToLoadBrokenUrlOn) {
-        //If any of the parameters aren't specified then exit the function because nothing constructive can be done
-        if (url === undefined || brokenUrl === undefined || imageToLoadBrokenUrlOn === undefined)  return;
+        //If these parameters aren't specified then exit the function because nothing constructive can be done
+        if (url === undefined || imageToLoadBrokenUrlOn === undefined)  return;
```
rasmusblockzero: Is there any path that would lead to a undefined url here? If the check is for debugging of a case that should never happen then maybe an internal error print/throw make sense ?

wimrijnders: Yes there is apath:

* `image: true` specified in node(s)
* Neither image nor selected image set
* And no 'broken url' image set

lib/network/modules/components/Node.js
```diff
@@ -168,6 +154,46 @@ class Node {
...
+    if (this.options.shape !== 'circularImage' && this.options.shape !== 'image') {
+      return;
+    }
+
+    if (this.options.image === undefined) {
+      throw "Option image must be defined for node type '" + this.options.shape + "'";
+    }
+
+    if (this.imagelist === undefined) {
+      throw "No imagelist provided";
```
wimrijnders: This is a fair point. `this.images` comes from `Network.images`, which is an internal structure.

All the user needs to know at this point is that no images at all were defined. So, this should be sufficient:

```
  throw "No images provided";
```

I think this covers it; I'm open to other suggestions.

lib/network/modules/components/nodes/util/CircleImageBase.js
wimrijnders: Yes, plenty of risks. All you need to do is call `setOptions()` with a new image struct.

Thankfully, accepted PR #3053 addresses this issue, so all is well now. I'll just delete those comments now.

lib/network/modules/components/nodes/util/CircleImageBase.js
wimrijnders: It *can* occur as user error, but the user will be notified through a check elsewhere. See response to your first review comment. In any case, `setImages()` is known to be broken and #3053 fixes it.

> Is there a standard in Visjs to separate things that users might see as...

Not that I know of, actually. Perhaps it's a good idea to differentiate the different kinds of errors. Something like prefix 'Internal error:' perhaps?

lib/network/modules/components/nodes/util/CircleImageBase.js
wimrijnders: For better naming: what is actually happening here? If the node is selected, the selected-image is set to be displayed instead of the regular image. So perhaps `setSelectedImage()`.

But I think the naming is good enough as it is. Leaving it like this.

PR #2985, opened 2017-04-15
Fix issue with updating an item's group

### LINKED ISSUE #2939: Updating group of existing items
I'm trying to change the group of an existing item in the timeline. However when I do I get an error:
```
Uncaught TypeError: Cannot read property 'itemSet' of null
    at BoxItem.Item.setData (http://visjs.org/dist/vis.js:26603:18)
    at ItemSet._updateItem (http://visjs.org/dist/vis.js:23317:10)
    at ItemSet.<anonymous> (http://visjs.org/dist/vis.js:23066:14)
    at Array.forEach (native)
    at ItemSet._onUpdate (http://visjs.org/dist/vis.js:23050:9)
    at Object.update [as callback] (http://visjs.org/dist/vis.js:22265:12)
    at DataSet._trigger (http://visjs.org/dist/vis.js:8190:20)
    at DataSet.update (http://visjs.org/dist/vis.js:8286:12)
    at <anonymous>:1:7
```

Can I not just call:
`dataSet.update({id: [itemId], group: [groupId]});`
to update the group of an item? Assuming a group with this id exists of course?

I've reproduced the issue in the example here:
`http://visjs.org/examples/timeline/groups/groups.html`

using this code in the Chrome console:
`items.update({id: 0, group: 1});`

Note: Since the grouping of items are random each time, I checked beforehand that the group exists and that the item I updated was not in that group originally.

Am I missing something or is there no way to update the group of an item without removing and re-adding it?

BTMorton:
Add check for parent existence when changing group in Item.setData.
Add subgroup updates to Item.setData.
Remove group updating logic from ItemSet._updateItem.
Fixes #2939

### DIFF
--- a/lib/timeline/component/ItemSet.js
+++ b/lib/timeline/component/ItemSet.js
@@ -1196,14 +1196,6 @@ ItemSet.prototype._addItem = function(item) {
  * @private
  */
 ItemSet.prototype._updateItem = function(item, itemData) {
-  var oldGroupId = item.data.group;
-  var oldSubGroupId = item.data.subgroup;
-
-  if (oldGroupId != itemData.group) {
-    var oldGroup = this.groups[oldGroupId];
-    if (oldGroup) oldGroup.remove(item);
-  }
-
   // update the items data (will redraw the item when displayed)
   item.setData(itemData);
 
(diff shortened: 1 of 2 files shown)


BTMorton:
Because it's duplicated in `Item.setData`, where it calls `ItemSet.moveToGroup` on the parent's item set. What was happening was the `Group.remove` call in `ItemSet._updateItem` was setting `Item.parent = null` which throws an error when trying to make that call. Removing the group update logic from `ItemSet._updateItem` means the error doesn't get thrown and the group is update during the `Item` data update. I think it was supposed to be removed when the `moveToGroup` call was added, but wasn't...

PR #2473, opened 2016-12-17
Add and fix mouse events on timeline

yotamberk:
fixes #226, #2421 and #2429

Issue #2421, opened 2016-12-07
doubleClick Events it not fired.


yubitib:
when "clickToUse" is true once i clicked item i could use click event again i click the same item i could not use click event it not fired.

Tooa:
It works for me with [this](http://jsbin.com/xadoyuhugo/edit?html,console,output) minimal example. I need to click on the timeline once since the option `clickToUse` is activated. Then the timeline is marked with a blue border. Then I click on `item 2` and the the `click` event is fired. A second click on the same element i.e. `item 2` issues another `click` event.

yubitib:
I've confined the option like [this](http://jsbin.com/yuduceziza/edit?html,console,output), not it's not working please check also scroll is not working on the background

yotamberk:
@yubitib Scrolling has been solved in the `develop` branch. In terms of the double click, I'll check it out.

yubitib:
OK Thanks, i need to fix the scroll in your master branch,
may you please help to it?

Tooa:
You can [cherry pick ](https://git-scm.com/docs/git-cherry-pick)the [commit](https://github.com/almende/vis/commit/5f7a721bf9728dd1ac8390b6835c00bb779df251) from the develop branch.

yubitib:
but i used bower for install libraries in my production

Tooa:
For the time being, you can build the develop branch yourself and include your own build in your production env. Otherwise, you have to wait for the next visjs release. However, this shouldn't take too long.

yotamberk:
Fixed in #2473

mojoaxel:
@yotamberk fixed this with #2473

PR #2988, opened 2017-04-16
Fix click and doubleclick events on items

yotamberk:
There was a major bug I encountered trying to catch click and doubleclick events on selected items. 
This PR solves this bug. (continues #2473 to fully solve: #2421, )
checkout the vis/examples/timeline/interaction/eventListeners.html example in this PR to see the click\doubleclick event getting fired when an item is already selected.


lib/timeline/component/item/Item.js
```diff
@@ -145,16 +145,26 @@ Item.prototype.repositionY = function() {
...
+    this.hammer = new Hammer(dragCenter)
 
-    new Hammer(dragCenter).on('doubletap', function (event) {
+    this.hammer.on('tap', function (event) {
+      me.parent.itemSet.body.emitter.emit('click',  {
+        event: event,
+        item: me.id
+      });
+    });
+    this.hammer.on('doubletap', function (event) {
```
wimrijnders: Do `tap` and `doubletap` always fire consecutively? Or is it either one or the other?
In the latter case, please consider usage of [requireFailure()](http://hammerjs.github.io/require-failure/).

lib/timeline/component/item/Item.js
```diff
@@ -145,16 +145,26 @@ Item.prototype.repositionY = function() {
...
       event.stopPropagation();
       me.parent.itemSet._onUpdateItem(me);
+      me.parent.itemSet.body.emitter.emit('doubleClick', {
```
wimrijnders: I am trusting that this block is correct; it is congruent to `click` above, so it looks OK.

lib/timeline/component/item/Item.js
yotamberk: They don't fire consecutively. It's one or the other. There is no need for `requireFailure()` here because it hasn't been used anywhere else. I can add it in another PR.

lib/timeline/component/item/Item.js
yotamberk: it is

yotamberk:
> Also, extra: in the timeline basicUsage example, I get this.groupsData === null in the first test of ItemSet.prototype.groupFromTarget(). Is that supposed to happen?

not sure when this was introduced. 
I'll take a look and fix in another PR

lib/timeline/component/item/Item.js
wimrijnders: OK

lib/timeline/component/item/Item.js
wimrijnders: OK

lib/timeline/component/item/Item.js
```diff
@@ -145,20 +145,20 @@ Item.prototype.repositionY = function() {
...
 Item.prototype._repaintDragCenter = function () {
   if (this.selected && this.options.editable.updateTime && !this.dom.dragCenter) {
-    var me = this
+    var me = this;
     // create and show drag area
     var dragCenter = document.createElement('div');
     dragCenter.className = 'vis-drag-center';
     dragCenter.dragCenterItem = this;
-    this.hammer = new Hammer(dragCenter)
+    var hammer = new Hammer(dragCenter)
```
wimrijnders: semicolon

yotamberk: done

Issue #3002, opened 2017-04-20
Bug in Week display


krazziekay:
By the help from issue #1837, I was able to show the week view in the minor label of the timeline. However, there is a bug when it come to Week 17. The timeline doesn't show week 17 and week 39 in the view.
![image](https://cloud.githubusercontent.com/assets/13076255/25228782/6eeb561c-25ed-11e7-9d83-a8065b7c3527.png)
![image](https://cloud.githubusercontent.com/assets/13076255/25228929/0060bc9a-25ee-11e7-881d-121cc0956a2b.png)

But when I move the timeline, these does show up but kind of messed up as in this pic:
![image](https://cloud.githubusercontent.com/assets/13076255/25228814/8924273e-25ed-11e7-8cca-279694a4d9d8.png)
![image](https://cloud.githubusercontent.com/assets/13076255/25228837/9fb6001c-25ed-11e7-819e-5bfcea965dfb.png)

Can this issue of week view be solved??

nothor84:
Hi!
There are some Tickets already opened regarding the WeekStep, something I think is really mandatory in a Timeline (for Projects).
This is one #1715 and this other is more active #1405

Issue #1715, opened 2016-03-08
Week Step in Timeline doesn't match


nothor84:
Hello!

First of all, congratulations for this super Tool!
Second, sorry for my english i will try to explain it clearly ;)

I'm using this tool to define the milestones for a project and I need to use a "Week Timestep".
I have set my Timeline format options in order to have the following:

When Scale = 'day'
- MinorLabel
  -- Step of 7 days (1 Week) 
- MajorLabel -- month step

But with this i have realized that the step of 7 days in the minorLabel it's restarted every time the major label ends.
For example,
I have a month with 31 days, and this month starts with the first day of a week, then the minorLabel will have the following steps (*\* means wrong):
`day 0, day 7, day 14, day 21, day 28, day 31 (month step), **day 38** (31+7), **day 45** (38+7), ...`
And it should be:
`day 0, day 7, day 14, day 21, day 28, day 31 (month step), day 35 (28+7), day 42 (35+7), ...`
So, the Week step don't fit with the real week steps...

Another visual example:
![weeks_steps](https://cloud.githubusercontent.com/assets/16415053/13597869/1a146b4c-e51b-11e5-8f46-5d0c3e340bd1.png)

There is anyway to fix it?

josdejong:
The Timeline doesn't have a built in scale displaying weeks (repeating every 7 days), so it's not possible to do what you want here. You could change the labels on the axis to display week numbers but not a major grid line every 7 days is not supported.

nothor84:
I see... :(
Well, in fact what I wish is to have an independent scale in the MinorLabel from the MajorLabel. 
For example, I could choose any step in the MinorLabel (for example every 7 days, but also every 3 days, 28 days or whatever) and another in the MajorLabel, and both independent one from the other.
The problem i see here is the followig:
It doesn't matter the scale you have in the MajorLabel, that will be 'forced' in the minorLabel.

This will allow you to have any 'Scale' you want. For example, weeks vs month, lunar month (28days) vs year, or whatever you imagine... But for sure it will also work with the current scales, like days vs weeks, or days vs month, month vs years, etc.

As I said this is a wish but it should be nice! :) 

Anyway thanks for the current work!

marcortw:
Ha! Sounds a bit like [https://github.com/almende/vis/issues/1405](url) which I reported a few months ago.

nothor84:
Yes! "marcortw" at the end the feature we would like to have is, a WeekStep.
[...]

PR #3009, opened 2017-04-23
Implementation of a week scale feature

### LINKED ISSUE #1405: Major gridlines and week display behaviour
I'm working on something which preferably displays data on week level rather than day/exact planning level. I have the following defined in my timeline:

``` javascript
[...]
format: {
            minorLabels: {
                millisecond: 'w',
                second: 'w',
                minute: 'w',
                hour: 'w',
                weekday: 'w',
                day: 'w',
                month: 'w',
                year: 'w'
            },
            majorLabels: {
                millisecond: 'MMM-YY',
                second: 'MMM-YY',
                minute: 'MMM-YY',
                hour: 'MMM-YY',
                weekday: 'MMM-YY',
                day: 'MMM-YY',
                month: 'MMM-YY',
                year: 'MMM-YY'
            }
        },

timeAxis: {scale: 'weekday', step: 7},
[...]
```

I also have a snap function which snaps new elements to the beginning of the week (which is Sunday in my case). This results in the following picture:

gridlines_and_weeks

As I see it, the vertical grid lines seem to be influenced by the beginning of the month, i.e. a new major grid line is calculated at every first of the month, and as it seems, influence the grid lines of that respective month. Take December as an example: I assume that because I defined step to be 7 and the major grid line at each month's beginning can't be ignored, a major grid line is drawn at the 1st + 7 = 8th of December which is a Tuesday. Is my assumption/understanding of the behavior correct?

What I'd like to have is a major grid line starting with every start of the week, ignoring the new grid line at every first day of the month. Incidentally, this year's November is a good example how i'd like to have it: the major grid lines neatly align with the position of my Sunday-snapped items.

Is there a possibility to change this behavior, for example to turn off the major grid line at the 1st of each month?

Best regards
Marco

marcortw:
This PR implements a week scale and should address and fix #1405 #1715 and #3002 . 
Known limitiations:
- Only works properly wehn locales are enabled
- Documentation is a bit sparse
- Introduces minor workarounds to overcome the limitiations of TimeStep.js
- Intentionally not integrated in the auto zoom scale


package.json
```diff
@@ -65,4 +65,4 @@
     "webpack": "^1.14.0",
     "yargs": "^6.6.0"
   }
-}
+}
```
bradh: Looks like this just removed the trailing newline, which you probably didn't intend.

marcortw: Nope, not at all. I had some problems using git and had to copy paste the package.json. Besides, I don't understand why there are 74 commits mostly from others associated with this PR.

krazziekay:
Is the issue regarding the week view fixed??

marcortw:
@krazziekay do you refer to the problem that a major step was not available at a month start? If so, then yes, i implemented a workaround to inject an additional step at each month's begin.

krazziekay:
@marcortw I mean the issue #3002 . Please give me some solutions...

marcortw:
@krazziekay PR #3009 should address your issue through the implementation of a new 'week' scale. However, the PR is open and I hope that one of the maintainers (maybe @yotamberk ?) will find the time to review it. If it will be accepted, it will still take a bit of time to have it available in the released version of the library.

yotamberk:
@marcortw  I'll take a look at this, this weekend

examples/timeline/styling/weekStyling.html
```diff
@@ -0,0 +1,101 @@
...
+<script type="text/javascript">
+    // DOM element where the Timeline will be attached
+    var container = document.getElementById('visualization');
+
+    // just a group for the effects
+    var groups = new vis.DataSet();
+    groups.add([{id: 1, content: "ISO Weeks"}, {id: 2, content: "US Weeks"}]);
+
+    // Create a DataSet (allows two way data-binding)
+    var items = new vis.DataSet([
```
yotamberk: It would be more elegant if you used a `for` loop to create this data

marcortw: In general, I agree, but I thought about this again and think it's maybe not a good idea. First reason: We would already be relying on moment's lib code to determine week numbers to create the items - but I think to prove that things are working properly, we should not rely on something which could fail itself, thus manually writing out the week start and end seems to be fine to me. Second: I specifically wanted to display the week calculation over turn of the year, thus the for loop should always return the same result. But thanks to your comment it came to my mind to extend the examples until the first of May so that we also cover the special case when a (ISO) week start aligns with a month's start.
Nevertheless, if you still strongly vote for a loop, I will implement it.

marcortw:
@yotamberk thanks for the review and the feedback! I'm working on a fix to remove the duplicate week number indication. Could you please explain a bit more in detail what you mean with the 'max week step'? Do your refer to my removal of the week scale from the auto-scale (as discussed in marcortw/vis@782dee4efaed241594679fff45cc5f5fd642b705)?

examples/timeline/styling/weekStyling.html
yotamberk: I strongly vote for a loop. 
I think momentjs is reliable enough

yotamberk:
What I mean is that when you have the week scale, I would expect to also be allowed to zoom in and to be able to see specific days and hours. Don;t you think?

marcortw:
I think this can only be achieved with the use of a non-fixed/auto scale. Whenever a fixed scale is chosen, the labels in the axis will stay as defined. For example, if `{scale: 'month', step: 1}` is set, the minor labels will always show the month and the major labels show the year. But as you can see in marcortw/vis@782dee4, I intentionally didn't implement the week number display in the auto-scale.

yotamberk:
Ok, i'll check this again and merge it once everything looks fine to me. I think future work should be done with allowing a "max step" and "min step" so one can (e.g.) start with a weeks axis, and be allowed to zoom in to seconds .

krazziekay:
please provide the details on how to display this week view. Thank you

yotamberk:
checkout the example he supplied

krazziekay:
When will this be merged in the master branch? How do i get this updated timeline with the week view??

yotamberk:
@krazziekay  pull the `develop` branch in the meantime. A release is planned sometime the next two weeks

marcortw:
cool 👍

danny-larsen:
Can this show the major labels grouped by weeks?

marcortw:
It shows the week number as a minor label and the month and year as a major label (per default).

PR #3046, opened 2017-05-10
Timeline 'showTooltips' option

cameronkloot:
Adds 'showTooltips' option to override popups displayed for items with titles.

### DIFF
--- a/docs/timeline/index.html
+++ b/docs/timeline/index.html
@@ -1027,6 +1027,13 @@ function (option, path) {
     </tr>
 
     <tr>
+      <td>showTooltips</td>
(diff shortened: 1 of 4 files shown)


yotamberk:
I must say, this feature is a bit weird to me because if one doesn't want to see tooltips, I would simply not provide the items tooltips... I can think of few cases needed to have this option, so I'll approve it anyways.

cameronkloot:
I use the option to show/hide tooltips based on user preference, without having to modify the item object itself.  I'd like to start contributing to the project, so I figured I'd start with something easy like this.

PR #3049, opened 2017-05-11
Passing options value from nodes handler to avoid recreating shapes

rasmusblockzero:
*Default text removed for conciseness*

### DIFF
--- a/lib/network/modules/NodesHandler.js
+++ b/lib/network/modules/NodesHandler.js
@@ -160,7 +160,7 @@ class NodesHandler {
       if (options.shape !== undefined) {
         for (let nodeId in this.body.nodes) {
           if (this.body.nodes.hasOwnProperty(nodeId)) {
-            this.body.nodes[nodeId].updateShape();
+            this.body.nodes[nodeId].updateShape(options.shape);
           }
         }
       }


bradh:
@rasmusblockzero Looks like the PR template didn't get deleted.

wimrijnders:
@rasmusblockzero Would you mind replacing the default text with a short description of the change?

I can see it's a one-liner, but this looks like a fix, not a refactorisation. What problem are you fixing?

rasmusblockzero:
The issue is that when I change global node options the options are changed to the new option object for all nodes. 
setOptions(options) {
    this.nodeOptions = options;

Thus, to keep font sizes et.c. I have an option object that I update and pass the whole new object every time to network.setOptions(

But if this new object includes the shape information (as I have the same shapes for all my nodes) the shapes get recreated every time.

Which this little one liner tries to remedy. 
The check could just as well have been made inside the Node.js file (saving what was last created and only change if needed). 

There is a bigger question on whether the way nodes handler handles options but this change would not IMO make that discussion more complicated. 

To me several things regarding nodes handler and each nodes option handling is messy - as mentioned on gitter. I can be part of cleaning that up but this little guy is what I needed to handle a costly and blinky situation so thought I would share it.

rasmusblockzero:
All right, I believe the current implementation is somewhat broken, this take care of that in my application (but I only use one type of shape) but I get your point. A larger change is needed as discussed elsewhere. I'll close this then.

wimrijnders:
...and that is *exactly* the reason why we should spend time on #3060, to get this option handling in a workable form. Really, this is the proof that that work is required.

PR #3059, opened 2017-05-17
mouseup and mousedown events fixes #3032

### LINKED ISSUE #3032: Timeline: Mousedown and mouseup events
Is is possible to add in the mousedown and mouseup events again?

I can see that it was in this pull request: #2473 but was removed again.

We are trying to make it so that on mousedown we start a new event. The user then drags over some days and on mouseup we want to create an event for the selected days.

danny-larsen:
fixes #3032


yotamberk:
Awesome! Thanks!

Issue #3051, opened 2017-05-12
Polymer and visjs timeline : click (selection) doesn't work


rom1504:
I'm trying to put visjs timeline in a [web component](https://www.webcomponents.org/) using [polymer](https://www.polymer-project.org/)

I've managed to get it to work but there is one problem : the click on item doesn't work. Only the long click is detected.

I think polymer or the [shadow dom concept](https://www.html5rocks.com/en/tutorials/webcomponents/shadowdom/) is probably conflicting with visjs handling of clicks.

Here is a demo of that : http://plnkr.co/edit/6OmfKqAl6UKzKbLQViei?p=preview
You can long click on items to select them but not simple click on them.

The update event is also not emitted (double click on item). I guess it works with the same functionnalities.

Is there something I can do in visjs to make this work ?
Is this a visjs bug ?

yotamberk:
Try the version in the `develop` branch. I've fixed the click and double click functionality there

rom1504:
I'm getting that error with the develop branch:
```
Uncaught TypeError: Cannot read property 'getIds' of null
    at ItemSet.groupFromTarget (ItemSet.js:2224)
    at Timeline.getEventProperties (Timeline.js:542)
    at HTMLDivElement.Timeline.dom.root.onmousemove (Timeline.js:143)
```
(I'm not using groups)

rom1504:
I fixed that bug in #3061 
About my problem, doing:
```js
timeline.on("click", (e) => {
          timeline.setSelection(e.item);
        });
```
fixes it.
I'm not sure whether that's something to fix in visjs.

PhilAndrew:
I want to use this timeline as a webcomponent, can you suggest how I can do that?

rom1504:
I've been using it there https://github.com/rom1504/camomile-polymer-client/blob/master/camomile-segmentation.html
It works using the last release (vis 4.20.0).

You can probably use that as an example.

PhilAndrew:
I see your example is a good place to start http://plnkr.co/edit/6OmfKqAl6UKzKbLQViei?p=preview

MeTaNoV:
Indeed working in 'develop', waiting impatiently for the upcoming released version integrating those fixes!

MeTaNoV:
@yotamberk , thanks for reopening this issue.

I create a repo with a simple encapsulation of a build of Vis from the `develop` branch. You can find it here: https://github.com/MeTaNoV

[...]

PR #3061, opened 2017-05-17
fix groupFromTarget when no group is defined in ItemSet.js

rom1504:
Whenever the groupFromTarget function is called in vis timeline and no group are defined this error is displayed in the console : 
```js
Uncaught TypeError: Cannot read property 'getIds' of null
    at ItemSet.groupFromTarget (ItemSet.js:2224)
    at Timeline.getEventProperties (Timeline.js:542)
    at HTMLDivElement.Timeline.dom.root.onmousemove (Timeline.js:143)
```

It for example breaks dragging a custom time bar.

The bug is just an incorrect check of the length of the group : it tries to get the groups when there is a negative number of group, which clearly cannot work. This bug was introduced in the develop branch (the code fixed here is not present in master)

### DIFF
--- a/lib/timeline/component/ItemSet.js
+++ b/lib/timeline/component/ItemSet.js
@@ -2220,7 +2220,7 @@ ItemSet.prototype.groupFromTarget = function(event) {
   var clientY = event.center ? event.center.y : event.clientY;
   var groupIds = this.groupIds;
   
-  if (groupIds.length <= 0) {
+  if (groupIds.length > 0) {
     groupIds = this.groupsData.getIds({
       order: this.options.groupOrder
     });


yotamberk:
Please add a description explaining the bug this solves

rom1504:
@yotamberk I added one

yotamberk:
I understand the problem but I don't understand the solution. 
It seems like the condition checks if the `groupIds` exists before redefining it. I'm not sure adding `groupIds.length > 0` is correct. It might actually be enough to check that `this.groupData` exists.
Notice there is a conflict now. Please check this case further if this condition is even needed.

yotamberk:
related to #3051

rom1504:
indeed 63794a32d833015ee9aa27c351d9dc7904738800 fixed the problem

rom1504:
@yotamberk is it planned to make a new visjs release any time soon ?

bradh:
@rom1504 In progress.

PR #3063, opened 2017-05-17
replacing all ES6 imports with CJS require calls

### LINKED ISSUE #2934: Mix of ES6 imports and common.js exports lead to problems with bundling.
**The problem**
In some cases, in particular with WebPack, it is not possible to bundle a specific component, like the timeline only for example.

**Error that will pop up**
`TypeError: Cannot assign to read only property 'exports' of object '#<Object>'`

Relevant discussions: [Webpack issue #4039](https://github.com/webpack/webpack/issues/4039)
[Vis is even mentioned in the comments](https://github.com/webpack/webpack/issues/4039#issuecomment-281136701)

**Some examples in vis code**
[Timeline line 18 using ES6 import](https://github.com/almende/vis/blob/master/lib/timeline/Timeline.js#L18)
[Timeline line 591 exports as common.js](https://github.com/almende/vis/blob/master/lib/timeline/Timeline.js#L591)

**What should be done**
The simple rule is: never mix ES6 with commonjs import/export in the same file. It can work sometimes but just don't take the chance, a few problems can be avoided simply by following this rule.  For example resolution of circular dependencies. 

Related: There is currently a [discussion about linting / code style / ES6](https://github.com/almende/vis/issues/2697), etc where this is particularly relevant as the linter should not allow such mix.

patsissons:
resolves #2934
used the following regex patterns to apply the changes in `lib`:
`s/import\s+(\w+)\s+from\s+(.*);\s*$/var $1 = require($2).default;/`
`s/import\s+(\w+)\s+from\s+(.*)\s*$/var $1 = require($2).default;/`

There weren't too many instances to patch, 91 total. ran build and tests, no issues on either end.

Open to suggestions if there are any. This PR is really designed as a quick band-aid patch to allow vis module imports from within webpack.

### DIFF
--- a/lib/network/Images.js
+++ b/lib/network/Images.js
@@ -19,8 +19,12 @@ class Images {
      * @return {Image} imageToLoadBrokenUrlOn   The image object
      */    
     _tryloadBrokenUrl (url, brokenUrl, imageToLoadBrokenUrlOn) {
-        //If any of the parameters aren't specified then exit the function because nothing constructive can be done
-        if (url === undefined || brokenUrl === undefined || imageToLoadBrokenUrlOn === undefined)  return;
+        //If these parameters aren't specified then exit the function because nothing constructive can be done
+        if (url === undefined || imageToLoadBrokenUrlOn === undefined)  return;
+        if (brokenUrl === undefined) {
+          console.warn("No broken url image defined");
+          return;
+        }
     
         //Clear the old subscription to the error event and put a new in place that only handle errors in loading the brokenImageUrl
         imageToLoadBrokenUrlOn.onerror = () => {
--- a/lib/network/Images.js
+++ b/lib/network/Images.js
@@ -1,4 +1,4 @@
-import CachedImage from './CachedImage';
+var CachedImage = require('./CachedImage').default;
 
 
 /**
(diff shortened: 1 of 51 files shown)


lib/network/Images.js
```diff
@@ -1,4 +1,4 @@
-import CachedImage from './CachedImage';
+var CachedImage = require('./CachedImage').default;
```
AoDev: `CachedImage` is an ES6 module because it is exported with ES6 export syntax.
`Images` is also an ES6 module.  `export default Images;` at the end.

So it's again mixing styles because it's importing `CachedImage` with a cjs syntax inside an ES6 module.

AoDev: Since you used a regex to search and replace, I suppose it must have happened in other places.

AoDev:
Thanks for the PR but this has also changed imports where it shouldn't.
I commented in some changes.

lib/network/Images.js
patsissons: i'll check for these occurrences. For mixed imports I'll modify the export to CJS style, for unmixed imports i'll revert to ES6 exports.

mojoaxel:
😞

Issue #3057, opened 2017-05-15
Network: DataView and switching from hierarchical to non-hierarchical layout


obglif:
Hello,
I create Network with a DataView object and switch the layout from hierarchical (UD/Hubsize) to non- hierarchical.
When I disable hierarchical and drag a node, it can only moves on the horizontal axis.
But every things works fine when I initialize  the network with the corresponding DataSet.

Example here https://jsbin.com/tolitam/edit?html,js,output

wimrijnders:
Interesting that this apparently works with DataSets, but not with DataViews (according to the jsbin). I'll have a look at it tomorrow.

wimrijnders:
Got it. `NodesHandler.refresh()` assumes the structure of the data is a DataSet:

```
  refresh(clearPositions = false) {
...
      let data = this.body.data.nodes._data[nodeId];  // _data is internal structure of DataSet
...
  }
```

There is a generic call `get()` for both `DataSet` and `DataView` which is better suited here. In other words, the fix is a one-liner.

A further scan of `NodesHandler` shows that this issue does not happen elsewhere. So the one-liner fix is enough.

-----
So: **Confirmed bug**. Thanks for reporting!

PR #3068, opened 2017-05-18
NodesHandler use get(nodeId) instead of _data[nodeId]

wimrijnders:
Fix for #3057.

`NodesHandler.refresh()` accesses member `_data` directly from the nodes. This is an internal structure of `DataSet` and will not work with `DataView`.

Replaced usage of `_data[nodeId]` with `get(nodeId)`, which works for both `DataSet` and `DataView`.


wimrijnders:
Crap! Sorry guys, my mistake. The screen update bit me. Redoing this one for `develop`.

Issue #2818, opened 2017-03-03
[Graph2d] "Cannot read property 'x' of undefined" when changing items via "graph2d.setItems"


cridus:
I have a page in which the user needs to be able to switch to different data via a `<select>`. When initializing the graph with the first set of data there's no error, but once you switch to another set via "graph2d.setItems" this exception is thrown:

    Uncaught TypeError: Cannot read property 'x' of undefined  
      at vis.js:31221  
      at Object.e.insertSort (vis.js:1348)  
      at n.setItems (vis.js:31220)  
      at n._updateAllGroupData (vis.js:29810)  
      at n._onUpdate (vis.js:29627)  
      at n._onRemove (vis.js:29633)  
      at n.setItems (vis.js:29566)  
      at n.setItems (vis.js:29124)  
      at HTMLSelectElement.<anonymous> (VM157:112) <- "graph2d.setItems" when the <select>'s value changes

Here's a JSFiddle that reproduces the issue: https://jsfiddle.net/ntaxc1f5/2/

Previously I was using version 4.15.1 of vis.js and this worked no problem. I updated to 4.18.1 because I had [this issue](https://github.com/almende/vis/issues/450) and I hoped it would be fixed in the latest version, at least for my situation.

Tooa:
I can confirm this issue. The problem is that [here](https://github.com/almende/vis/blob/bc2bc581aaa16905f2ee15b26d3915648c34cc54/lib/timeline/component/LineGraph.js#L523) the passed variable `groupsContent` is empty. It looks like this check [here](https://github.com/almende/vis/blob/bc2bc581aaa16905f2ee15b26d3915648c34cc54/lib/timeline/component/LineGraph.js#L481)  prevents that there is any data written to the array. I haven't figured out the semantic of the check yet.

lazychaser:
Same issue, works fine in v17.0, but there are other issues in this version

blin4444:
This issue seems to happen when calling setItems, with items of the same group as items already in the graph. This functionality last worked in version 4.17.0.

Here is a simple example to demonstrate the problem:

```
<!DOCTYPE html>
<html>
    <head>
        <link rel="stylesheet" href="https://cdnjs.cloudflare.com/ajax/libs/vis/4.19.1/vis.min.css" />

        <!-- Switch to 4.17.0 and the graph shows a horizontal line, as expected -->
[...]

PR #3071, opened 2017-05-19
LineGraph: Remove array pre-fill since unchanged items are skipped anyways

AngeloYoun:
Fix for #2818 .

Not sure if the desired solution was removing the 
```
if (!groupIds && ids && item[fieldId] !== idMap[item[fieldId]]) {
        continue;
}
```
or the array pre-fill but this seemed more performant. 

As mentioned in the https://github.com/almende/vis/issues/2818

Since items not changed are being skipped already, the item count is being thrown off causing the empty object to not be overwritten.


AngeloYoun:
Misread code, fix doesn't achieve the intended result.

PR #3073, opened 2017-05-19
timeline - items always draggable upgrade for range

keksa:
Added new suboption to the itemsAlwaysDraggable option of timeline.

When activated dragLeft and dragRight elements are always added to range items, thus supporting changing of range without the need to select first.

### DIFF
--- a/lib/timeline/component/item/Item.js
+++ b/lib/timeline/component/item/Item.js
@@ -168,8 +168,13 @@ Item.prototype._repaintDragCenter = function () {
     });
 
     if (this.dom.box) {
-      this.dom.box.appendChild(dragCenter);
-    } 
+      if (this.dom.dragLeft) {
+        this.dom.box.insertBefore(dragCenter, this.dom.dragLeft);
+      }
+      else {
+        this.dom.box.appendChild(dragCenter);
+      }
+    }
     else if (this.dom.point) {
       this.dom.point.appendChild(dragCenter);
     }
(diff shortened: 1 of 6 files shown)


lib/timeline/component/item/Item.js
```diff
@@ -169,8 +169,13 @@ Item.prototype._repaintDragCenter = function () {
     });
 
     if (this.dom.box) {
-      this.dom.box.appendChild(dragCenter);
-    } 
+      if (this.dom.dragLeft) {
+        this.dom.box.insertBefore(dragCenter, this.dom.dragLeft);
+      }
+      else {
```
yotamberk: `dragCenter` should be always added. This breaks current behavior

keksa: It does not break it, I will provide an example on monday.

The dragCenter was only added along with the dragLeft and dragRight elements before on selection. It was not added before selection and the dragging still works without it if itemsAlwaysDraggable is enabled.
What I did, is add the dragLeft and dragRight elements before selection if enabled by new option, thus enabling range dragging without selection. The dragCenter is then added on selection before these 2 elements to keep correct z-index.
When the new option is not enabled, all these elements are added on selection and the current behaviour is kept.
I tried it only with the current release though, it's possible it works differently in the develop branch, I'll check it.

keksa: @yotamberk I added the example and checked against develop branch. In the develop branch the `dragCenter` is only added on selection, exactly as it's done in this PR. When `itemsAlwaysDraggable` is true, it is still **not** added, but the dragging works.

PR #3079, opened 2017-05-20
Add support for multiple class names in utils add/remove class methods.

BTMorton:
Reset subgroup heights in resetSubgroups method.

### DIFF
--- a/lib/util.js
+++ b/lib/util.js
@@ -621,11 +621,14 @@ exports.getAbsoluteTop = function (elem) {
  * @param {Element} elem
  * @param {String} className
  */
-exports.addClassName = function (elem, className) {
+exports.addClassName = function (elem, classNames) {
   var classes = elem.className.split(' ');
-  if (classes.indexOf(className) == -1) {
-    classes.push(className); // add the class to the array
-    elem.className = classes.join(' ');
+  var newClasses = classNames.split(' ');
+  for (var i = 0; i < newClasses.length; i++) {
+    if (classes.indexOf(newClasses[i]) == -1) {
+      classes.push(newClasses[i]); // add the class to the array
+      elem.className = classes.join(' ');
+    }
   }
 };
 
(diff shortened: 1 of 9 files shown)


lib/util.js
```diff
@@ -634,12 +637,15 @@ exports.addClassName = function (elem, className) {
...
  */
-exports.removeClassName = function (elem, className) {
+exports.removeClassName = function (elem, classNames) {
   var classes = elem.className.split(' ');
-  var index = classes.indexOf(className);
-  if (index != -1) {
-    classes.splice(index, 1); // remove the class from the array
-    elem.className = classes.join(' ');
+  var oldClasses = classNames.split(' ');
+  for (var i = 0; i < oldClasses.length; i++) {
```
yotamberk: I think you could do this better: 
``` javascript
classes = classes.concat(oldClasses.filter(function (className) {
    return classes.indexOf(className) < 0;
}));
```

BTMorton:
You make fair points.

These are issues that I came across while trying to add dynamic heights to the Timeline for our use case. Each group/item has a class that adds various CSS properties to them based on the type of event, and I was trying to add a second class that changes things based upon the height. I could have done it using more complex single classes or with partial class selectors, but that seemed wrong. I could have changed it to allow for an array of class names, but that would change the timeline options and affect larger parts of the code base, whereas this adds support for it without breaking any existing functionality.

The reason they're together was because I encountered them while working on the same thing, and I was being lazy and trying to sneak them in before the release :P

yotamberk:
Look, I'll approve them for now because the release is tomorrow and I think this PR is important enough to go in. Please don't let this happen again. I don;t like lazy PRs =P 
You could write it better by allowing an array or a string in the second parameter. 
But for now this will do.

BTMorton:
Fair enough, apologies for that. And thanks for accepting it anyway. :)

PR #3088, opened 2017-05-22
Consolidate code for determining the pixel ratio

wimrijnders:
A refactoring of the 'can not unsee' variety.

Consolidate code for determining the pixel ratio in `CanvasRenderer`, with the same code present in `Canvas`.

### DIFF
--- a/lib/network/CachedImage.js
+++ b/lib/network/CachedImage.js
@@ -66,6 +66,10 @@ class CachedImage {
    * This methods takes the resizing out of the drawing loop, in order to
    * reduce performance overhead.
    *
+   * TODO: The code assumes that a 2D context can always be gotten. This is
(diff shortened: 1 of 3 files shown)


wimrijnders:
@bradh Yeah, I agree. Some more thought could be put in combining more related common functionality. E.g. in all cases but one, the pixel ratio is used to set the transform in a context - which invariably comes from the `Canvas` instance, so why should `CanvasRendered` be doing it? Things like that.

But current is a quick fix to remove the hurt in my eyes. Any more works requires non-zero brainpower.

wimrijnders:
Actually, keep this PR open. I'll add some more refactoring for `pixelRatio` RSN (let's say tomorrow).

wimrijnders:
You can review this PR now BTW. I've done enough refactoring for pixel ratio.

PR #3099, opened 2017-05-25
Add option definitions and validation to Graph3d

wimrijnders:
This finally adds the options checking to `graph3d`. I felt the need to do this now, because of expected changes in the options in the near future. Better to have the validation enabled for that.

- Added option definitions to `graph3d`
- Added unit test for checking the validity of `graph3d` default options
- Commented out 'autoByDefault' from default options - module `Validator` can't handle explicit setting of value to `undefined`. The comments now serve as internal documentation
- Checked all `graph3d` examples for correct options, modified these where necessary (notably `playground`)
- Added `onclick` to documentation


yotamberk:
One question: Why do you need to remove options that are undefined? Does the Graph3D not accept undefined options? Does it error?

wimrijnders:
`Validator` chokes on undefined. I've been looking into the code but I can't fix it.
It doesn't understand the difference between a variable that's not there and a variable that is explicitly set to undefined. I gave up on it.

yotamberk:
I don't love it, but I can accept it for now. It's definitely a TODO.

wimrijnders:
Yeah, I agree. But I haven't found a fix for it (yet). Maybe if my head is clearer.

wimrijnders:
Unit test fails now. I'll also see if I can do something about the `undefined` handling.

wimrijnders:
OK, I figured out why the unit tests didn't work. The problem was that I was an idiot. I'm not sure how to fix that, I hope it's just a glitch.

Also, explicit `undefined` is now handled properly in the `Graph3d` options. Removed the silly code in the examples (2 instances) that removed explicitly `undefined` options.

yotamberk:
@wimrijnders conflict to be resolved =)

wimrijnders:
@yotamberk conflict fixed.

.eslintrc
```diff
@@ -16,10 +16,10 @@
     "complexity": [2, 55],
     "max-statements": [2, 115],
     "no-unreachable": 1,
-    "no-useless-escape": 0,
+    "no-console": 0,
```
yotamberk: Are you sure you want this in?

wimrijnders: Yes, for now. `console` is used in the running code to output warning messages.

But this is from a practical viewpoint. I would be happiest if as many checks as possible are enabled for `eslint`. It's just that `console` is used a bit too often to casually fix this.

wimrijnders:
🎉 Yay progress! All `Graph3d` PR's are now merged, which means I can advance it again. But first, bugfixes 😞.

PR #3106, opened 2017-05-26
Refactoring and unit testing of Validator module

wimrijnders:
Changes in this PR:
- Did some obvious refactorings in `Validator.js`, some comment cleanup
- Added a unit test to test these refactorings
- Added module `test-console` to `package.json` to aid the unit tests

The refactoring was mainly done to aid better understanding of the code. In addition, I will note here that this 'better understanding' is required to add some new features in the near future to `Validator`, namely:

- ~~Allow explicitly setting `undefined` for options (required in `graph3d`)~~ **Edit:** This works already, it's a matter of correctly defining value `undefined` for a given field.
- Add the possibility of marking an option `deprecated`.

The unit test has been constructed in a manner that it tests all possible paths affected by the refactoring. Hence, the unit test is not complete; there is room for adding more tests here.


wimrijnders:
Added extra unit test for explicit `undefined` values, to be absolutely sure it will always work as expected.

wimrijnders:
Void comment to see if this triggers the travis test

wimrijnders:
Void comment to see if this triggers the travis test

wimrijnders:
Void comment to see if this triggers the travis test

Issue #3091, opened 2017-05-22
How to get parents from node in network?


sjorsvanheuveln:
I want to extract the parents from a node in a directed hierachical network graph. I couldn't find a method available in the documents. How to do this?

wimrijnders:
There is no method on `network` itself to return these. You best option is to use `DataSet`s and query the data directly.

Assuming you use `DataSet`s, i.e. your data looks something like this:
```
var data = {
  nodes: new vis.DataSet(data_nodes),
  edges: new vis.DataSet(data_edges)
}
```

Then following gets the parent for a given node id:
```
function getParents(nodeId) {
  // Retrieve the from-id's for the given node id
  var result = data.edges.get({
    fields: ['from'],
    filter: function(item) {
      return item.to == nodeId;
    }
  });

  // Make the return value a simple array of id's
  var parentIds = result.map(function(item) {
    return item.from;
  });

  // Get the nodes with the found id's
  var parentNodes = data.nodes.get(parentIds); 
  return parentNodes;
}
```

HTH.

sjorsvanheuveln:
Thanks Wim, this works excellently. Perhaps for larger networks this iterative approach might not be suitable for speed reasons. Perhaps it could be useful that nodes carry the parents and children as inherit variables.

wimrijnders:
The thing is, the data is already present within `network`, but there currently is no way to get to it. There is the method `Network.getConnectedEdges()`, but it doesn't differentiate between parent and child relations. I'm actually seriously considering adding a parameter to it to allow differentiating from, to, and both type edges.

I think it's worth it; you're not the first to ask this question so there's a need. Also, using the network-internal method will be way more efficient than anything done on the outside.

wimrijnders:
OK, I've added the option to the `network` API. It was a small fix which I think will lead to overall good.

Please close the issue if you're satisfied with the resolution.

wimrijnders:
Closed the issue due to being released in `v.4.20.1`.

PR #3108, opened 2017-05-27
Added param 'direction' to Network.getConnectedNodes()

wimrijnders:
This is a better resolution for #3091. The question of getting parents or children only gets asked regularly, so I think it's good to offer the option.

- Added parameter `direction` to `Network.getConnectedNodes()`. Parent-only and child-only nodes can be selected with values `from` and `to` respectively.
- Adjusted the documentation for this change.

*This changes the API, so it presumably will be a MINOR version change.*


wimrijnders:
@bradh :smile:

nadhir-falta:
Any estimate of when this will be released ?

yotamberk:
next week

PR #3121, opened 2017-06-01
Refactoring of `BezierEdgeStatic._getViaCoordinates()`

wimrijnders:
This method is needlessly complicated. Code has been adjusted to better show the intent.

In addition:
- reuse of variables already present, notably `dx` and `dy`
- put recurring code fragments into local variables
- removed second conditions in constructs of the form `if (condition) ... else if (!condition) ...`

The refactoring can be taken further, but it would change the conditional flow, which would complicate reviewing. The current changes highlight the similarities between the code blocks.


wimrijnders:
@yotamberk Thanks for being thorough!

Issue #3133, opened 2017-06-05
check for mass value of zero on vis.js Network nodes.


odranoel75:
I was updating an old library to the latest vis.js network, when I spotted quite an odd behaviour, where only one node out of hundreds would appear.  After a fair amount of debugging, I noticed that I was accidentally setting the mass some of the nodes to zero.   This was causing havoc, with a NaN eventually as the x and y coords.

I've created the following jsfiddle to demonstrate it:
[http://jsfiddle.net/1y8L119L/](http://jsfiddle.net/1y8L119L/)

Setting the mass to 1 or higher solves the issue; however, the code used to work with a mass of 0 back in version 4.3.0.  

Please add the extra safety check to prevent the pain I have just had to go through on other distracted silly programmers that set the mass to zero.

wimrijnders:
There's two years between `4.3.0` and current `4.20.0`. I honestly tried to dig back through time to see what changed, but I can't manage it. So I'll skip trying to clarify what changed in the meantime to make this happen.


But really I have to agree with you. The value zero should be tested in some way and reported. I'll be paranoid and have a look if negative values should be disqualified as well. Thanks for reporting.

----
Background info:

Then, as now, `barnesHut` was the default physics solver, which in the docs comes with the warning for the mass field:

> Values lower than 1 are not recommended. 

Methinks that zero qualifies.

wimrijnders:
It turns out that setting negative mass is a really bad idea. The nodes make a sincere attempt at escaping and disappear into the distance, never to stop. I'm inclined to block negative mass as well.

wimrijnders:
The question now is how to handle this. Should `vis.js` just plain throw an error when this happens? Or perhaps a warning and then reset the masses (to 1 would be most obvious)?

wimrijnders:
I've added the checks, they should be in the next release. Thanks for reporting!

Please close the issue if you're happy with this.

PR #3135, opened 2017-06-06
Protect Network from zero and negative mass values

wimrijnders:
Fix for #3133

Option-field `node.node.mass` must be >= 0.
Checks have been added at the nodes level, for both nodes-global and nodes specific options.
In addition, an internal check has been added for `NodeHandler.defaultOptions`.

The documentation has been adjusted for this change.

### DIFF
--- a/examples/network/events/interactionEvents.html
+++ b/examples/network/events/interactionEvents.html
@@ -48,9 +48,7 @@
         nodes: nodes,
         edges: edges
     };
-
     var options = {interaction:{hover:true}};
-
     var network = new vis.Network(container, data, options);
 
     network.on("click", function (params) {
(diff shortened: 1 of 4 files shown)


examples/network/events/interactionEvents.html
```diff
@@ -48,7 +48,9 @@
         nodes: nodes,
         edges: edges
     };
+
```
bradh: This looks like noise from earlier edits.

lib/network/modules/NodesHandler.js
```diff
@@ -129,6 +129,13 @@ class NodesHandler {
       x: undefined,
       y: undefined
     };
+
+
+    // Protect from idiocy
```
bradh: Perhaps not the best phrasing of "Protect from inadvertent invalid values".

examples/network/events/interactionEvents.html
wimrijnders: Yah. I believe I removed them in a following commit.

PR #3162, opened 2017-06-12
Bidirectional scrolling in timeline - make horizontalScroll and verticalScroll work together

grimalschi:
Hello!

I found there is no bidirectional scroll. We can scroll only vertically or horizontally.

There is two modes, and even in docs we have this:
Notice that defining this option as true will override verticalScroll scroll event but not eliminate the vertical scrollbar.

I did them work together. You can use vertical scroll, horizontal scroll, or both together.

Just look at the updated examples:
Vertical scroll (with vertical+horizontal scroll demo):
http://plnkr.co/edit/zbpeBt4YeUmOpmIj0F42?p=preview
Horizontal scroll:
http://plnkr.co/edit/zGzVxkhSsAngcLMH1xhV?p=preview


lib/timeline/Core.js
```diff
@@ -183,22 +204,31 @@ Core.prototype._create = function (container) {
...
-      this._setScrollTop(adjusted);
-      this._redraw();
-      this.emit('scroll', event);
+      if (this.isActive()) {
+        this._setScrollTop(adjusted);
+        this._redraw();
+        this.emit('scroll', event);
+      }
+    } else if (this.options.horizontalScroll) {
+      if (Math.abs(deltaX) >= Math.abs(deltaY)) var delta = deltaX;
```
yotamberk: The usage of `var` in the `if` is no good. It's very bad practice to redefine a `var` of an existing variable

yotamberk: I would expect to have 
``` javascript
var delta;
if (Math.abs(deltaX) >= Math.abs(deltaY)) delta = deltaX;
else delta = deltaY;
```

It is working when the delta `var` is only defined in the `if` scopes and then used in a different scope, because this is one of those "bad parts of JS". ES6 won't let this work... I'm afraid this is bad practice in general and will be a pitfall when we upgrade to es6.

yotamberk:
Nice work btw! Just that small comment. and I'll get this in

grimalschi:
Done! I used ternary operator for this variable, so it looks now more simple.

If it is not ok, I can change.

grimalschi:
I also updated docs

Issue #3163, opened 2017-06-12
network.findNode() crashes when incorrect id passed.


dheerajbapat:
network.findNode() crashes when incorrect id passed.
Error line: stack.push(this.body.nodes[nodeId].id);
Please can you put a safe check?

wimrijnders:
I'm supposing that you mean the `findNode()` in `Clustering.js`. Is that correct?
I see more occurences of `stack.push()` in this source file, I'll check them all.

wimrijnders:
Confirmed. Preparing a fix for this, also fixing `Clustering.getClusterEdges()`.

PR #3166, opened 2017-06-12
Prevent crashes from invalid id's in `Clustering.findNode()`

wimrijnders:
Fix for #3163

- Added safeguards in said method, to prevent exceptions happening when invalid id's are passed in.
- Adjusted documentation for said method

### DIFF
--- a/lib/network/modules/Clustering.js
+++ b/lib/network/modules/Clustering.js
@@ -689,22 +689,32 @@ class ClusterEngine {
 
   /**
   * Get the stack clusterId's that a certain node resides in. cluster A -> cluster B -> cluster C -> node
-  * @param nodeId
+  *
+  * If a node can't be found in the chain, return an empty array.
+  *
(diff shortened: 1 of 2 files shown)


lib/network/modules/Clustering.js
```diff
@@ -726,6 +736,8 @@ class ClusterEngine {
   * Using a base edgeId, update all related clustered edges with the new options
   * @param startEdgeId
   * @param {object} newOptions
+  *
+  * NOTE: never ever called!
```
bradh: Appears to be used in a couple of examples.

wimrijnders: Ah, OK. Good catch. Didn't grep on those.

I only see it in one example, though: `network/other/changingClusteredEdgeNodes()`. Are there more?

wimrijnders: OK, I'll put some effort in then to fix that one as well.

PR #3170, opened 2017-06-14
Refactoring of Node Drawing

wimrijnders:
The methods `draw()` in all classes derived from `NodeBase` contain the same recurring blocks of code.
These have been consolidated into method instances in `NodeBase`.

In addition, the common instances of method `updateBoundingBox()` has been moved to `NodeBase`.

A lot of source files have been touched, but if you diff, you'll see that it's mostly the common code blocks replaced by a single call.

This refactoring can be taken further, but I didn't want to make the review overly hard.


wimrijnders:
Thanks! :heart:

PR #3175, opened 2017-06-14
Adjusting timeline TimeStep.roundToMinor - fixes #3105

### LINKED ISSUE #3105: Timeline - 4.20.0 - range items seem to be shifted back in time by almost a month
We noticed an issue with version 4.20.0 of the Timeline where range items seem to be shifted back in time by almost a month.

I have put together a jsfiddle for 4.20.0 here: https://jsfiddle.net/2oxhxurw/1/ You should be able to see a timeline like this:

visjs-4 20 0

Here is a fork of the jsfiddle but using 4.19.1: https://jsfiddle.net/cob05xde/1/ You should be able to see a different timeline like this:

visjs-4 19 1

Both examples are using the same config and data, just a different version of visjs.

Interestingly, on 4.20.0, if we set the `timeAxis` to `{ scale: 'day', step: 1 }` then we don't see the issue - https://jsfiddle.net/jr2tp5be/2/

Is there a difference in how we need to supply the data for the newer version?

I am testing with Chrome 58.0.3029.110 (64-bit) on Ubuntu

madebydavid:
- `TimeStep.roundToMinor` function was shifting year and month steps down to the first day of the week.
- Added tests for TimeStep to check the `step` is now working as expected.
- Should fix issue #3105

### DIFF
--- a/lib/timeline/TimeStep.js
+++ b/lib/timeline/TimeStep.js
@@ -148,6 +148,10 @@ TimeStep.prototype.start = function() {
  */
 TimeStep.prototype.roundToMinor = function() {
   // round to floor
+  // to prevent year & month scales rounding down to the first day of week we perform this separately
+  if (this.scale == 'week') {
+    this.current.weekday(0);
+  }
   // IMPORTANT: we have no breaks in this switch! (this is no bug)
   // noinspection FallThroughInSwitchStatementJS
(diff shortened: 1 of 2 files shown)


yotamberk:
Awesome! Thanks!!!

PR #3200, opened 2017-06-22
Fix subgroup collision

### LINKED ISSUE #3199: Non-colliding subgroups cause stacking
When subgroup 1 has an item with an end equal to a second subgroup 2 start time, The subgroups will stack even when `stackSubgroups: false`

yotamberk:
fixes #3199


lib/timeline/component/Group.js
```diff
@@ -498,7 +496,7 @@ Group.prototype._updateSubgroupsSizes = function () {
       })
 
       me.subgroups[subgroup].start = newStart;
-      me.subgroups[subgroup].end = newEnd;
+      me.subgroups[subgroup].end = new Date(newEnd - 1) // -1 to compensate for colliding end to start subgroups;
```
wimrijnders: Why is the conversion to `Date` necessary here? Field `start` doesn't appear to need it, and the code works fine without it.

Also, the result is that it's possible to have the new end become `newEnd - 2`. I know this is nitpicking about milliseconds, but still....I present an alternative in the main review comment, see if that makes sense (I am a virgin with respect to `timeline`).

examples/timeline/items/expectedVsActualTimesItems.html
```diff
@@ -85,13 +85,32 @@
...
+            group:'group1', 
+            subgroup:'sg_3'
+        },
+        {
+            id: 6, 
+            content: 'item 3 (actual time)', 
+            start: '2014-01-17T12:00:00',
+            end: '2014-01-18T12:00:00',
+            group:'group1', 
+            subgroup:'sg_3'
```
wimrijnders: 👍 for updating example.

lib/timeline/component/Group.js
```diff
@@ -486,7 +484,7 @@ Group.prototype._updateSubgroupsSizes = function () {
   if (me.subgroups) {
     for (var subgroup in me.subgroups) {
       var newStart = me.subgroups[subgroup].items[0].data.start;
-      var newEnd = me.subgroups[subgroup].items[0].data.end;
+      var newEnd = me.subgroups[subgroup].items[0].data.end - 1;
 
       me.subgroups[subgroup].items.forEach(function(item) {
         if (new Date(item.data.start) < new Date(newStart)) {
```
wimrijnders: I understand this is not part of your change, but why is it necessary to convert to `Date` here? You can compare millisecond values directly. Is there perhaps a mixup of types for the start and end fields? That they can be either milliseconds or of type `Date`?

I hereby note that all comparisons between `Date` and milliseconds work, except the obvious:

```js
a = 1000000;
b = new Date(1000000);
a == b;    // false!
a === b;   // false!

// But following as expected
a < b;    // false
a <= b;   // true
a >= b;   // true
a > b;    // false
```

Which is just one more inconsistency in javascript. But it does mean that all `new Date()` parts can be removed in the conditions here. Just an observation.

yotamberk: I agree. But because we allow the user to add strings to the `start` and `end` of items, we might encounter comparisons of strings, which is very not good. So as a precaution and to keep consistency in the timeline, Dates are always used for comparisons. I think a better way would have been using `Moment` but This works too

lib/timeline/component/Group.js
yotamberk: Yeah, it can be `newEnd-2` but this is a simple solution for the colliding subgroups. This is the simplest solution to avoid collision of subgroup items. These are by definition a wrapper for items and don't have a real meaning as items. That's why solving their limits by subtracting 1 ms is more correct than fixing the `isColliding` function in `stack.js` file in my opinion.

lib/timeline/component/Group.js
wimrijnders: Ah right, string dates. I haven't thought that far. OK then the comparisons make sense. However, see new points.

lib/timeline/component/Group.js
```diff
@@ -486,7 +484,7 @@ Group.prototype._updateSubgroupsSizes = function () {
   if (me.subgroups) {
     for (var subgroup in me.subgroups) {
       var newStart = me.subgroups[subgroup].items[0].data.start;
-      var newEnd = me.subgroups[subgroup].items[0].data.end;
+      var newEnd = me.subgroups[subgroup].items[0].data.end - 1;
```
wimrijnders: If `data.end` can be a string then `- 1` is not adequate.

lib/timeline/component/Group.js
wimrijnders: If `newEnd` can be a string then `- 1` is not adequate.

lib/timeline/component/Group.js
wimrijnders: Ah right, string dates. I haven't thought that far. OK then the comparisons make sense. However, see new points.

lib/timeline/component/Group.js
yotamberk: I was wrong. It's already converted to a date...

lib/timeline/component/Group.js
yotamberk: The reason I do want to leave the `new Date()` is for consistency. I don't want the `start` to be a Date and the `end` to be a Date in milliseconds.

wimrijnders:
OK, so given that `start` and `end` are always `Date` instancess, I can see that the code will work and I can approve.

For reference, I will hereby record my criticism:

* There is no need to do the interim conversions to `Date` if the start and end fields are `Date` instances already.
* The correction on `end` can be more than 1 millisecond.

But these are not critical notes. The code will work regardless.

yotamberk:
Thanks again @wimrijnders . Notes taken!

Issue #3203, opened 2017-06-23
Image nodes in network are square until interacted with


lucaswerkmeister:
After upgrading from vis 4.19.1 to 4.20.0, images in the network visualization are squeezed to square aspect ratio until the network is somehow interacted with, at which point they assume their correct aspect ratio. (Usually the switch happens as soon as you select a node, sometimes only upon deselection.)

We can reproduce this on Firefox and Chromium (both Ubuntu Xenial).

Small reproducer:

```html
<!DOCTYPE html>
<html>
  <head>
    <meta charset="utf-8">
    <title>Vis Network Image Aspect Ratio Test</title>
    <script src="node_modules/vis/dist/vis.js"></script>
    <link rel="stylesheet" href="node_modules/vis/dist/vis.css">
  </head>
  <body>
    <div id="mynetwork"></div>
    <script>
      new vis.Network(
        document.getElementById('mynetwork'),
        {
          nodes: new vis.DataSet([
            { id: 1, label: 'Test node', shape: 'image', image: 'https://upload.wikimedia.org/wikipedia/commons/thumb/f/f8/Aspect-ratio-16x9.svg/320px-Aspect-ratio-16x9.svg.png' }
          ]),
          edges: new vis.DataSet([])
        },
        {}
      );
    </script>
  </body>
</html>
```
(Assumes that `vis` is in `node_modules`; adjust the paths in `<head>` if necessary.)

Before interacting with the graph:
![screenshot from 2017-06-23 13-32-57](https://user-images.githubusercontent.com/2346599/27480511-84c722f6-5818-11e7-8f29-760a82d8c44f.png)
After selecting and deselecting the node:
![screenshot from 2017-06-23 13-33-04](https://user-images.githubusercontent.com/2346599/27480527-938e6aa6-5818-11e7-9757-a321304e3df5.png)

From a quick glance over the release notes, #3010 sounds like it might be responsible for this.

lucaswerkmeister:
Here’s a live example with a slightly larger graph on query.wikidata.org: http://tinyurl.com/yazbfwy8 – as soon as you click one of the nodes, all of the images correct their aspect ratio (so it’s not per-node – I forgot to mention that in the issue description).

(Note that we might roll back to 4.19.1, which would break the live example. See [T168714](https://phabricator.wikimedia.org/T168714) for more information on what we’re doing about this bug.)

wimrijnders:
Confirmed. I've seen this happen as well with SVG's.

[...]

PR #3205, opened 2017-06-24
Set dimensions properly of images on initialization.

wimrijnders:
Fix for #3203

Image nodes were assigned the default size on initialization, leading to very compressed images.
This fix adjusts the default size as soon as the images used have been loaded.

The approach for dealing with this has been adapted from `CircularImage`.

### DIFF
--- a/lib/network/CachedImage.js
+++ b/lib/network/CachedImage.js
@@ -23,6 +23,7 @@ class CachedImage {
   init() {
     if (this.initialized()) return;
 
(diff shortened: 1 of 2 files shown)


wimrijnders:
ES6: weirdness. I had the impression that the `package.json` dealt with all this properly. Would [this link](https://markgoodyear.com/2015/06/using-es6-with-gulp/) help in any way?

PR #3206, opened 2017-06-25
Nested groups order logic 

### LINKED ISSUE #3126: Timeline nested groups work incorrect if set them after timeline init
I take an example from http://visjs.org/examples/timeline/groups/nestedGroups.html and initialize the timeline before creating groups and elements
Http://jsfiddle.net/ildarnm/dkfo75f9/

yotamberk:
fixes #3126 and refactor of nested groups logic


wimrijnders:
For my understanding: the issue is that the nested groups appear in the wrong location, correct?

So, `cook`, `clean house` and `shop` should be placed directly under `Lee`.

Do I understand the issue like this?

lib/timeline/component/ItemSet.js
```diff
@@ -146,6 +146,31 @@ function ItemSet(body, options) {
   this.groupListeners = {
     'add': function (event, params, senderId) {
       me._onAddGroups(params.items);
+      
+      if (me.groupsData && me.groupsData.length > 0) {
+        var groupsData = me.groupsData;
+        if (me.groupsData instanceof DataView) {
+          groupsData = me.groupsData.getDataSet()
```
wimrijnders: Same remark for line `1687`.

lib/timeline/component/ItemSet.js
```diff
@@ -1063,7 +1088,7 @@ ItemSet.prototype._onAddGroups = function(ids) {
   ids.forEach(function (id) {
     var groupData = me.groupsData.get(id);
     var group = me.groups[id];
-
+
```
wimrijnders: Pedantic overdrive: spaces in an empty line.
*(I might be going over the top here, but I really do intend to review to the best of my capabilities)*

lib/timeline/component/ItemSet.js
```diff
@@ -1663,16 +1688,18 @@ ItemSet.prototype._onGroupClick = function (event) {
     groupsData = this.groupsData.getDataSet()
   }
 
-  group.showNested = !group.showNested;
+  var nestingGroup = groupsData.get(group.groupId)
+  nestingGroup.showNested = !nestingGroup.showNested;
 
   var nestedGroups = groupsData.get(group.nestedGroups).map(function(nestedGroup) {
     if (nestedGroup.visible == undefined) { nestedGroup.visible = true; }
```
wimrijnders: Result of this overwritten by next line.

lib/timeline/component/ItemSet.js
```diff
@@ -1663,16 +1688,18 @@ ItemSet.prototype._onGroupClick = function (event) {
...
-    nestedGroup.visible = !!group.showNested;
+    nestedGroup.visible = !!nestingGroup.showNested;
```
wimrijnders: Why not a direct assignment? `nestingGroup.showNested` is forced to boolean by line `1692`.

lib/timeline/component/ItemSet.js
```diff
@@ -146,6 +146,31 @@ function ItemSet(body, options) {
...
+              groupData.showNested = true;
+            }
+            groupData.nestedGroups.forEach(function(nestedGroupId) {
+              var updatedNestedGroup = groupsData.get(nestedGroupId);
+              if (!updatedNestedGroup) { return; }
+              updatedNestedGroup.nestedInGroup = groupData.id;
+              if (groupData.showNested == false) {
+                updatedNestedGroup.visible = false;
+              }
+              groupsData.update(updatedNestedGroup);
```
marcortw: Could it be that the senderId of the initial change is not forwarded here?

tfsarlo:
is there a way to use multiple levels of groups and subgroups? so far i have seen only 2 levels.

lib/timeline/component/ItemSet.js
yotamberk: done

lib/timeline/component/ItemSet.js
yotamberk: true. It was left when it wasn't forced to be boolean

yotamberk:
@wimrijnders please review again =) I addressed the comments.
not my best PR... Good thing you've reviewed it well!

lib/timeline/component/ItemSet.js
```diff
@@ -1662,21 +1685,19 @@ ItemSet.prototype._onGroupClick = function (event) {
...
 
-  var groupsData = this.groupsData;
-  if (this.groupsData instanceof DataView) {
-    groupsData = this.groupsData.getDataSet()
-  }
+  var groupsData = this.groupsData.getDataSet();
 
-  group.showNested = !group.showNested;
+  var nestingGroup = groupsData.get(group.groupId)
+  nestingGroup.showNested = !nestingGroup.showNested;
```
marcortw: When a nesting group is initialy loaded as 'open' (the implicit "showNested: true"), showNested is undefined. Negating it turns it to true which means expand, correct? This causes that an open loaded nesting group needs to be clicked twice. Adding this the line before should do the trick:

`if (nestingGroup.showNested == undefined) nestingGroup.showNested = true;`

lib/timeline/component/ItemSet.js
```diff
@@ -1662,21 +1685,19 @@ ItemSet.prototype._onGroupClick = function (event) {
...
   var nestedGroups = groupsData.get(group.nestedGroups).map(function(nestedGroup) {
-    if (nestedGroup.visible == undefined) { nestedGroup.visible = true; }
-    nestedGroup.visible = !!group.showNested;
+    nestedGroup.visible = nestingGroup.showNested;
     return nestedGroup;
   });
-  groupsData.update(nestedGroups);
 
-  if (group.showNested) {
+  groupsData.update(nestedGroups.concat(nestingGroup));
```
marcortw: I think it's a bit problematic to not have a sender here. But introducing something like a fake sender like 'nesting' wouldn't be clean either. My problem is that I catch the update event on the groups and do a lot of stuff afterwards. If there's no sender, I assume that the user modified the group, but in this case the user only cliked on the expand/collapse icon and didn't change the groups themselves. Let me illustrate this:

```
that.groups.on('update', function (event, properties, senderId) {
	switch (senderId) {
	case 'api':
		// do nothing / accept the changes
		break;
	case 'initialization':
		break;
	default:
		// item is supplied locally from the user/the library
		let currObj = that.groups.get(group.id);
		// --> update the changes to the API
	}
});
```
With your changes, I may now perform a rather complicated object comparison to see if the changes were only in the visibility of the group, but that's a bit cumbersome. Nevertheless, I have no good idea on how to approach this and I can live with it.

lib/timeline/component/ItemSet.js
yotamberk: Yeah... I had that line removed because I though it had no use. I'll re-add it

lib/timeline/component/ItemSet.js
yotamberk: I'll take another look at it now.

lib/timeline/component/ItemSet.js
wimrijnders: Ah, that was due to my review comment. Fine, so it needs to be in.

wimrijnders:
The travis test failing is at least partly due to #3313; it's not due to `timeline`. It's a unit test failing for `network`.

#3313 has been fixed to get rid of this testing error.

marcortw:
@yotamberk just a quick question: Are we supposed to push the child group id to the parent's nestedGroups array and then update the parent group? Or the child group as well? I only updated the parent group and it seems fine except for that one odd effect.

I am not sure if my code is buggy or if there's still a little bug in this PR - I experience an effect which (sometimes?) seems to cause that the last added child group is not properly nested in its parent group. It seems like an update to a child group's nestedInGroup property happens only after a subsequent update to any parent group. My updates lazily sneak in through an API. I might be totally wrong though, which is why I'd like to know how exactly an update is supposed to be executed by the library user.

yotamberk:
@marcortw Unfortunately this PR still does not resolve the problem of initially hidden nested groups so the issue you are dealing with is still open (there is an issue about it somewhere). It will need to be solved at some point. I'l currently trying to solve some other issues juggling my day job. I hope to be in holiday next week and pick up this issue aswell

Issue #3185, opened 2017-06-18
Width constraint does not break words...


jmvenancio:
I set the `widthConstraint` to `maximum: 300` and `minimum: 300`.

This is what happens if I write a word that has a bigger "length" than 300.

![image](https://user-images.githubusercontent.com/15384201/27265035-61f6335c-5485-11e7-8dc3-1d09b17ee47c.png)

Is there anyway to overcome this problem?

Thanks in advance!

wimrijnders:
This is how the handling of a single, long word works at the moment. It gets put on a line by its own and the `widthConstraint.maximum` is broken.

I have the following questions:

- Is there a real-live use case for this? Does it actually happen that such a long word is present?
- What would you expect to happen? Is the second image perhaps a representation of what you would expect?

bradh:
An overflow might make more sense.

wimrijnders:
By coincidence, I'm actually working on label break-down into lines, see #3171.

This issue made me ponder a bit; I sort of came to the conclusion that if `widthConstraint.maximum` is defined, it should be respected always. So really, big words should just be broken into lines always.

Here is some test output, what do you think of following?

Without width maximum:
![untitled2](https://user-images.githubusercontent.com/769037/27273648-36acb054-54d0-11e7-9339-422095affd2f.png)

With width maximum:
![untitled](https://user-images.githubusercontent.com/769037/27273654-3b66e128-54d0-11e7-941b-8a6b276d4619.png)

wimrijnders:
> Is there anyway to overcome this problem?

To answer your question, the only way around it now, is to shorten the label yourself.

Otherwise, wait for the fix as described above to be released.

jmvenancio:
Thank you for your answer.

That behaviour is what I was expecting. Do you know when will it be the next release? Should I put my effort on breaking words myself or should I wait for the next release? 

Answering your questions:

The future users of my project can do whatever they want... So, although it is not likely to write a 1000 letters word, they can do it. And of course the word should be broken to the next line (and perhaps adding a `-` in the previous line, even if it does not respect the language rules) what do you think?

wimrijnders:
> Do you know when will it be the next release?

It is planned next weekend (24 jun 2017). However, this will not be in it, it's too late to do the full review cycle.

> Should I put my effort on breaking words myself or should I wait for the next release?

If you need it short term, then yes, put the effort into it. At the earliest, the release it will be in will be done in more than a month.

> ...what do you think?

[...]

Issue #3171, opened 2017-06-14
Network: nodes.widthConstraint: similar min/max values cause infinite loading loop


cityy:
Creating a network with similar minimum/maximum values for nodes.widthConstraint causes an infinite loading loop. I stumbled across the issue trying to set circle shape nodes to all have the same size.

The following fails to load:
```
var options = {
nodes:{
  widthConstraint:
     { minimum: 150, maximum: 170,}
   }
};

network = new vis.Network(container, data, options);
```
I've been able to reproduce the issue by simply modifying the Network Node Shapes example: http://visjs.org/examples/network/nodeStyles/shapes.html

It's really weirdly breaking if min/max values get too close to each other:
```
{ minimum: 150, maximum: 150,} // fails
{ minimum: 150, maximum: 175,} // fails
{ minimum: 150, maximum: 176,} // works
{ minimum: 150, maximum: 190,} // works 
```

**Update:** Just a note. The exact values that cause it to fail differ between my current project and the node shapes example.

wimrijnders:
Got a fix on the infinite loop. It's in `Label._processLabel()` roundabout line 910.

In a nutshell:

```js
            while (w < words.length) {
...
              if ((measure.width > this.fontOptions.maxWdt) && (lastMeasure.width != 0)) {
                // Do things here, variables in if() don't change
                // w not incremented!
              } else {
                // Do other things here
                w++;
              }
            }

```

A situation occurs that the if-condition always holds and thus is always entered. Variable `w` is not incremented and infinity ensues. It seems to happen at the end of a label text where the loop processes an empty string.

@cityy Great catch, and thanks for supplying instructions to reproduce. Here, have a virtual beer on me :beer:.

cityy:
@wimrijnders My pleasure!

satyajitsinh:
I also fond similar behavior.
`Label._processLabel()` also an infinite loading loop in below condition.
I had a network with node shape 'box' and applied widthConstraint.

1. If I apply large label with only one word box will expand breaking widthConstraint.maximum.
2. If I apply large label with two words where first word is smaller and second is large then network will go in infinite loading loop

wimrijnders:
@satyajitsinh:
1. confirmed, although I really don't know what to do about this. Seems to me that this behavior is the best option. Do you have a better idea perhaps?

[...]

PR #3228, opened 2017-07-03
Fix infinite loop on drawing of large labels

wimrijnders:
Fix for #3171 and #3185.

- Fixed the infinite loop which can occur in specific case when drawing node labels.
- Labels with one huge word longer than width constraint are now broken into multiple lines.

This can lock up `Network` due to the infinite loop, so I feel that the PRIORITY label is in order here. You can't have that happening in cases that users would regard as 'normal usage'.

The actual fix for the issue (see above) is the code in function `splitStringIntoLines()`.

----
Changes:

- Added new local class `LabelAccumulator` to `Label.js`. This moves away recurring actions within the main loops of the label handling.
- Moved code specific to splitting up label text to new method `Label._processLabelText()`
- Added unit tests for this method
- Added unit tests for proper handling of html and markdown tags
- Added unit tests for checking proper handling of `widthConstraint.maximum`.
- Added commenting, notes and TODO's where appropriate.


wimrijnders:
### Some extra info

I'm afraid that this fix is going to be pretty tough to review, if you're relying on diffing.
The code in question has changed significantly; it was pretty convoluted, and I found myself cleaning it up and refactoring just to understand it. It was not my intention at all to make such a drastic change, but I believe it is for the better.

The goal of this refactoring is to make the code for the label splitting clearer.
This does not mean it has become simpler, but the code has been rearranged to make it more understandable.

I have added unit tests to be sure that all is working as expected. This also tests the markup and tags for when using multi-fonts. The unit tests may be expanded in the future.

The actual fix for the issue (see above) is the code in function `splitStringIntoLines()`.

----
Changes:

- Added new local class `LabelAccumulator` to `Label.js`. This moves away recurring actions within the main loops of the label handling.
- Moved code specific to splitting up label text to new method `Label._processLabelText()`
- Added unit tests for this method
- Added unit tests for proper handling of html and markdown tags
- Added unit tests for checking proper handling of `widthConstraint.maximum`.
- Added commenting, notes and TODO's where appropriate.

jmvenancio:
I started testing and found a problem already :( @wimrijnders 

[(https://i.gyazo.com/6e9c441b4045c43a780ab2861a836458.mp4)]

jmvenancio:
It only happens after the zoom exceed 1.2

wimrijnders:
@jmvenancio What exactly is the problem? What I see is that the particular subject of this fix works as expected.

Perhaps you mean the 'flash' that happens when the node height is larger than the view height? And the whole view turns the color of the node? I would call this a separate issue, and to be honest not a very critical one. But perhaps you differ in opinion of this.

jmvenancio:
@wimrijnders you're right. It is a different issue. Although it still does not work as expected, because on that particular case the width constraint is broken. And of course that is not a critical issue, I will continue testing but I'll mention this problem on the review.

Thank you for all your work!!

wimrijnders:
> you're right. It is a different issue.

@jmvenancio If it really bothers you, please make a new issue out of it.

As for the second issue, you mean the overflow in the big rectangle on the left, right?
It looks like there's a max height constraint on that rectangle. I only handled the width in this fix, and my thought is that then the overflow is a separate issue. If it's not due to a max height constraint, please show me example code that produced that.

I have to say you're very much stress-testing this fix. I would also say that that is a good thing.

jmvenancio:
I am not really sure it is only about the height constraint. Notice on the left green box, the "a a a" goes out of the rectangle, on the other hand on the right green box the "bolas bolas" does not go out and it has a superior height. And I set the maximum height to the height of the small rectangles so, maybe two issues should be created.

In my specific case this fix isn´t enough since I really want all the nodes to have the same size. Also I think that the maximum height constraint should always be respected, much like width constraint is with this fix... And maybe this is another issue...

wimrijnders:
@jmvenancio Could you give me the code that made the previous screen dump? I'd like to examine that in more detail.

yotamberk:
@wimrijnders what's the status for this PR? Can it be reviewed?

wimrijnders:
@yotamberk Yes, it can definitely be reviewed. The core issue here is the fix on infinite loop. The discussion here is about an edge case with respect to layout.

@jmvenancio Could you open a new issue, please, for the overflowing of the big labels? I think it is worthwhile to pursue this, but not in the current PR. Thanks.

wimrijnders:
@Maniae That went straight into the unit tests. Unfortunately, the test succeeds. Would you mind adding some minimal code in order to reproduce the error?

I'm hereby assuming that you are using the `develop`-branch, since `LabelAccumulator is not in production yet. If I can confirm this, I'm going to ask you to open a new issue on this.

In any case, thanks for reporting!

Issue #3220, opened 2017-06-30
[Network] GUI Manipulation, node level doesn't update visually


cridus:
I have a Network with a hierarchical tree. I added the GUI buttons to add and edit the nodes, however when I edit a node and I change its level, the node doesn't move to that level, visually. As a workaround I call `network.clustering.updateClusteredNode(data.id, {level : theNewLevel});` after the callback to update the node and that does make the node appear in its new position.

Here's a fiddle that shows the issue: https://jsfiddle.net/y2kwv76v/3/
Edit a node changing its level and you'll see it won't move.
After that, click on the button "Use workaround" and the node will go to the right level.

For some reason the fiddle only works on Chrome...

wimrijnders:
It didn't work for me either in chromium. When I remove this line from options:

```
  locale: navigator.language, 
```
...all was well, also in firefox.

Hey, I know that network! Great example BTW.

wimrijnders:
Adding a new node appears to work fine. So the problem is in the update. Digging in...

I believe it's the following:
* when *adding* a node, `LayoutEngine.setupHierarchicalLayout()` is called, which re-initializes the layout, thus determining the position of the new node.
* This method is not called when *updating* a node, thus a new position is not calculated. The workaround forces the recalculation.

Thus, the logical thing to do is to force the recalculation on the update as well. However, it's a bit overkill to do this on every arbitrary change in the node data; `LayoutEngine.setupHierarchicalLayout()` is a pretty hefty method to run.

I propose to add a check on update if option `level` has changed; if so, do the recalculation. There may be other node data that have an effect; I can't think of any right now but these can be added as needed.

wimrijnders:
Fix has been submitted. It was pretty straightforward after preceding analysis.

PR #3239, opened 2017-07-06
Update hierarchy when node level changes

wimrijnders:
Fix for #3220

On change of data of an existing node, field `level` is checked for change.
If changed, force a recalculation of the hierarchical layout.

- The fix does not explicitly check for hierarchical layout; this should not be a problem.
- The added code can be used to add further node fields which may trigger recalculation of layout.

### DIFF
--- a/lib/network/modules/NodesHandler.js
+++ b/lib/network/modules/NodesHandler.js
@@ -17,7 +17,7 @@ class NodesHandler {
 
     this.nodesListeners = {
       add: (event, params) => { this.add(params.items); },
-      update: (event, params) => { this.update(params.items, params.data); },
+      update: (event, params) => { this.update(params.items, params.data, params.oldData); },
       remove: (event, params) => { this.remove(params.items); }
     };
 
@@ -273,10 +273,12 @@ class NodesHandler {
 
   /**
    * Update existing nodes, or create them when not yet existing
-   * @param {Number[] | String[]} ids
+   * @param {Number[] | String[]} ids id's of changed nodes
+   * @param {Array} changedData array with changed data
+   * @param {Array|undefined} oldData optional; array with previous data
    * @private
    */
-  update(ids, changedData) {
+  update(ids, changedData, oldData) {
     let nodes = this.body.nodes;
     let dataChanged = false;
     for (let i = 0; i < ids.length; i++) {
@@ -285,7 +287,9 @@ class NodesHandler {
       let data = changedData[i];
       if (node !== undefined) {
         // update node
-        dataChanged = node.setOptions(data);
+        if (node.setOptions(data) === true) {  // explicit test because setOptions() can return undefined
+          dataChanged = true;
+        }
       }
       else {
         dataChanged = true;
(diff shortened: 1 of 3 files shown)


lib/network/modules/NodesHandler.js
```diff
@@ -285,7 +287,9 @@ class NodesHandler {
       let data = changedData[i];
       if (node !== undefined) {
         // update node
-        dataChanged = node.setOptions(data);
+        if (node.setOptions(data) === true) {  // explicit test because setOptions() can return undefined
```
yotamberk: why isn't ` if (node.setOptions(data))` enough?

lib/network/modules/NodesHandler.js
```diff
@@ -294,6 +298,24 @@ class NodesHandler {
         nodes[id] = node;
       }
     }
+
+    if (!dataChanged && oldData !== undefined) {
```
yotamberk: Why is ` if (!dataChanged && oldData)` enough?

lib/network/modules/NodesHandler.js
```diff
@@ -294,6 +298,24 @@ class NodesHandler {
...
+      // Check for any changes which should trigger a layout recalculation
+      // For now, this is just 'level' for hierarchical layout
+      // Assumption: old and new data arranged in same order; at time of writing, this holds.
+      for (let i = 0; i < changedData.length; i++) {
+        let oldValue = oldData[i];
+        let newValue = changedData[i];
+
+        if (oldValue  !== undefined && newValue !== undefined) {  // paranoia
```
yotamberk: `(oldValue  !== undefined && newValue !== undefined)` bleh...

lib/network/modules/NodesHandler.js
```diff
@@ -294,6 +298,24 @@ class NodesHandler {
         nodes[id] = node;
       }
     }
+
+    if (!dataChanged && oldData !== undefined) {
+      // Check for any changes which should trigger a layout recalculation
+      // For now, this is just 'level' for hierarchical layout
+      // Assumption: old and new data arranged in same order; at time of writing, this holds.
+      for (let i = 0; i < changedData.length; i++) {
```
yotamberk: this loop could be written nicer:
```javascript
dataChanged = changedData.some(function(newValue, index){
  var oldValue = oldData[index]
  return (oldValue  && oldValue.level !== newValue.level)
})
```

lib/network/modules/NodesHandler.js
wimrijnders: `Node.setOptions()` can return `undefined`.

...yes, very ugly. Let's see if we can change that.

wimrijnders: OK, I've convinced myself that the handling in case of `undefined` is good. Adjusting it.

lib/network/modules/NodesHandler.js
wimrijnders: Well, I like explicit testing because I'm paranoid with respect to conditions in the code. But I'll admit that I can go over the top sometimes.

lib/network/modules/NodesHandler.js
wimrijnders: TIL

lib/network/modules/NodesHandler.js
wimrijnders: :stuck_out_tongue_closed_eyes:

PR #3247, opened 2017-07-10
Update index.html to reflect 4.20.1

bradh:

### DIFF
--- a/index.html
+++ b/index.html
@@ -29,7 +29,7 @@
 
     <script src="./js/smooth-scroll.min.js"></script>
 
-    <script src="https://cdnjs.cloudflare.com/ajax/libs/vis/4.20.0/vis.min.js"></script>
+    <script src="https://cdnjs.cloudflare.com/ajax/libs/vis/4.20.1/vis.min.js"></script>
 
     <script language="JavaScript">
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -37,7 +37,7 @@
         function evalLinks() {
             if (typeof vis === 'undefined') {
                 console.log(document.getElementById("linkStatus"));
-                document.getElementById("linkStatus").innerHTML = "Note: The latest version (4.20.0) is not yet available on cdnjs, <a href='https://cdnjs.com/libraries/vis'>click here</a> to to pick the latest available version.<br />";
+                document.getElementById("linkStatus").innerHTML = "Note: The latest version (4.20.1) is not yet available on cdnjs, <a href='https://cdnjs.com/libraries/vis'>click here</a> to to pick the latest available version.<br />";
```
YMIndustries: This check is not accurate because 4.20.0 is still loaded on the page. Need to update the version loaded on the page to 4.20.1 as well in order to correctly detect if new Vis version is available on cdnjs. (Spoiler: It already is, but it's still worth getting this check right)

bradh: Should be fixed now

Issue #3251, opened 2017-07-12
Graph3d setOptions throws exception setData does not exist


DaveWare:
I just started trying to use Graph3d yesterday and used the available example code 10_styling.html to reproduce the excpetion. Btw; nice work! I changed the example to call setBarWidth() when x or y width is changed.

I cloned the latest repository and am using the distribution:
 * @version 4.20.1
 * @date    2017-07-02

I wanted to allow the users to dynamically set the value range for coloring. When I pass updated options to `graph3d.setOptions()` an exception is throw from `DataGroup.reload line 13850` that says DataGroup does not have a setData method. I changed reload to take a `graph3d` as an argument and called `graph3d.setData(this.dataTable)` rather than `this.setData(dataTable)`. This seems to have the correct behavior. The graph updates and has the new value range. Or in the case of the example code below the correct bar width.

Not sure if this would be the proper fix as I have just started looking at your component.

> Object doesn't support property or method 'set

```js
  /**
   * Reload the data
   */
  DataGroup.prototype.reload = function () {
    if (this.dataTable) {
      this.setData(this.dataTable);
    }
  };
```

10_styling.html
[10_styling.txt](https://github.com/almende/vis/files/1142783/10_styling.txt)

wimrijnders:
OK, great, a `graph3D` issue. I would be happy to help, but would you mind cleaning up your comment a bit? Seems like some text is missing. I cleaned it up as far as I could.

wimrijnders:
I think you're on to something here. I must have missed this in the refactoring to `DataSet`. I'll look at it in depth tomorrow. Thanks for reporting!

wimrijnders:
Confirmed. This is a silly oversight. I owe you a beer for finding this.

Your fix works, but I find it a very roundabout way of setting the data. I propose the following, which will do exactly the same thing:

```js
/**
 * Update the options. Options will be merged with current options
 *
 * @param {Object} options
 */
Graph3d.prototype.setOptions = function (options) {
...
  // re-load the data
  //this.dataGroup.reload();
  var dataTable = this.dataGroup.getDataTable();
  if (dataTable) {
    this.setData(dataTable);
  }
...
};

```

This gets rid of the dependency to the `Graph3D` instance within `DataGroup`. In fact, `DataGroup.reload()` can just plain die now.

I'll make a PR for this. I'm considering adding a fix to pending PR #3099. Again, thanks for reporting.

DaveWare:
[...]

PR #3255, opened 2017-07-14
Fix missing reference to Graph3D instance in DataGroup

wimrijnders:
Fix for #3251.

A reference was missed in `DataGroup.reload()` during refactoring.
Rather than fix this method, it has been removed and the logic moved to `Graph3d`.
This makes for somewhat cleaner code.

### DIFF
--- a/lib/graph3d/Graph3d.js
+++ b/lib/graph3d/Graph3d.js
@@ -171,9 +171,7 @@ function Graph3d(container, data, options) {
   this.setOptions(options);
 
   // apply data
-  if (data) {
-    this.setData(data);
-  }
+  this.setData(data);
 }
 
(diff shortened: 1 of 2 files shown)


lib/graph3d/Graph3d.js
```diff
@@ -702,7 +702,7 @@ Graph3d.prototype.setOptions = function (options) {
   this._setSize(this.width, this.height);
 
   // re-load the data
-  this.dataGroup.reload();
+	this.setData(this.dataGroup.getDataTable());
```
bradh: tab vs space?

wimrijnders: Space of course. Spaces rule!

PR #3262, opened 2017-07-16
Upgrade packages and tools for Travis unit testing

wimrijnders:
This addresses the warnings and notifications that pop up in the job log of the `travis-cl` unit tests.
[This job log](https://travis-ci.org/almende/vis/builds/253886493?utm_source=github_status&utm_medium=notification) has been used as a starting point. The goal is to resolve all warning/error messages (**Update:** The goal has failed).

Compare latest job log of this PR with the previously linked job log for results.

----
**For review, please do an `npm install` from scratch and run the unit tests**
It all works on my own machine and on `travis-cl`, but YMMV due to local platform differences.

----

### Changes

- upgrade compiler to C++11 as per [these instructions](https://docs.travis-ci.com/user/languages/javascript-with-nodejs#Node.js-v4-%28or-io.js-v3%29-compiler-requirements)
- add `canvas` module and minimal required components (`libgif`). This module will be used for unit testing `Network` and `Graph3D`.
- Upgraded versions of modules in `packages.json` as much as possible/necessary. This required some fixes in `gulpfile.js` for usage of `webpack`.
- Small fix in `Network` due to linting

### Gulp latest stable links to deprecated modules

During `npm install` the following warnings are displayed:

> npm WARN deprecated minimatch@2.0.10: Please update to minimatch 3.0.2 or higher to avoid a RegExp Dos issue

(Also for `minimatch@0.2.14`)
> npm WARN deprecated graceful-fs@1.2.3: graceful-fs v3.0.0 and before will fail on node releases >= v7.0. Please update to graceful-fs@^4.0.0 as soon as possible. Use 'npm ls graceful-fs' to find it in the tree.

Both of these warnings are traceable to `gulp v3.9.1` (latest stable), which has not been updated wrt. these warnings. Try as I might, I can't force the versions. We'll just have to wait till the `gulp` developers fix this.

### Warning on `webpack` when running `gulp`
With the upgraded `webpack v2.x.x`, the following notice may appear when running `gulp`:

> loaderUtils.parseQuery() received a non-string value which can be problematic, see https://github.com/webpack/loader-utils/issues/56
parseQuery() will be replaced with getOptions() in the next major version of loader-utils.

Following the link, it is explained that this is a notice to the webpack-loader developers and **can be ignored by the application developers**.
*(it sure is ugly, though)*

### Warnings on installing module `fsevents`
The following notification may appear:

> npm WARN optional dep failed, continuing fsevents@1.1.2

It turns out that `fsevents` is OS-X specific, and is not required for running on other platforms. If you don't run on a Mac, you can safely ignore this message.


package.json
```diff
@@ -48,21 +51,25 @@
     "babel-preset-es2015": "^6.6.0",
     "babel-runtime": "^6.22.0",
     "babelify": "^7.3.0",
+    "canvas": "^1.6.5",
```
mojoaxel: @wimrijnders What do we need this for? This module [requires Cairo](https://github.com/Automattic/node-canvas#canvas-graphics-api-backed-by-cairo) wich is an unacceptable external dependency. Since this I can not install vis anymore on linux 😭

wimrijnders: `canvas` is used to run unit tests on `Network`. Why is `cairo` an unacceptable external dependency?

mojoaxel: cairo can not be installed via npm and therefor adds a OS-specific dependency. How shozuld the user (like me) know that he has to install g++ and libgif-dev? `npm install` now just fails. I personally think this is not acceptable!
We should not have dependencies other that `npm` (and maybe `git`) as far as I'm concerned.
The least thing would be to document this in the README.

mojoaxel: I've created an issue for that: #3515 Lets discuss there, what the best way is to solve this.

Issue #3245, opened 2017-07-09
Multiple edges between cluster and connected nodes


gubr123:
As @wimrijnders said:

> The cluster edge maintains a direct relation with the one single edge it is replacing. This does not work here because multiple edges need to be replaced with the cluster edge.

Simply, the issue is that when two nodes are connected with more than one edge, the edges remain when clustering, instead of creating a 'cluster edge'.

Here is an example of the issue:

![image](https://user-images.githubusercontent.com/10476634/27992549-b73f2074-649f-11e7-9a5c-6c80e725fce1.png)

Thats I want it to look - (but I want to leave physics enabled):

![image](https://user-images.githubusercontent.com/10476634/27320294-2e25247e-559e-11e7-889f-8757c0dd76de.png)

If I set the physics to false, so all the edges are lapping and it seems like there is one edge between the cluster and the node.

wimrijnders:
Discussion moved from #3187. First post edited with more info from said issue.
Here is [the demo](http://jsfiddle.net/9m433scr/).

----
Additional from @gubr123 previous discussion:

The problem I'm ecountering with clustering is that the nodes are clustered but the edges aren't. So It causes something like this:

![image](https://user-images.githubusercontent.com/10476634/27291299-830738aa-5518-11e7-9874-b03d72146f25.png)

I've seen that disabling physics might help, but I don't want to affect the whole network. Is there any option to replace all the edges in one edge?

wimrijnders:
I've started work on fixing this; I'll continue with this after my final exam on Monday (last minute stress!)

gubr123:
@wimrijnders Good luck!

nanialves:
I want to cluster multiple edges but not with clustered node(s). Is there a solution yet?


I asked [here](https://stackoverflow.com/questions/49626545/cluster-multiple-edges-between-two-nodes) if anyone can help. Thanks.

PR #3268, opened 2017-07-18
Multiple base edges in clustered edge

wimrijnders:
This fix adjusts the clustering edges so that they can refer to multiple edges instead of just one.

Fix for #3245.
   
### Multiple base edges can be assigned to single clustering edge

Basically, the original single value field `clusteringEdgeReplacingId` has been replaced by the array-field `clusteringEdgeReplacingIds`, and all associated operations on it have been adjusted.
    
### Deprecated method `clustering.getBaseEdge()`
This API method is now insufficient, since multiple base edges can be returned.
    
- Added replacing method `clustering.getBaseEdges()`
- Adjusted example `changingClusteredEdgeNodes` for the new method. This is the *only* place where `getBaseEdge()` was used
- Adjusted documentation for new method and deprecation old method.
    
Method `getbaseEdge()` should now be considered `deprecated`, and in due time should be removed.


yotamberk:
@wimrijnders It seems to have conflicts that need to be resolved. 
I have a problem with this PR. I don't want to accept a PR with a deprecation because it means we need to update to a major version. I would leave the method and add a `// TODO: deprecate in 5.0.0` to avoid major release.

wimrijnders:
> I would leave a note to deprecate the method in v5.0.0 in a TODO

That's perfectly reasonable; I'll do that.
TBH I am not aware how major/minor releases are handled; I asked about this in the lobby. But I think this is a perfectly good idea.

wimrijnders:
Conflict fixed

nanialves:
How to use this feature work for [this](https://stackoverflow.com/questions/49626545/cluster-multiple-edges-between-two-nodes)? In other words, this interact like clustered nodes?

Issue #2311, opened 2016-11-14
Node drawn on top of circuit when using hierarchical layout. Gives misleading circuit paths.


ghost:
When using version 3.12.0, I do not have this problem. However, when using version 4.0.0 or higher, a node will overlap onto a circuit path and give the impression that the circuit is going to that overlapping node. See attached screenshots. This first screenshot is when hierarchical layout is set to false.
![newlibhierarchical](https://cloud.githubusercontent.com/assets/19678058/20252165/88e8018e-a9e5-11e6-9786-afe88a56d81c.JPG)

This second screenshot is what happens when the hierarchical layout is set to true.
![newlibhierarchical_1](https://cloud.githubusercontent.com/assets/19678058/20252164/88e78bc8-a9e5-11e6-8616-95ccb6c5b643.JPG)

As you can see, the second and third nodes have edges pointing to each other. But when in hierarchical mode, the third node is drawn to the left of the second node. Also, the edges pointing between the second and third nodes are overlapping each other, which in this case, means that there are three edges overlapping each other.

The latest version that does not appear to have this problem is 3.12.0. After that, the issue exists.
Here is my setup below:

```js
var nodes = new vis.DataSet([
{id: 1, label: 'First'},
{id: 2, label: 'Second'},
{id: 3, label: 'Third'},
{id: 4, label: 'Fourth'},
{id: 5, label: 'Fifth'}
]);

var edges = new vis.DataSet([
{from: 1, to: 2},
{from: 2, to: 3},
{from: 3, to: 2},
{from: 2, to: 4},
{from: 4, to: 5}
]);

var container = document.getElementById("network-area");
var data = {
	nodes: nodes,
  edges: edges
};

var options = {
autoResize: true,
edges:{
arrows: 'to'
},
interaction:{
dragNodes: false
},
layout:{
improvedLayout: true,
hierarchical:{
enabled: true,
direction: 'LR',
sortMethod: 'directed'
}
}
};

var network = new vis.Network(container, data, options);
```

Here is a JSFiddle to see it in action - [LINK](https://jsfiddle.net/redcameron/dwzd6znq/3/)

mojoaxel:
@redcameron I see the problem, but how would you expect the graph to look in a hierarchical layout?

redcameron:
[...]

PR #3294, opened 2017-07-26
Network: Block recalculation of level in LayoutEngine._determineLevelsDirected()

wimrijnders:
**Note:** Small fix, should be easy to review.

Fix for #2311.
    
Nodes with bidirectional edges got their levels shifted due to the handling of both edge directions.
This fix adds a check on bidirectionality and blocks any subsequent level adjustment.
Pure tree layouts are unaffected by this change.


wimrijnders:
:heart: Thanks. Believe it or not, it took me a great while to find a solution for this.

yotamberk:
Yeah... some of the most elegant solutions are the ones that take the longest o find...

Issue #2186, opened 2016-10-19
create webpack example (project)


mojoaxel:
There should be an example or even an example project that shows how to integrate vis.js in a webpack project.

mojoaxel:
Here is a simple example project that uses webpack to import vis.js:

https://github.com/mojoaxel/vis-webpack-demo

wimrijnders:
OK, that's useful as a reference. Would a similar demo be possible for Ionic/Angular? I'm thinking of #3292.

PR #3295, opened 2017-07-26
updated the webpack documentation

mojoaxel:
see also https://github.com/mojoaxel/vis-webpack-demo


SeanSilke:
@mojoaxel  Can we achieve tree shaking with babel-plugin-transform-imports?
https://www.npmjs.com/package/babel-plugin-transform-imports

mojoaxel:
> @mojoaxel Can we achieve tree shaking with [babel-plugin-transform-imports](https://www.npmjs.com/package/babel-plugin-transform-imports)?

I have no idea.
Please experiment with it and document your findings in #2186 or [in the demo project](https://github.com/mojoaxel/vis-webpack-demo/issues).

SeanSilke:
@mojoaxel Firs of all thank you for the work! 
Please add some clarity to my observation.
There is seem to be  significant difference between babel-plugin-transform-imports and what you done.
 babel-plugin-transform-imports just transform input to point exactly to the needed parts. 
What you are doing is rebuilding part of the project.
So tree shaking can't be achieved with transform-imports even if it configuration in right way. Correct?

mojoaxel:
As far as I understand babel-plugin-transform-imports only makes sense if you import vis like this: 

```js
import { DataSet, Network } from 'vis/dist/vis';
```

If you manually defined the parts you want to import the babel-plugin-transform-imports makes no difference:

```js
import { DataSet, Network } from 'vis/index-network';
```

```js
import DataSet from 'vis/lib/DataSet';
import Network from 'vis/lib/network/Network';
```
Here babel-plugin-transform-imports might help because "index-network" also exports unnecessary stuff like moment.js and hammer.js that you might have imported already. Webpack >2 already [supports](https://webpack.js.org/guides/tree-shaking/) unused module export detection so I don't think the plugin is needed (try it!):

```js
import vis from 'vis/index-network';

var nodes = new vis.DataSet(...
var network = new vis.Network(...
```

What you definitive should not do is importing the dist version. Than, not even the babel-plugin-transform-imports can help:

```js
import vis from 'vis/dist/vis';

var nodes = new vis.DataSet(...
var network = new vis.Network(...
```
_For further discussions please use an issue. Thanks!_

Issue #1852, opened 2016-05-13
moveTo documentation and/or implementation is wrong


sgsellmann-miovision:
According to http://visjs.org/docs/graph2d/#Methods the `moveTo()` method accepts a param with the key `animate` to control animation, however in the [code](https://github.com/almende/vis/blob/bc2bc581aaa16905f2ee15b26d3915648c34cc54/lib/timeline/Core.js#L678) on line 18317 the property examined is `animation`.

Tooa:
I suggest to simply change `animate` to `animation` in the documentation. Timeline and Network documentations are correct.

macleodbroad-wf:
This should be labeled as fixed awaiting release I think

PR #3302, opened 2017-07-27
#1852 - Corrects documentation

macleodbroad-wf:
R: @wimrijnders


wimrijnders:
It tool me too long to figure out this `moveTo()` method. Turns out that `timeline` and `graph2d` inherit from the same class `Core`, so that for both components the `moveTo` is in fact identical.

And you said you had 'relatively little experience with javascript'? Well, it that's true you have plenty of other skills to compensate for that.

Issue #2839, opened 2017-03-07
DataSet.get([undefined]) returns [null], should return [] instead


Head:
The documentations says:
>When no item is found, null is returned when a single item was requested, and and empty Array is returned in case of multiple id's.

But that's not the case, reading the code at https://github.com/almende/vis/blob/master/lib/DataSet.js#L373 it pushes item to the array no matter what is found.

wimrijnders:
Ah, right. I see what you mean, when id's are not found `null`'s are returned in the Array.

I'll spend some time on this. The fix for this particular issue is easy of course, but I also want to add unit tests as well to catch this and as many similar edge cases as possible.

wimrijnders:
The documentation issue has been resolved by adjusting it to fit the current working.

PR #3303, opened 2017-07-27
#2839 - Corrects documentation for DataSet get and how multiple missi…

macleodbroad-wf:
…ng ids are handled.


macleodbroad-wf:
R: @wimrijnders @mojoaxel

macleodbroad-wf:
Originally I had written a `fix`, that excluded unfound entries, but I would consider that a backwards breaking change, and after some reflection, I'm not sure it is a better API.

I prefer APIs with a symmetrical input and output for cases like this.

For example:
```
input: [nodeId1, missingNodeId, nodeId2]
output: [<Instance of Node1>, null, <instance of Node2>]
```
 if I pass in a list of 3 arguments, I expect a list of length 3 back, with nulls for anything that was missing, and I expect those results in the order that I had passed in the arguments. 

To me, this better follows the [Principle of Least Astonishment](https://en.wikipedia.org/wiki/Principle_of_least_astonishment).

Also, it is already implemented in the way I describe, and to change it now would be surprising to consumers, who would now have to cycle through the result list, checking node ids, to understand which nodes were found and which were not.

macleodbroad-wf:
Should this also be opened against develop?

wimrijnders:
Wrt. nulls in returned array: I understand your reasoning perfectly, and the argument for not changing the output is good. However, I personally would be least astonished if `null`'s were *not* returned.

OK, how about this idea: add an option for returning `null`'s or not. The default would be to do what it does now, thereby breaking no-one's code. But removal of the `null`'s can be  set in options.

> Should this also be opened against develop?

Aargh! I missed that. Yes, `develop`.

PR #3304, opened 2017-07-28
Makes Network, DataSet and DataView eslint compliant

macleodbroad-wf:
R: @wimrijnders @mojoaxel 

Rather than follow a dogmatic approach, I did break some eslint rules. 

I believe that even though some variables are unused, they are well placed to communicate the intent and expected usage of the function.

### DIFF
--- a/lib/DataSet.js
+++ b/lib/DataSet.js
@@ -321,7 +321,7 @@ DataSet.prototype.update = function (data, senderId) {
  * {String | function} [order]  Order the items by a field name or custom sort function.
  * @throws Error
  */
-DataSet.prototype.get = function (args) {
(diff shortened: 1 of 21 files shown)


wimrijnders:
> Rather than follow a dogmatic approach, I did break some eslint rules....

I see it, unused parameters for certain methods. Tagging them with `eslint-disable` is a valid solution here.

wimrijnders:
Your linting fixes are all good; however, when I run linting, I also get errors on the following files:

```
lib/network/modules/components/nodes/shapes/Circle.js
lib/network/modules/components/nodes/shapes/CircularImage.js
lib/network/modules/components/nodes/shapes/Dot.js
lib/network/modules/components/nodes/util/CircleImageBase.js
```

Did you skip these intentionally? Or didn't they show up during linting?

Issue #3246, opened 2017-07-10
showMajorLabels to false and scale set to 'week' creates vertical lines display bug


Franzyval:
Hello and first, thanks for this really useful tool :),

I think i have identified a bug: when setting both options showMajorLabels to false and timeAxis to  {  	scale: 'week',   	step: 1    },  there is a display bug for vertical lines where months were supposed to be. this seems to just be a display bug.

on version 4.20.0 of the Timeline

Here you can see that the div for week 22 appears two times
![bug-timeline](https://user-images.githubusercontent.com/26770818/28010174-4e75ac74-655e-11e7-9bb1-3d5b7a7bc48a.png)

here is sample code to reproduce : 
var container = document.getElementById('example-timeline');

// Create a DataSet with data 
var data = new vis.DataSet([{
    id: 1,
    content: 'First event',
    start: '2014-08-01'
}, {
    id: 2,
    content: 'Pi and Mash',
    start: '2014-08-08'
}, {
    id: 3,
    content: 'Wikimania',
    start: '2014-08-08',
    end: '2014-08-10'
}, {
    id: 4,
    content: 'Something else',
    start: '2014-08-20'
}, {
    id: 5,
    content: 'Summer bank holiday',
    start: '2014-08-25'
}]);

// Configuration for the Timeline as JSON object
var options = {
    showMajorLabels: false,
    timeAxis: {
    	scale: 'week',
    	step: 1
    },
    width: '100%',
    locale:'fr',
    editable: true,
    margin: {
        item: 20
    }
};

// Create a Timeline
var timeline = new vis.Timeline(container, data, options);

Thanks !
Regards

bradh:
Can you try with 4.20.1? If it still happens, can you create a jsbin or jsfiddle that shows the problem?

Franzyval:
Hi, thanks, for your interest in this problem, 

https://jsfiddle.net/e7uodse2/1269/
this jsFiddle is using vis 4.20.1, error still occurs. Look for week 27 or week 40

marcortw:
Nice to see that people are using the week scale feature! I only saw this issue now and I am not sure if there's a misunderstanding here. The additional vertical line at the month's start is a feature, not a bug. See my comment in https://github.com/almende/vis/pull/3009#issuecomment-297300219 . @yotamberk, do you remember this discussion?

Additionally, this fix seems to break the weekStyling.html example:

[...]

PR #3305, opened 2017-07-28
#3246 - fix issue when showMajorLabels == false is used with a weekly scale.

ponml:
The referencing issue's example (#3246) should be sufficient to see the fix. This is my very first PR to this repo, and I'm hoping I've followed all the guidelines properly, so please let me know if not.

From what I can gather, there is some special logic to handle the creation of TimeSteps when a week crosses over a month (which is the `Major` unit for weekly mode).

This logic needs to be ignored if `showMajorLabels` is set to false.

-Chris.


yotamberk:
Awesome! Great job!

yotamberk:
@ponml can you please address the issue occurring with this PR: https://github.com/almende/vis/issues/3246#issuecomment-321995229

ponml:
@yotamberk I'm a bit confused on the exact issue happening. Is it a matter of reverting my change, as the extra step is an intended feature? Or rather, is it that there seems to be something wrong in the labeling in the weekStyling.html example since merging this PR?

yotamberk:
The example seemed to have broke. This means that the application of your fix broke something else. I agree with this implementation of the bug fix, but this causes another bug to pop out

ponml:
Alright, I can take a look 👍

ponml:
#3361 -> fix for the minor labelling

PR #3309, opened 2017-07-29
lint shared

macleodbroad-wf:
R: @wimrijnders @mojoaxel

### DIFF
--- a/lib/shared/ColorPicker.js
+++ b/lib/shared/ColorPicker.js
@@ -15,8 +15,8 @@ class ColorPicker {
     this.applied = false;
 
     // bound by
-    this.updateCallback = () => {};
-    this.closeCallback = () => {};
+    this.updateCallback = () => {}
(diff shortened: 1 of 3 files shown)


lib/shared/ColorPicker.js
```diff
@@ -15,8 +15,8 @@ class ColorPicker {
     this.applied = false;
 
     // bound by
-    this.updateCallback = () => {};
-    this.closeCallback = () => {};
+    this.updateCallback = () => {}
```
wimrijnders: I'm not too happy with these two; they're assignments of functions and usually these would be terminated with a `;`. This looks like an omission in `eslint`.

macleodbroad-wf:
No. It was my IDE that suggested those changes that you wanted rolled back, not eslint.

wimrijnders:
Ah, ok. But I'll assume you also did the tests with `eslint`. So all is well.

PR #3315, opened 2017-07-29
Fixes #2839 - Re-words documentation to reflect symmetrical input/out…

### LINKED ISSUE #2839: DataSet.get([undefined]) returns [null], should return [] instead

macleodbroad-wf:
…put of get() when passed multiple ids.

R: @wimrijnders @mojoaxel

### DIFF
--- a/docs/data/dataset.html
+++ b/docs/data/dataset.html
@@ -312,7 +312,7 @@ var data = new vis.DataSet([data] [, options])
       <td>Object | Array</td>
       <td>
         Get a single item, multiple items, or all items from the DataSet.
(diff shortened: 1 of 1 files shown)


wimrijnders:
😞 I still don't like it. I'll let another reviewer make the call here and go along with that.

macleodbroad-wf:
Need another reviewer to wrestle with @wimrijnders and I. Maybe @yotamberk ?

wimrijnders:
OK, so I'm going along wih @yotamberk's resolution.

Issue #3316, opened 2017-07-30
Documentation issue for joinCondition + clusterByConnection


CiaranG:
When clustering network nodes, the various cluster functions all take an options object which includes a `joinCondition` callback. This is documented (`docs/network`) as receiving a single `nodeOptions` parameter. That's true for most of the cluster functions, but not for `clusterByConnection` where it receives two - one for the parent node and one for the actual candidate (child) node.

wimrijnders:
Confirmed. Well spotted. We'll put someone on this to fix it.

PR #3319, opened 2017-07-30
Fixes Issue #3316 - Updates network documentation to account for edge…

macleodbroad-wf:
R: @wimrijnders @mojoaxel


macleodbroad-wf:
addressed feedback from @wimrijnders

Issue #2546, opened 2017-01-05
Automate release process


mojoaxel:
**We really need to do something about the release process of this library!**
There is way too much manual work required to create a release. That is why canches bulk up and then releases even get harder to do.

Let's think about what we can improve..

### Automatic test (travis-ci)
* Did any of the existing parameters/options change => Breaking change. This could be done checking e.g. the `optionsTimeline.js` for deletes or changes (additions are ok).
* checkout if the examples are still running after `npm install`
* checkout if the examples are still running after `bower install`
* Do the examples still run without any error on the console (e.g. [phantom.js can monitor console outputs](http://phantomjs.org/api/webpage/handler/on-console-message.html)).
* Do (reproducable) examples still produce the same screen-output (using e.g. [wraith](https://github.com/BBC-News/wraith/)

## Automate the release
* Get the commit-history since the last tag from git and append it to a changelog.
* Auto-bump up the version using e.g. [semantic-release](https://github.com/semantic-release/semantic-release). This would require a strict [guideline for commit messages](https://docs.google.com/document/d/1QrDFcIiPjSLDn3EL15IJygNPiHORgU1_OOAqWjiDU5Y/edit)!
* Create a release on github using the changelog and the new version.

### Update the gh-pages branch after each new release:
* copy the `dist` folder from the recent `master` branch.
* copy the `docs` folder from the recent `master` branch. (In future maybe even generate the docs from source comments!?)
* copy the `examples` folder from the recent `master` branch.
* generate the example-overviews using all existing examples and images. For examples that not yet have a preview-image a default.image should be used, that can be replaced later on.
* Update the library version number in the index.html page.
* Update the CDN links at the download section of index.html AND the CDN link at the top. (replace all)

yotamberk:
I'll take care of pluging in Travis CI

Tooa:
We could automate the changelog generation process similar to the angularjs team. They have restrictions for their commit messages ([reference](https://github.com/angular/angular.js/blob/master/CONTRIBUTING.md#-git-commit-guidelines)). You can then check whether the commit message meets the criteria with Travis by running a script like [this](https://github.com/JensRantil/angular.js/blob/ffe93bb368037049820ac05ef62f8cc7ed379d98/test-commit-msgs.sh). Even better if we use a git commit hook that checks the message before the commit. 

This allows us to generate the changelog automatically. See [here](https://github.com/angular/material/blob/master/release.js#L94). What do you think?

mojoaxel:
[...]

PR #3329, opened 2017-08-02
[Proof of Concept] [Experimental] Adds automated visual regression testing using phantomjs and image ma…

macleodbroad-wf:
…gick

Based on conversation on https://github.com/almende/vis/issues/2546 I've started this as a proof of concept for visual regression testing the static examples.

### What this currently does

1. Opens some examples on the visjs.org website and renders them as `png` files to the respective example directory.
2. Opens some examples on the `gh-pages` branch in git, and renders them as `png` files to the respective example directory.

screen shot 2017-08-01 at 8 14 07 pm

3. Does an image compare of one live example against one example from the `gh-pages` branch

I went with PhantomJS and ImageMagick, simply because they're already available in the standard Travis environment.

### To test this branch
- Install phantomjs
- Install ImageMagick
- Pull down this branch
- From the the vis directory:
    - run `phantomjs examplesToPng.js`
    - run `./magick.sh`
- <strong>Note the `-live.png` and `-local.png` examples in `examples/graph2/` </strong>
- <strong>Note that the result code on magick.sh will be 0 if the images match. </strong>

### Long-term intent
- Automate visual regression testing of all static examples
- Include all examples by default, with a way to opt-out
- Have visual regression tests on every PR, comparing the PR'd branch against the live page
- Fail the build if visual regression tests turn up bugs

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,7 +12,6 @@ addons:
       - g++-4.8
 before_script:
   - npm run lint
-  - sudo apt-get install imagemagick
   - phantomjs examplesToPng.js
   - npm install gulp
 script:
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,7 +12,9 @@ addons:
       - g++-4.8
 before_script:
 #  - npm run lint
+  - phantomjs examplesToPng.js
   - npm install gulp
 script:
   - gulp
   - npm test
+  - ./magick.sh
--- a/examplesToPng.js
+++ b/examplesToPng.js
@@ -35,11 +35,10 @@ function waitFor(testFx, onReady, timeOutMillis) {
         }, 250); //< repeat check every 20ms
 }
 
-var examplePaths = ["examples/graph2d/01_basic.html", 
+var examplePaths = ["examples/graph2d/01_basic.html",
 					"examples/graph2d/02_bars.html",
 					"examples/graph2d/03_groups.html",
-					"examples/graph2d/04_rightAxis.html",
-					];
+					"examples/graph2d/04_rightAxis.html"];
 var tasks = examplePaths.length*2;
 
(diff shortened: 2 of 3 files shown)


macleodbroad-wf:
Here's the Travis Build Log, showing the phantomJs generating screenshots: 

screen shot 2017-08-01 at 8 24 53 pm

macleodbroad-wf:
Sadly this failed because it can't find magick. I'm not sure why that is. Maybe it has something to do with the builder image?

screen shot 2017-08-01 at 8 49 54 pm

yotamberk:
@macleodbroad-wf  REALLY AWESOME! I like this direction. If you can make it work and pass the test, that would be perfect.

macleodbroad-wf:
Although this appears to work, it's not as reliable as it needs to be.

The main reason it's not reliable enough right now, is because there is no sufficient callback mechanism that guarantees the graph is fully rendered.

This is an instance in which the lifecycle callbacks as described in https://github.com/almende/vis/issues/3352 would come in handy.

mojoaxel:
I also worked on this quite a lot and I'm happy that some else is going in this direction.
I don't think that taking the screenshots and comparing therm should be done manually like in @macleodbroad-wf approach. There are [multiple Visual Regression Frameworks](https://github.com/mojoaxel/awesome-regression-testing) out there and we should try to use one of them.

My current experiments are in the [gemini branch](https://github.com/almende/vis/tree/gemini), but I was not satisfied with the integration, yet.

mojoaxel:
For further discussion please use #3355

macleodbroad-wf:
Sounds good @mojoaxel, I'll await your progress. If there's anything I can do in this area, let me know.

Issue #1315, opened 2015-09-21
Network: Cannot uncluster nodes when removing a node that has an edge to that cluster


dragonzone:
Here is the example: http://jsfiddle.net/oj61w366/

AlexDM0:
Hi,

This is all part of the same shortcoming. It is not possible to use dynamic data around clusters. Once this is addressed, all of these issues will be resolved.

Thanks for reporting.

Regards

Tooa:
"inc" for this feature :+1:

wimrijnders:
FYI I fixed this. However, it's part of a much larger code addition which completely handles the clustering updating due to changing data. It's going to take a while to even it out.

Issue #1291, opened 2015-09-11
Network: Removing node inside a cluster and then opening cluster throws error


dragonzone:
Here is the example: http://jsfiddle.net/umecqna8/

AlexDM0:
Errr... yes. At the moment, dynamic data and clusters don't play nice together. I'll update this in future versions. This is also the reason your other issue does not work at the moment.

Thanks for the heads-up!

Regards

dragonzone:
Which one, I forgot. Can you refresh me?

AlexDM0:
#1218

dragonzone:
The power is really being able to change the data on the fly. Not static data, that's old school. B)

AlexDM0:
Yes, but the book-keeping is not up to speed yet. This will have to be implemented nicely and recursively.

hengchu:
Hi @AlexDM0, for this issue and #1218, I'd like to help fix them if you haven't had the chance to work on them yet. Would you mind letting me know which part of the code base needs to be modified to fix them? Thanks! BTW this is a great library :)

AlexDM0:
Hi,

Its all in the clustering file. It needs to listen to the dataChanged event and cleanly and recursively update the bookkeeping. You'll also need to fix #1315 

Regards

AoDev:
@mojoaxel I was checking the code regarding this today and I actually made the clustering work and support node removal while clustered. But as I read the code, I got many questions and I wonder if it would be possible to speak with someone about it.

wimrijnders:
> But as I read the code, I got many questions and I wonder if it would be possible to speak with someone about it.

@AoDev I only noticed this statement just now. Are you still looking for someone to speak with? I am hereby available.

AoDev:
@wimrijnders Thank you. But since February I am not working with vis. I can not work on it at the moment.

wimrijnders:
@AoDev Yeah, well it *was* a long time ago, from before I became admin. Well, if you ever want to revisit it, notify me.

Issue #1218, opened 2015-08-24
Networks: Phantom link after creating cluster and creating link to node in cluster


dragonzone:
Here is the example:

http://jsbin.com/rafari/16/edit

AlexDM0:
Hi Dragon,

I'll see if we can fix this before the upcoming release. It should have been today, but it will probably be tormorrow.

Regards

dragonzone:
Thx. Will wait for release.

dragonzone:
BTW, I notice when I use jsbin to create the sample file, if I point the script file to: https://raw.githubusercontent.com/almende/vis/master/dist/vis.js, it will not load/run. I have to point to an CDN for it to load. Do you know why?

AlexDM0:
It's in the MIME settings, you'll have to pass it through https://rawgit.com/

On the other hand, CDN is better because your issue is for a specific version of vis, if we release, your jsbin will be invalid.

Regards

AlexDM0:
I think that for this issue, you'll have to decluster and cluster again.. I'm not sure if this will be easily fixed before the release. The easiest fix I can think of is not allowing edges to hidden nodes be shown, on the other hand, this may interfere with people using that as a cheat to redirect edges.

dragonzone:
it's a workaround. but with simple logic in the example, it should work. It may be confusing to other users when they use it.

AlexDM0:
I'll have to think of a nice way to do this, the edges should not need to know about the clustering to keep the modularity of the network. I'll look into it.

dragonzone:
Let me know. Because if it gets to buggy, I may have to write my own way of clustering. Holding my separate list.

AlexDM0:
I have reworked the clustering in the next release, greatly speeding it up for stacked clusters. The edge problems we had before are also solved, but this is a different case. There is no catch for making an edge connected to a node that is in a cluster as it stands now.

dragonzone:
You have an excellent tool! I'm just trying to help it improve it more since it's the "de facto" here and it's now, I think, the main attraction in our organization and is spreading wildly to others org. Just for a heads up, clustering is becoming a more demanding piece here, where I will be concentrating more. So I may find a few more issues.

AlexDM0:
I hope you do (in the sense that its difficult to test hehe)!

Glad you guys are happy with it :)

The next release should have great performance gains in the physics during initialization and stabilization all around.

dragonzone:
[...]

Issue #3025, opened 2017-05-02
Error with adding a node to the dataset after setOptions


AlexDM0:
You can see the issue here:
https://jsbin.com/hesexidaxi/edit?html,console,output

rasmusblockzero:
To me it looks like the main oddity is this.nodeOptions = options; in nodesHandler.setOptions.
It's set to incoming options without any parsing, instead of reflecting nodeOptions it becomes a snapshot what ever was the latest option to be set. It beats me why this is important to pass to new nodes but I also must say that there is a fair bit in the options complex that beats me.

In either case when this becomes undefined this cause a raise condition in the labelHandler

It was added fairly recently maybe @eymiha or @mojoaxel has some input there as they added the parameter ?

wimrijnders:
Analysis of @rasmusblockzero is valid. The node options get reset completely when `NodesHandles.setOptions()` is called.

This also happens in:
* EdgesHandler

This does **not** happen in:

* Canvas
* CanvasRenderer
* Groups
* InteractionHandler
* KamadaKawai
* LayoutEngine
* ManipulationSystem
* NavigationHandler
* PhysicsEngine
* SelectionHandler

...in other words, in all the other modules with options.

So there is a discrepancy between the handling of options for Nodes & Edges, and the rest of the code.

I admit this is confusing; while it is understandable that sometimes options need to get initialized from scratch, it is not possible to update just a subset of the options. IMHO this would be a useful feature. Will think about this.

wimrijnders:
As a workaround, the easiest way to avoid the error is to set `options.nodes.font.multi: false`. This will skip the code in which the error occurs (`Label.propagateFonts()`).

wimrijnders:
Added PR #3025 to fix the occurrence of the reported error. The example will now work as expected.

It should be noted that this is a *partial fix*; the issue with the resetting of the Nodes & Edges options is not addressed.

AlexDM0:
Thanks for the quick fix! Should an issue be made for that?

wimrijnders:
Reopening because the fix was partial; all it did was hide the error. This issue still needs some work.

AlexDM0:
[...]

PR #3330, opened 2017-08-02
[Network] Dynamically adjust clustering when data changes.

wimrijnders:
Fix for #1218, #1291, #1315.

Changes in data are now detected and clustering is changed accordingly if required.
Unit tests have been added to check correct working of fixes for issues.

### Added method `updateState()` to module `edgesHandler`
Called in handler of `_dataChanged`.

- Scan for deleted nodes, remove regular edges accordingly in the active edge state.

'Orphaned' edges were never deleted; this was not apparent on-screen, because of a check for valid connection points of edges during visibility determination.
This could have lead to memory leaks in the case of many node deletions.

### Added method `updateState()` to module `Clustering`
Called in handler of `_dataUpdated`.

- Scan for deleted non-clustered nodes and edges and updates the clustering state accordingly.

Also checks the internal state of regular nodes and edges. If some
setting, which most likely have been changed during the data change, conflicts with the 
current clustering state, the clustering state is adjusted.

### Added header comment in `Clustering.js`
The clustering is non-trivial and complex. I added a description of the state which is
relevant for clustering, for future reference.

## Note
While this fixes the issues mentioned, the dynamic update of clustering is *not complete*.
I added TODO's at the top of `Clustering.js` to record these.

The current PR is big enough already and is enough to fix the mentioned issues, so these TODO's will
be addressed in a separate PR.


wimrijnders:
Hooray! That was a difficult birth.

justinharrell:
So this is causing an issue for me but I want to confirm if this is intended behavior or not. We have a concept of collapsing a node, but clustering behavior was not what we wanted. The collapsed node shouldn't change.

To do this we just delete all "outlier" nodes (nodes connected to the node being collapsed that have no other connections) from the dataset. We then store these "hidden" nodes elsewhere. Later if you expand we add them back to the dataset. In the dev branch now the edges get deleted, so when the nodes are re added they are no longer connected, whereas before this commit they where just invisible. Whats more confusing is they are still in the edge dataset, so now the edges are out of sync with the dataset.

Currently to fix this I just add a dataSets.edges.update(dataSets.edges.get()); to resync. 

This seems a little surprising, should the edges be removed from the dataset too? Should changing the node Dataset resync the edge data set? Maybe the nodes shouldn't be deleted after all? Am I just doing it wrong?

Repro code:

`    var node2 = {id:2, label:"test2"};
    var nodes = new vis.DataSet([{id:1 ,label:"test1"}, node2]);
    var edges = new vis.DataSet([{from: 1, to:2}]);

    // create a network
    var data = {
        nodes: nodes,
        edges: edges
    };    

    var network = new vis.Network(container, data, options);

    nodes.remove(node2.id);
    nodes.add(node2);

    console.log(edges.get());`

wimrijnders:
**TL;DR** Yes, this is the intended behavior.

@justinharrell let's see how I can clarify this.

First, as of this and related PR's, the edges are indeed removed from the internal state for dynamic updates. Formerly, 'phantom' edges were retained in the state and, by chance, were not handled in the regular flow.  This led to all kinds of sorrow, notably memory leaks.

From my perspective, this was a serious flaw, needing urgent repair. But I can understand the current confusion on your side.

Next, the internal node+edge data in `Network` is completely separate from the data in the `DataSet` instances for the nodes and edges. The rule abided to is that `DataSet` content must not be changed in any way. This content is in the ownership of the user.

Bottom line: managing the `DataSet`'s is the user's responsibility. And if you need to reset edges, you will need to do this explicitly.

> Currently to fix this I just add a `dataSets.edges.update(dataSets.edges.get());` to resync.

Indeed, something like this. You could optimize that by doing the update only for the nodes that are reset.

----
> We have a concept of collapsing a node, but clustering behavior was not what we wanted. The collapsed node shouldn't change.

Change in what way? The nodes within the clusters are completely retained and will pop up as defined - the coordinates are likely to change, however.
Do you perhaps mean that the cluster node looks different? If so, I think that that can be solved.

justinharrell:
Since the fix on my side is pretty easy, this isn't a huge deal, but I do worry this really goes against principle of least surprise and may cause others to be confused in the future. I remember writing the code and wondering what happened to edges that connected to nothing and being glad it did the obvious which is simply not draw them, but as soon as I added a node to connect them they started drawing. Now they won't appear until they are updated again, even though nothing has changed with the edge.

Is there precedent for a dataset to be out of sync with the graph? I haven't used the manipulation system, but looking at it, it makes changes to the datasets. That is if you delete a node through the manipulation function it removes it from the dataset. It seems as though the intention is the dataset is "the truth" and should never be out of sync with whats in the graph.

I am not sure of the details of the memory leaks, but I don't think invisible edges should be considered a leak, they are in my dataset after all. In the end its my responsibility to remove them if no longer valid and wasting resources. 

I would suggest at least resyncing the edges dataset internally if the nodes dataset changes to prevent the user from having to do so, but this seems heavier than what it was doing before based on my understanding of the code.

>Do you perhaps mean that the cluster node looks different? If so, I think that that can be solved.

Yes from looking at it originally, it seemed like overkill and much more complicated than what I did which was just remove outliers and put them back. The node being collapsed doesn't need to change, and I didn't want to deal with everything being different because it was a clustered node rather than the original node.

Either way though that not really the issue I imagine other scenarios where this will become an issue, it seems now the order of operations is very important, you must always add nodes first then edges second or the edges will not display yet they will be in the dataset.

```
    var node1 = {id:1 ,label:"test1"};
    var node2 = {id:2, label:"test2"};
    var edge = {from: 1, to:2};
    
    var nodes = new vis.DataSet([]);
    var edges = new vis.DataSet([]);

    // create a network
    var data = {
        nodes: nodes,
        edges: edges
    };    

    var network = new vis.Network(container, data, options);

    edges.add(edge);

    nodes.add(node1);
    nodes.add(node2);
   
    //wheres my edge?

```

wimrijnders:
OK, I don't see the point in doing a step-by-step rebuttal of all the above, for a simple reason: I want `Network` to be eminently usable for all.

I pondered about your comment for a while and came to the conclusion that the reason `Network` works like this is because it was a guiding principle during construction. I even searched and found a [relevant comment](https://github.com/almende/vis/issues/3025#issuecomment-303029108) on this.

And the reason why it still works like this is out of a respect of the original coders. When I think about that, I don't find it a very good reason (no offense meant, @AlexDM0 and others). Not because it was a bad design decision but because `Network` should be allowed to evolve.

So, I'm perfectly prepared to consider a 'bidirectional' connection, actually. If it can be fitted in without breaking current working, I don't see a reason not to add it. I would think there are plenty of gotcha's to consider, though.

For me, this is a long-term implementation thing. I'm currently on a self-imposed break till the next release. But feel free to put some though and/or effort in it.

justinharrell:
Sorry I am not trying to be argumentative, just worried this may cause subtle issues for existing code and it at least may need to be documented better. Specifically you should not add edges before nodes now, and if you you do then you must update all effected edges even if their data has not changed. 

>So, I'm perfectly prepared to consider a 'bidirectional' connection, actually. If it can be fitted in without breaking current working, I don't see a reason not to add it. I would think there are plenty of gotcha's to consider, though.

I think you are misinterpreting my proposal. I would NOT want a bidirectional connection here, I did not remove an edge, I only removed a node. It is very good the change did not also remove the edge from the dataset, that would be very bad. The problem is though now you have edges in the dataset that are not actually in the graph, the dataset is now out of sync, I don't believe there is another situation that should cause this. 

I did find a post where JOs and Alex discissed this and fixed issues related to it not working  the way I am proposing: https://github.com/almende/vis/issues/235#issuecomment-51442790

wimrijnders:
And I'm trying to avoid the argument, if any. I realize I went on a tangent. I will also admit here that I have a huge resistance to the idea of having an incomplete state with `Network`.

Nevertheless, I understand your point of view. The state of `Network` should correspond with whatever is in the datasets. I realize now that this not the same thing as keeping in unconnected edges within the network state, so I propose the following:

> If a node is added, a check is done in the edges `DataSet` to see if there are any edges connecting to the new node. If so, add these edges to state.

Would this work for you? As far as I can tell, the current addition/deletion of edges and the deletion of nodes can be retained as is. If you see any pitfalls, please let me know.

----
Also, even if you didn't mean it, I think the bidirectional thing is a good idea; there are plenty of issues related to the idea that the datasets should be updated from the network. In due time, I will spend time on this.

justinharrell:
Yes that would work. It just takes care of something I would have to do manually that wasn't necessary before and could cause possible issues in existing code. 

The internal state is just a implementation detail whether unconnected edges are not drawn or deleted completely from the internal state doesn't matter, but the that the dataset state is not honored is the issue.

I agree that checking added nodes and re-adding connected edges seems to be the best approach based on how the internal state works now. I could try and do a PR but it will take me a littel while to get a grip on the best place to do it, I think _updateState in the EdgeHandler?

The bidirectional update can be tricky, I think I prefer the way things are done now where the dataset aren't manipulated by the network (except for the manipulation module). Perhaps the storePositions method could be expanded, since it basically reverse syncs the datasets from the network but the developer is under control when/if that occurs.

wimrijnders:
> ...but that the dataset state is not honored is the issue.

Yes, I understand that now and I'm inclined to agree.

> I could try and do a PR but it will take me a little while to get a grip on the best place to do it, I think _updateState in the EdgeHandler?

Yes, that's the best place. Spot on.

> The bidirectional update can be tricky, I think I prefer the way things are done now where the dataset aren't manipulated by the network

It will definitely be tricky. This is a long-term thing which needs a lot of thinking about, it won't be implemented any time soon. So just consider it 'vapor-ware'.

> (except for the manipulation module).

Uhm, I don't really think that the manipulation module changes the datasets. I actually checked this with the examples. On what do you base your observationi?

> Perhaps the storePositions method could be expanded,...

An other option I'm considering is a new method like `Network#saveState()`, which creates a json with all the state needed to recreate the network later in another instance. This will actually solve a load of issues, e.g. #3366.

In addition, this will make it possible to prepare a network server-side, e.g. #2255. This would be a good way to deal with huge network, which may take ages to stabilize (e.g. #2741).

wimrijnders:
@justinharrell Irrespective of who is going to do the fix, would you mind making a new issue on checking existing edges when adding nodes? So we can move the discussion out of this PR.

justinharrell:
>Uhm, I don't really think that the manipulation module changes the datasets. I actually checked this with the examples. On what do you base your observationi?

In ManipulationSystem.js the default behavior seems to be to change the dataset, interestingly the in the issue comment you reference further up Alex confirms this: https://github.com/almende/vis/issues/1490#issuecomment-163263790

For example: https://github.com/almende/vis/blob/bc6330d612ef1b5e13156eee4ef8b129d01013f6/lib/network/modules/ManipulationSystem.js#L463

Unless there is something else going on, I haven't actually tried the manipulation system myself.

>An other option I'm considering is a new method like Network#saveState(), which creates a json with all the state needed to recreate the network later in another instance.

This would be very useful for all the reasons you mention, since one of my interest is exporting the network to SVG #723 which I currently have working by replacing the current canvas context for one frame to get and exact snapshot of network state, it might make more sense to load into a second network with say a different renderer if the full state could be preserved.

>Irrespective of who is going to do the fix, would you mind making a new issue on checking existing edges when adding nodes? 

Will do

wimrijnders:
Well, crap. I open `ManipulationSystem` and I see the calls staring me in the face.
I understand why I didn't see the changes in example `manipulation`: `getscaleFreeNetwork()` returns the edge/node data as array's, so the DataSets are internal to `Network`. Okay, TIL.

I really find this behavior to be inconsistent. It's the *only* place in the entire code base where it happens. I'm just going to ignore that for now.

Issue #2123, opened 2016-09-28
iOS pinch zoom no longer works


MiikaL:
Hi there,

Previous to the iOS 10 update, it was possible to use "pinch zoom" on the visjs Timeline as long as the page had the meta tag <meta name="viewport" content="width=device-width, initial-scale=1.0, maximum-scale=1" />. The newest version of iOS Safari no longer respects this meta tag, which has consequently also broken the pinch zoom functionality of the Timeline. Is there a solution for this problem or should I dig deeper?

Ideally, pinch zoom on Timeline AND allowing the user to zoom the page itself would be ideal, for accessibility reasons.

MiikaL:
The 4.17 update made the zooming work kind of, but it seems like it is zooming the timeline AND the page at the same time. In other words, is there a missing preventDefault/stopPropagation/return false in there somewhere which would make the timeline fully capture the zoom event if it starts within the confines of the timeline?

mojoaxel:
@MiikaL Thanks for reporting this. Can you recommend an example to test this? I have access to an iPadMini.

MiikaL:
The issue can be seen on http://visjs.org/examples/timeline/interaction/limitMoveAndZoom.html . 

It is a little hard to explain what is happening, as it appears to be behaving a little inconsistently. If you pinch/extend your fingers horizontally, it appears to work (more often than not), as long as you are very careful to stay within the bounds of the timeline element. pinch/extend vertically appears to fail more often, causing the zoom to impact the page rather than the timeline element.

I am testing this on my iPhone, so the screen is small, making it harder to keep my fingers within the timeline element.

Possibly also the limits set by that example are further complicating the issue, almost as if once the limit has been reached then it stops catching the pinch events and lets it bubble up to the page.


TLDR: Pinch/extend repeatedly on example timeline seems to -sometimes- cause the action to be caught by the page rather than the timeline.

yotamberk:
@MiikaL The problem I see happening is that the pinch/extend stop one the timeline gets to it's maximum/minimum zoom. It has nothing to do with the screen size, but more of what's the zoom level. The event returns and stops the pinch/extend event once reaching on of these min/max zooms making it feel like it's stuck. 
This can be fixed quite easily. I'll try getting to it the next few days.

wansco:
[...]

PR #3334, opened 2017-08-03
Disable the default handling of the pinch event when handling it.

darthsteven:
This should address some of the issues in #2123


yotamberk:
Awesome! Thanks!

PR #3342, opened 2017-08-04
Bug fix for null parent

danny-larsen:
This fixes an error on updating items in nested groups


yotamberk:
@danny-larsen can you reference the issue this solves?

danny-larsen:
No sorry. I didn't have time to make an issue. But it is just one line and it seems like an error to me that other places in that function the parent is checked for null before accessing it. Maybe @yotamberk can help. I can see that he added it?

yotamberk:
Funny, I am @yotamberk :laughing:

danny-larsen:
Ha. I guess I was a bit quick there :)

Issue #3276, opened 2017-07-20
Timeline: Performance on first render is extremely slow


levzem:
Hi,

I am working with timeline (4.20) in chrome 59 on windows 7. I am working locally though, and not on an actual website. 

I am working with extremely large datasets and large amounts of groups (~40 groups with an additional nested group for each and 30K - 80K items). 

It is my understanding that the timeline only loads the visible items to enhance performance. I set limits on the start and end date as well as the max zoom level. I am happy with the scrolling/moving performance, however, when I first load the timeline. it takes an extremely long time to actually first generate the timeline (5 minutes for 30K items, 15+ minutes for 80K items) even when the visible items on the timeline are incredibly low (4 items). 

Your examples load quite quickly for 100K items (~10 seconds) and a simple JSBin program from #3248 does it in about ~20 seconds. Attached for reference -> https://jsbin.com/jupelipala/edit?html,js,output

Is there something I can do to improve my own performance or is this an issue that the timeline initially loads all items regardless of visibility?

Thanks

bradh:
I'm not sure I understand the difference between your case and the performance test examples. Can you try to explain that a bit more?

levzem:
The problem is that the initial loading/generation of the timeline is slow (performance after the fact is satisfactory). 

If I run the same exact code from JS Bin locally from my machine, it takes an extraordinary amount of time, but on JS Bin, it runs pretty quickly.  If I run the example code from the visjs.org locally even with many groups (40), it is pretty quick. Likewise, my code renders the timeline slowly as well. I am not sure where this difference comes from.

bradh:
So its really only slow on your machine? That is going to be quite hard for us to debug. Anything unique that you can suggest might be related?

mojoaxel:
I can confirm this as a "problem". See also #3248

levzem:
When do you approximately think that you will be able to resolve this issue?
 Is there any fix that I can do in the meantime?

Thanks

levzem:
It appears that providing a reasonable start and end time in the options resolves all issues for anyone with the same problem. It can be marked as not a problem as along as this is strictly document that start and end are necessary for performance.

mojoaxel:
I can confirm that setting a start and end improves performance very much. 
We should have a look at this fixing #3248 !

We also should add a hint to the docs, that a missing start/end date can reduce performance.

yotamberk:
Will be added to docs

abrander:
[...]

Issue #3248, opened 2017-07-11
Timeline: Performance issues


mojoaxel:
I noticed some performance issues with groups.

[I created a demo](https://jsbin.com/vulenob) with 20 groups an 400 items:
![image](https://user-images.githubusercontent.com/600565/28072437-4d570036-6653-11e7-9cba-21151c6d7230.png)

[Source can be found on jsbin](https://jsbin.com/vulenob/edit?js,output).

The rendering takes about  2 seconds:
![image](https://user-images.githubusercontent.com/600565/28072277-d38cf378-6652-11e7-8737-3f3af584aa8c.png)

The main problem is, that every `RangeItem.redraw` triggers a DOM reflow:
![image](https://user-images.githubusercontent.com/600565/28072417-3a44ebd4-6653-11e7-9742-43258481e3c3.png)

Maybe it is possible to redraw the items all at once?

grimalschi:
It looks like question is not about redraw items all at once.

Problem is that we read properties, then we write properties, then we read properties again, and browser make reflow, then we write, read, then reflow again.

I think, the way to fix this is to separate reading attributes and writing.

Same as this library does: https://github.com/wilsonpage/fastdom

grimalschi:
And actually I dont understand why we dont have such a problem with no groups.

If we have no groups, but a lot of items, timeline works good.

Maybe there is special read/write properties operation in group redraw.

grimalschi:
@mojoaxel i have some success here.

This is performance before changes (from my app):

![image](https://user-images.githubusercontent.com/9027592/28291729-f870fe52-6b4b-11e7-83ce-7b416eb19511.png)

This is performance after changes:
![image](https://user-images.githubusercontent.com/9027592/28291816-6424d89e-6b4c-11e7-88e7-239f28e95677.png)

About 5 times increased performance :)

grimalschi:
In your case I have only 2 times performance increasing.

Before:
![image](https://user-images.githubusercontent.com/9027592/28292076-d562a8dc-6b4d-11e7-9966-e9fc6b622a60.png)

After:
![image](https://user-images.githubusercontent.com/9027592/28292095-e5a4502e-6b4d-11e7-888a-a00fbfe91c5a.png)

Demo: https://jsbin.com/bibehifufu/1/edit?html,js,output

grimalschi:
This is what I did: https://gist.github.com/grimalschi/df9a44ac12b5420e0cc6f2d42f1febed/revisions

[...]

PR #3313, opened 2017-07-29
Enable unit testing for module Network

wimrijnders:
Adds mocks for certain components, in order to let module `Network` be run in unit tests.

### Changes

- Create a mock object for `hammer.js` when running browserless. The inspiration is taken from [here](https://github.com/uber/deck.gl/pull/658)
- Create a mock function for `window.requestAnimationFrame()` when running browserless in `network/modules/CanvasRenderer.js`
- Added unit test for `Network` to show that unit testing for it now works
- Fixed naming of container in `test/Graph3d.test.js`

Since `hammer.js` is also used in other modules, this fix is potentially an enabler for full-module unit tests for those as well.

----

### Update

- Added fix for failing unit test. This was actually a unit test problem!
- Added unit tests for loading examples. This is because I noticed that some examples are failing
- Added fix to `Network` for example `disassemblerExample`. This is just to make the unit tests for this issue succeed.

I noticed that some examples are not working. I will scan them all and make a separate PR to fix all of these.

PR #3360, opened 2017-08-12
Add performance tips to timeline docs

yotamberk:
I've added "Performance Tips" to the docs to help users with issues like #3276.
Feel free to add more tips as we go along. 
I hope most of these performance tips could be eventually addressed in the code (such as warning when missing start and end in the options, fixing performance issues with redraws, such as suggested in #3248, etc.)

### DIFF
--- a/docs/timeline/index.html
+++ b/docs/timeline/index.html
@@ -2113,6 +2113,15 @@ var options = {
 &lt;/style&gt;
 </pre>
 
+ <h3 id="Performance_Tips">Performance  Tips</h3>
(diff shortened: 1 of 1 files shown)


wimrijnders:
Test is failing because of #3313 and related not covering all possibilities. The fail has nothing to do with the  content of the current PR. Chances are that rerunning the test will make it succeed.

Sorry about this, there were unexpected consequences with using `jscanvas` in the unit tests for `network`. Will make an effort to make it work flawless.

PR #3361, opened 2017-08-12
#3246 FIX - weekly scale minor label fix

ponml:
This is a follow up PR after an issue was discovered when a previous PR of mine was merged to `develop` 

see: #3246

Minor labels were displaying as `undefined`

see the `weekStyling.html` for an example.

image

### DIFF
--- a/lib/timeline/TimeStep.js
+++ b/lib/timeline/TimeStep.js
@@ -582,7 +582,7 @@ TimeStep.prototype.getLabelMinor = function(date) {
       if(this.isMajor() && date.weekday() !== 0){
           return "";
       }
-      break;
+      //no break; fall to the next case (default)
     default:
       return (format && format.length > 0) ? this.moment(date).format(format) : '';
   }
--- a/lib/timeline/TimeStep.js
+++ b/lib/timeline/TimeStep.js
@@ -579,14 +579,11 @@ TimeStep.prototype.getLabelMinor = function(date) {
   // noinspection FallThroughInSwitchStatementJS
   switch (this.scale) {
     case 'week':
-      if (this.isMajor() && date.weekday() !== 0) {
-        return "";
-      } else {
-        //default
-        return format && format.length > 0 ? this.moment(date).format(format) : '';  
+      if(this.isMajor() && date.weekday() !== 0){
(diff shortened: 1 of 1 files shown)


lib/timeline/TimeStep.js
```diff
@@ -582,7 +582,7 @@ TimeStep.prototype.getLabelMinor = function(date) {
       if(this.isMajor() && date.weekday() !== 0){
           return "";
       }
-      break;
+      //no break; fall to the next case (default)
```
ponml: From a code style POV, I wasn't sure what would be preferred 

1) use an else statement in the `week` case, so that it provides the default labeling. e.g.

```
if(this.isMajor() && date.weekday() !== 0){
     return "";
} else {
    return (format && format.length > 0) ? this.moment(date).format(format) : '';
}
```
2) remove the break statement, and let this fall to the default case. (I opted for this, with a  comment)

lib/timeline/TimeStep.js
```diff
@@ -579,12 +579,14 @@ TimeStep.prototype.getLabelMinor = function(date) {
   // noinspection FallThroughInSwitchStatementJS
   switch (this.scale) {
     case 'week':
-      if(this.isMajor() && date.weekday() !== 0){
-          return "";
+      if (this.isMajor() && date.weekday() !== 0) {
+        return "";
+      } else {
+        //default
+        return format && format.length > 0 ? this.moment(date).format(format) : '';
```
macleodbroad-wf: This seems a bit complicated to appease the linter. I think it would be better to use the code you had before and just add an lint ignore:

```javascript
  // noinspection FallThroughInSwitchStatementJS
  switch (this.scale) {
    case 'week':
      if (this.isMajor() && date.weekday() !== 0) {
        return "";
      }
    default:  // eslint-disable-line no-fallthrough
      return format && format.length > 0 ? this.moment(date).format(format) : '';
  }
};
```

ponml: Oh ya, that sounds great to me ( I also prefer what I originally had, but didn't know I could add an ignore).

mbroad:
+1

PR #3373, opened 2017-08-20
Enables "ClassDeclaration" for require-jsdoc lint rule, and adds all …

macleodbroad-wf:
…missing class-level jsdoc

R: @yotamberk


wimrijnders:
See my belated review on #3382 for my feedback/opinions of this.

Issue #3350, opened 2017-08-07
options-edges-color-inherit-to not working?


sawatts:
**VIS.JS 4.20.1**

Network of nodes with simple edges.  Using options to specify that edges should be coloured according to their destination (`options.edges.color.inherit = "to"`).  Edges are still coloured by according to their "from" node.

The network is defined in JSON (currently embedded in the JSP hosting the network - this will change).

The options are pulled from the server as a JSON file and loaded into the network object after it is created (immediately after), using an `XMLHttpRequest` call with `network.setOptions(o) ; network.redraw();` as the consumer function.

I _thought_ that the edges were coloured correctly (as "to") - so its possible that the error in the consumer or `setOptions`?  The rest of the configuration options are applied correctly.

The `options.json` being loaded is minimal, as follows:

	{
		"edges" : { "color" : { "inherit" : "to" } },
		"groups" :
		{
			"Group1" : {},
			"Group2" : {},
			"Group5" : 
			{
				"icon"  :
				{
					"face" : "FontAwesome",
					"code" : "\uf013",
					"size" : 20
				},
				"shape" : "icon"
			},
			"Group6" : {}
		}
	}

Tested on Windows 7 x64 with:
- Chrome 60.
- Firefox 54.

wimrijnders:
I did a quick test on the `basicExample` example of `network`, option `edges.color.inherit: 'to'` appears to be working correctly.

I made a [`jsbin`](http://jsbin.com/havale/1/edit?html,output) of this; can you look at it and see if you can change it to illustrate your issue?

sawatts:
I reproduced the issue from your example by setting the options after the network is created:
```
  var network = new vis.Network(container, data, {});
  network.setOptions (options);
```
Node 1 has edges _to_ nodes 2 and 3, which should be coloured green and red respectively (and are so when the network is created with the options).  When the options are applied after creation, the edges are colours blue as node 1.

wimrijnders:
I've been looking into this, there is definitely an issue here with several things going wrong.

I have trouble wrapping my mind around the option handling, it's quite tricky. This might take a while.

PR #3377, opened 2017-08-20
Fix setting of edge color options via Network.setOptions()

wimrijnders:
Fix for #3350

- Options for colors of edges where only set on initialization of `Network`; if `Network.setOptions()` was called after initialization, any color options directly affecting Edge instances were not set.
- Added several unit tests for edge colors in order to ensure proper working.
- The option handling for edges is pretty obscure; for this reason cleanup has been done and extensive commenting has been added.

Further changes:

- **Consolidated choosify()**. There where near-identical versions in `Node`, `Edge` and `Label`. This has been reduced to a single instance. Unit tests for this have been added for this.
- **Refactored updateLabelModule()** - common parts of this method in `Node` and `Edge` have been moved to `Label`.

### DIFF
--- a/lib/network/modules/EdgesHandler.js
+++ b/lib/network/modules/EdgesHandler.js
@@ -128,9 +128,9 @@ class EdgesHandler {
           // only forcibly remove the smooth curve if the data has been set of the edge has the smooth curves defined.
           // this is because a change in the global would not affect these curves.
           if (edgeData !== undefined) {
-            let edgeOptions = edgeData.smooth;
-            if (edgeOptions !== undefined) {
-              if (edgeOptions.enabled === true && edgeOptions.type === 'dynamic') {
+            let smoothOptions = edgeData.smooth;
+            if (smoothOptions !== undefined) {
+              if (smoothOptions.enabled === true && smoothOptions.type === 'dynamic') {
                 if (type === undefined) {
                   edge.setOptions({smooth: false});
                 }
@@ -367,6 +367,16 @@ class EdgesHandler {
   }
 
   create(properties) {
+    // It is not at all clear why all these separate options should be passed:
+    //
+    // - this.edgeOptions is set in setOptions()
+    //   the value of which is also added to this.options with parseOptions() 
+    // - this.defaultOptions has been added to this.options with util.extend() in ctor
+    //
+    // So, in theory, this.options should be enough.
+    // The only reason I can think of for this, is that precedence is important.
+    // TODO: make unit tests for this, to check if edgeOptions and defaultOptions are redundant
+    //
     return new Edge(properties, this.body, this.options, this.defaultOptions, this.edgeOptions)
   }
 
--- a/lib/network/modules/EdgesHandler.js
+++ b/lib/network/modules/EdgesHandler.js
@@ -108,7 +108,7 @@ class EdgesHandler {
       value: undefined
     };
 
-    util.extend(this.options, this.defaultOptions);
+    util.deepExtend(this.options, this.defaultOptions);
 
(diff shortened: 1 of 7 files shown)


lib/network/modules/components/Edge.js
```diff
@@ -1,30 +1,37 @@
...
- *                                width (number), style (string),
- *                                length (number), title (string)
- * @param {Network} network       A Network object, used to find and edge to
- *                                nodes.
- * @param {Object} constants      An object with default values for
- *                                example for the color
- * @class Edge
+ * An edge connects two nodes and has a specific direction.
+ *
+ * @class
```
yotamberk: The jsdocs have been removed. Did you intend to do that?

lib/network/modules/EdgesHandler.js
```diff
@@ -367,6 +367,16 @@ class EdgesHandler {
...
+    //
+    // - this.edgeOptions is set in setOptions()
+    //   the value of which is also added to this.options with parseOptions() 
+    // - this.defaultOptions has been added to this.options with util.extend() in ctor
+    //
+    // So, in theory, this.options should be enough.
+    // The only reason I can think of for this, is that precedence is important.
+    // TODO: make unit tests for this, to check if edgeOptions and defaultOptions are redundant
+    //
     return new Edge(properties, this.body, this.options, this.defaultOptions, this.edgeOptions)
```
yotamberk: I agree this is weird... Further work should be done to eliminate the separation of options. 
I'm okay with accepting these comments.

test/Network.test.js
```diff
@@ -12,18 +12,46 @@ var Validator = require("./../lib/shared/Validator").default;
...
 
 /**
+ * Merge all options of object b into object b
+ * @param {Object} a
+ * @param {Object} b
+ * @return {Object} a
+ *
+ * Adapted merge() in dotparser.js
+ */
+function merge (a, b) {
```
yotamberk: I feel like this function is really unnecessary. Lodash and even es6 have functions to do this. I would way prefer including lodash than using this function.

lib/network/modules/EdgesHandler.js
```diff
@@ -108,7 +108,7 @@ class EdgesHandler {
       value: undefined
     };
 
-    util.extend(this.options, this.defaultOptions);
+    util.deepExtend(this.options, this.defaultOptions);
```
yotamberk: Why aren't we using Lodash for extending?

wimrijnders: Beats me; I only knew `lodash` by name, not what it does. It might be worth investigating, and perhaps if relevant open a new issue for conversion.

I just used the functionality that is present.

lib/network/modules/components/Edge.js
wimrijnders: Yes, for the following reasons:

- Not one of the documented `@param` values exist; they appear to be an artifact of previous changes.
- The `@param` definitions are in the wrong place anyway; they should be defined at the constructor, not the class.

A new correct header comment has been added to the constructor to correct this.

lib/network/modules/EdgesHandler.js
wimrijnders: OK. This is something I intend to pick up in due time, the comment serves as a reminder/warning. It's just that it was IMHO outside of the scope of the issue, and I was doing too much already.

test/Network.test.js
wimrijnders: Sure. I propose to address `lodash` integration in a separate issue. I'm trying to fix a confirmed bug here.

PR #3394, opened 2017-08-27
DataSet - Adds unit tests for add, setOptions and on/off

macleodbroad-wf:
@yotamberk @wimrijnders

### DIFF
--- a/lib/DataSet.js
+++ b/lib/DataSet.js
@@ -73,11 +73,6 @@ function DataSet (data, options) {
     }
   }
 
-  // TODO: deprecated since version 1.1.1 (or 2.0.0?)
-  if (this._options.convert) {
-    throw new Error('Option "convert" is deprecated. Use "type" instead.');
-  }
-
   this._subscribers = {};  // event subscribers
 
   // add initial data when provided
@@ -122,7 +117,7 @@ DataSet.prototype.setOptions = function(options) {
 
 /**
  * Subscribe to an event, add an event listener
- * @param {String} event        Event name. Available events: 'put', 'update',
+ * @param {String} event        Event name. Available events: 'add', 'update',
  *                              'remove'
  * @param {function} callback   Callback method. Called with three parameters:
  *                                  {String} event
@@ -142,14 +137,6 @@ DataSet.prototype.on = function(event, callback) {
 };
 
 /**
- * TODO: remove this deprecated function some day (replaced with `on` since version 0.5, deprecated since v4.0)
- * @throws {Error}
- */
-DataSet.prototype.subscribe = function () {
-  throw new Error('DataSet.subscribe is deprecated. Use DataSet.on instead.');
-};
-
-/**
  * Unsubscribe from an event, remove an event listener
  * @param {String} event
  * @param {function} callback
@@ -159,11 +146,6 @@ DataSet.prototype.off = function(event, callback) {
   if (subscribers) {
     this._subscribers[event] = subscribers.filter(listener => listener.callback != callback);
   }
-};
-
-// TODO: remove this deprecated function some day (replaced with `on` since version 0.5, deprecated since v4.0)
-DataSet.prototype.unsubscribe = function () {
-  throw new Error('DataSet.unsubscribe is deprecated. Use DataSet.off instead.');
 };
 
 /**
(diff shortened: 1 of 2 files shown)


test/DataSet.test.js
```diff
@@ -1,44 +1,43 @@
...
-    data.add([
-      //{id: 3, content: 'Item 3', start: moment(now)}, // TODO: moment fails, not the same instance
-      {id: 3, content: 'Item 3', start: now},
-      {id: 4, content: 'Item 4', start: '/Date(' + now.valueOf() + ')/'}
-    ]);
+  var sort = function (a, b) {
+    return a.id > b.id;
+  };
+
+  it('can handle multiple data types', function () {
```
wimrijnders: With respect to usage of variable `data`, defined globally in the `describe`-block:

You are making a potentially dangerous assumption here, namely that the tests will *all* run, and will always run in the *same order*.

From the `mocha` command line usage:

```
-g, --grep <pattern>                    only run tests matching <pattern>
-f, --fgrep <string>                    only run tests containing <string>
```

I.e. it is possible to run specific tests, not all of them. In addition, a fix to `mocha` for optionally randomizing the test order is pending.

While this is not directly an issue with the current tests, it's something which time or a user can (and therefore will) break.

My suggestions here are:

- Wrap the initialization of `data` in an initialization function, call this in every `it`-block
- *and/or* initialize the data in a `beforeEach` block.

These can be combined. The goal in any case is to allow the `it`-blocks to run independently.

test/DataSet.test.js
```diff
@@ -1,44 +1,43 @@
...
+    }
+  });
 
-    var data = new DataSet({
-      type: {
-        start: 'Date',
-        end: 'Date'
-      }
-    });
+  // add single items with different date types
```
wimrijnders: See comment gelow on initialization of used data. In any case, don't leave this init code to run 'bare'.

test/DataSet.test.js
```diff
@@ -1,44 +1,43 @@
...
-      {id: 3, content: 'Item 3', start: now},
-      {id: 4, content: 'Item 4', start: '/Date(' + now.valueOf() + ')/'}
-    ]);
+  var sort = function (a, b) {
+    return a.id > b.id;
+  };
+
+  it('can handle multiple data types', function () {
 
     var items = data.get();
```
wimrijnders: Ensure that var `data` is always in a well-defined state before  running any check on it. This comment applies to *all* usages of `data` in this file.

test/DataSet.test.js
```diff
@@ -282,7 +285,111 @@ describe('DataSet', function () {
       {id: 3, content: 'Item 3'},
       {id: 4, content: 'Item 4'}
     ]);
+  });
+
+  describe('add', function () {
+    var dataset = new DataSet([]);
```
wimrijnders: See comments in previous `describe`-blocks on usage of bare data. Please just ensure that data is re-initialized upon execution of every `it`-block.

test/DataSet.test.js
```diff
@@ -282,7 +285,111 @@ describe('DataSet', function () {
       {id: 3, content: 'Item 3'},
       {id: 4, content: 'Item 4'}
     ]);
+  });
+
+  describe('add', function () {
```
wimrijnders: 👍 For new tests. Looks clean except for the data initialization thing.
This is stuff which is assumed to 'just work'. Which is exactly the reason why they should be regression-tested.

macleodbroad-wf: Has been addressed

macleodbroad-wf:
@wimrijnders - I've addressed your comments

lib/DataSet.js
```diff
@@ -122,7 +117,7 @@ DataSet.prototype.setOptions = function(options) {
 
 /**
  * Subscribe to an event, add an event listener
- * @param {String} event        Event name. Available events: 'put', 'update',
+ * @param {String} event        Event name. Available events: 'add', 'update',
```
wimrijnders: Yesterday, I learned this should be `{string}`. But I suppose that the other PR will handle that.

lib/DataSet.js
```diff
@@ -234,6 +216,7 @@ DataSet.prototype.add = function (data, senderId) {
  * @param {Object | Array} data
  * @param {String} [senderId] Optional sender id
  * @return {Array} updatedIds     The ids of the added or updated items
+ * @throws {Error} Error - Unknown Datatype
```
wimrijnders: Second `Error` shouldn't be there, just the description.

lib/DataSet.js
macleodbroad-wf: Yes, to be addressed in a follow up PR.

yotamberk:
tests are failing again. Not because of this PR

PR #3396, opened 2017-08-28
Fixed #3395: allow for multiline titles

### LINKED ISSUE #3395: Funky resizing with multiline titles
I have encountered a bug in the vis.js network component.

When I make a title of a node multiline with the style properties `maxWidth` and `whiteSpace`, I encounter unwanted behavior.

When I hover over a node with a small text, and afterward hover over a node with a long multiline title, the small title-div is reused, and is rendered too small. When I wiggle the mouse over the node, the title grows slowly to the desired width. What I want is a title that has directly the correct size, not after wiggling the mouse over the node.

This bug can be reproduced in this js-bin: http://output.jsbin.com/hobeparusa

I run:
* vis.js 4.20.1 (latest)
* Chrome Version 59.0.3071.115 (64-bit)
* Ubuntu 16.04 LTS

michel404:
(#3395)

### DIFF
--- a/lib/shared/Popup.js
+++ b/lib/shared/Popup.js
@@ -118,6 +118,8 @@ class Popup {
    */
   hide() {
     this.hidden = true;
+    this.frame.style.left = "0";
+    this.frame.style.top = "0";
     this.frame.style.visibility = "hidden";
   }
 


wimrijnders:
I'm am quite amazed that this simple change fixes the 'funky resizing' issue. Would you mind explaining why this works?

michel404:
> Would you mind explaining why this works?

I'm not sure wat _causes_ the weird behavior. I suppose this has to with the fact that there is one div that is reused for every tooltip, and that the browsers text reflowing algorithm gets distorted when the the div is already (albeit hidden) somewhere on the screen.

I've observed that when the multiline tooltip is the first tooltip that is showed, it works fine; and when a smaller tooltip is showed before (causing the div to hang somewhere invisibly on screen) showing the multiline tooltip starts of weird.

My approach to fixing this weird behavior is resetting the position of the hidden div, such that previously showed tooltips do not have an effect on the text of the newly opened tooltips.

wimrijnders:
Still abracadabra to me, but it's a clean fix and it works, so 👍. Also check the examples and the tooltips there continue to work as usual.

PR #3407, opened 2017-09-01
Resolving comparison bug identified by SonarQube

dbhubbell:
Sonar says:

> Remove this "!==" check; it will always be true. Did you mean to use "!="?
> 
> Comparing dissimilar types using the strict equality operators === and !== will always return the same value, respectively false and true, because no type conversion is done before the comparison. Thus, such comparisons can only be bugs.


wimrijnders:
Sonar is wrong. You can test this simply by trying out `a === undefined` for various values, including `undefined`, in your favorite javascript environment.

In addition, you're trying to merge the `master` branch into `develop`. You're supposed to do the changes in `develop` only. I believe the default text on opening a new PR mentions this.

PR #3409, opened 2017-09-02
Group redraw performance

yotamberk:
Addressing some of the group performance issues and implementation of discussion from #3248 (using gist https://gist.github.com/grimalschi/df9a44ac12b5420e0cc6f2d42f1febed/revisions#diff-32e31f368549caca635e81cdb8a0a82dR21132)
This is the best implementation in the meantime - I've tried using fasdom and some other methods but concluded that it's not what we need. Fastdom allows prioritization of "read" operations over "write" operations, whilst our problem has to do with paralleling the operations done in a group redraw for all groups. 
Further analysis that will be done:
- The exact operations done in the group redraw - we might be able to use fasdom/combining "reads" and "writes" to fewer operations for every redraw queue. 
- Implementation like this should also be done in items' redraws.

Shout out to @grimalschi for supplying the solution.

### DIFF
--- a/lib/timeline/component/Group.js
+++ b/lib/timeline/component/Group.js
@@ -214,109 +214,149 @@ Group.prototype.getLabelWidth = function() {
  * @param {boolean} [forceRestack=false]  Force restacking of all items
  * @return {boolean} Returns true if the group is resized
  */
-Group.prototype.redraw = function(range, margin, forceRestack) {
-  var resized = false;
- 
-  // force recalculation of the height of the items when the marker height changed
-  // (due to the Timeline being attached to the DOM or changed from display:none to visible)
-  var markerHeight = this.dom.marker.clientHeight;
-  if (markerHeight != this.lastMarkerHeight) {
-    this.lastMarkerHeight = markerHeight;
-    util.forEach(this.items, function (item) {
-      item.dirty = true;
-      if (item.displayed) item.redraw();
-    });
+Group.prototype.redraw = function(range, margin, forceRestack, returnQueue) {
+  var queue = [];
 
-    forceRestack = true;
-  }  
-  
-  // recalculate the height of the subgroups
-  this._calculateSubGroupHeights(margin);
+    queue.push((function () {
+      // force recalculation of the height of the items when the marker height changed
+      // (due to the Timeline being attached to the DOM or changed from display:none to visible)
+      var markerHeight = this.dom.marker.clientHeight;
+      if (markerHeight != this.lastMarkerHeight) {
+        this.lastMarkerHeight = markerHeight;
+        util.forEach(this.items, function (item) {
+          item.dirty = true;
+          if (item.displayed) item.redraw();
+        });
 
-  // calculate actual size and position
-  var foreground = this.dom.foreground;
-  this.top = foreground.offsetTop;
-  this.right = foreground.offsetLeft;
-  this.width = foreground.offsetWidth;
+        forceRestack = true;
+      }
+    }).bind(this));
 
-  var lastIsVisible = this.isVisible;
-  this.isVisible = this._isGroupVisible(range, margin);
-  
-  var restack = forceRestack || this.stackDirty || (this.isVisible && !lastIsVisible);
+    queue.push((function () {
+      // recalculate the height of the subgroups
+      this._updateSubGroupHeights(margin);
+    }).bind(this));
 
-  this._updateSubgroupsSizes();
-  // if restacking, reposition visible items vertically 
-  if(restack) {
-    if (typeof this.itemSet.options.order === 'function') {
-      // a custom order function
-      // brute force restack of all items
+    queue.push((function () {
+      // calculate actual size and position
+      var foreground = this.dom.foreground;
+      this.top = foreground.offsetTop;
+      this.right = foreground.offsetLeft;
+      this.width = foreground.offsetWidth;
+    }).bind(this));
 
-      // show all items
-      var me = this;
-      var limitSize = false;
-      util.forEach(this.items, function (item) {
-        if (!item.dom) { // If this item has never been displayed then the dom property will not be defined, this means we need to measure the size
-          item.redraw();
-          me.visibleItems.push(item);
+    var resized;
+    var lastIsVisible;
+
+    queue.push((function () {
+      resized = false;
+      lastIsVisible = this.isVisible;
+      this.isVisible = this._isGroupVisible(range, margin);
+    }).bind(this));
+
+    queue.push((function () {
+      var restack = forceRestack || this.stackDirty || this.isVisible && !lastIsVisible;
+
+      // if restacking, reposition visible items vertically
+      if (restack) {
+        if (typeof this.itemSet.options.order === 'function') {
+          // a custom order function
+          // brute force restack of all items
+
+          // show all items
+          var me = this;
+          var limitSize = false;
+          util.forEach(this.items, function (item) {
+            if (!item.displayed) {
+              item.redraw();
+              me.visibleItems.push(item);
+            }
+            item.repositionX(limitSize);
+          });
+
+          // order all items and force a restacking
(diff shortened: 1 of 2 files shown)


yotamberk:
@wimrijnders @macleodbroad-wf  can one of you please review and approve this? It is extremely important

grimalschi:
Actually I did not think you will use my code :)

That was just demo to show the idea.

yotamberk:
if the shoe fits, wear it.

The code is good. It might be cleaner but for now it's good. I've tried using fasdom and some other techniques but always came back to your code as being clearer than what I did. I'm working on further work on items redraw. I hope to have anther refactor iteration to clean this code

lib/timeline/component/Group.js
```diff
@@ -212,119 +212,160 @@ Group.prototype.getLabelWidth = function() {
...
-
-      // show all items
-      var me = this;
-      var limitSize = false;
-      util.forEach(this.items, function (item) {
-        if (!item.dom) { // If this item has never been displayed then the dom property will not be defined, this means we need to measure the size
-          item.redraw();
-          me.visibleItems.push(item);
+Group.prototype.redraw = function(range, margin, forceRestack, returnQueue) {
+  var queue = [];
```
macleodbroad-wf: If pure performance is the concern, I think it would be more readable slightly more performant (maybe a micro optimization?) to have all of these anonymous functions defined at a higher level, and then simply:

```javascript
var queue = [recalcHeight(this), recalcSubGroupHeights(this), calcSizeAndPosition, etc...];
if (returnQueue) {
    return queue;
} else {
    var result;
    queue.forEach(function (fn) {
        result = fn(this);
    });
    return result;
}
...
```

yotamberk: I agree but I was planning doing it in another PR

macleodbroad-wf: How often does redraw get called?

yotamberk: every zoom/scroll/drag - alot

macleodbroad-wf: Ok, can you include a reference to the follow-up issue? Just want to make sure it doesn't get lost.

yotamberk: I can include it here if you think it's really important. 
Wait with this PR if so

macleodbroad-wf: I think it's worth doing, given how often this is used.

lib/timeline/component/ItemSet.js
```diff
@@ -681,13 +681,37 @@ ItemSet.prototype.redraw = function() {
...
+    var returnQueue = true;
+    redrawQueue[key] = group.redraw(range, groupMargin, forceRestack, returnQueue);
+    redrawQueueLength = redrawQueue[key].length;
   });
+
+  if (redrawQueueLength) {
+    var redrawResults = {};
+
+    for (var i = 0; i < redrawQueueLength; i++) {
+      util.forEach(redrawQueue, function (fns, key) {
```
wimrijnders: Oof this stuff is almost over the top of my head.

But shouldn't this be: `util.forEach(redrawQueue[i]...`?

`redrawQueue` is a hash, of which every value is a queue of functions to run. I'm inferring that you want to run all function in the queue for every key of `redrawQueue`.

If I got it wrong here, please explain.

lib/timeline/component/ItemSet.js
```diff
@@ -681,13 +681,37 @@ ItemSet.prototype.redraw = function() {
...
+        redrawResults[key] = fns[i]();
```
wimrijnders: But then the `[i]` does not make sense. Why go through all the trouble of creating a `Queue` instance, when the separate items in it are meant for set individual items?

lib/timeline/component/Group.js
```diff
@@ -212,119 +212,160 @@ Group.prototype.getLabelWidth = function() {
...
-    item.repositionY(margin);
-    if (!this.isVisible && this.groupId != "__background__") {
-      if (item.displayed) item.hide();
-    }
-  }
 
-  if (!this.isVisible && this.height) {
-    return resized = false;
-  }
+      if (!this.isVisible && this.height) {
```
wimrijnders: Tabs?

lib/timeline/component/ItemSet.js
grimalschi: Im sorry for bad naming :(

`group.redraw(range, groupMargin, forceRestack, returnQueue)` returns array of functions.

So redrawQueue is an queue, and every element in this queue is array of functions.

And when we release this queue, we call all functions from every element of queue.

wimrijnders: Nice to see you react to a question on the code. I was expecting @yotamberk, but I understand the original comes from you.

Will try to wrap my brain around it again. Perhaps some commenting would be in order here.

yotamberk:
I've reacted to the comments. Please address.
@wimrijnders @macleodbroad-wf

lib/timeline/component/Group.js
```diff
@@ -206,158 +206,174 @@ Group.prototype.getLabelWidth = function() {
...
+}
+
+Group.prototype._applyGroupHeight = function(height) {
+  this.dom.background.style.height = height + 'px';
+  this.dom.foreground.style.height = height + 'px';
+  this.dom.label.style.height = height + 'px';
+}
+
+Group.prototype._updateItemsVerticalPosition = function(resized, margin) {
+  // update vertical position of items after they are re-stacked and the height of the group is calculated
```
wimrijnders: Might be better as function header comment

lib/timeline/component/Group.js
```diff
@@ -206,158 +206,174 @@ Group.prototype.getLabelWidth = function() {
...
+  for (var i = 0, ii = this.visibleItems.length; i < ii; i++) {
+    var item = this.visibleItems[i];
+    item.repositionY(margin);
+    if (!this.isVisible && this.groupId != "__background__") {
+      if (item.displayed) item.hide();
+    }
+  }
+
+  if (!this.isVisible && this.height) {
+    return resized = false;
```
wimrijnders: Why this construct? I think `return false` would be enough here

lib/timeline/component/Group.js
```diff
@@ -206,158 +206,174 @@ Group.prototype.getLabelWidth = function() {
...
+}
+
+Group.prototype._didResize = function(resized, height) {
+  resized = util.updateProperty(this, 'height', height) || resized;
+  // recalculate size of label
+  var labelWidth = this.dom.inner.clientWidth;
+  var labelHeight = this.dom.inner.clientHeight;
+  resized = util.updateProperty(this.props.label, 'width', labelWidth) || resized;
+  resized = util.updateProperty(this.props.label, 'height', labelHeight) || resized;
+  return resized
```
wimrijnders: ;

lib/timeline/component/Group.js
```diff
@@ -206,158 +206,174 @@ Group.prototype.getLabelWidth = function() {
...
+    }).bind(this),
+
+    // apply group height
+    (function() {
+      this._applyGroupHeight.bind(this)(height)
+    }).bind(this),
+
+    // update vertical position of items after they are re-stacked and the height of the group is calculated
+    (function() {
+      return resized = this._updateItemsVerticalPosition.bind(this)(resized, margin)
```
wimrijnders: For my understanding: this is the last function in the queue and it returns something. The return value is collected [here](https://github.com/yotamberk/vis/blob/329d9817f94d53738222be9ae361633e3a912209/lib/timeline/component/ItemSet.js#L701). Is this the intention?

wimrijnders: I suppose the assignment to `resize` isn't really necessary for the return, but it's kept in for completeness. I guess this is OK (I would probably do the same).

yotamberk:
@wimrijnders done! Thanks!

Issue #824, opened 2015-05-11
Network: detect click on edge's label?


rwu823:
Currently I can detect click on whole edge if  `prop.edges.length === 1 && !prop.nodes.length`, but is it possible detect clicked on edge's label?

AlexDM0:
Hi,

No, we do not take the label into account when checking for clicks.
We could put this up as a feature request?

Regards

AlexDM0:
Hi, 

I've added this to the feature requests:
http://visjs.org/featureRequests.html

We cannot say when this feature will be implemented at this time. We try to collect all feature requests on our website and close the issues. This does not mean we will ignore this request. More information on this policy can be found here:

http://visjs.org/blog.html#New\ website\ for\ vis\.js\!

Regards,

Alex

mojoaxel:
Reopening as Feature-Request issue (see #2114). 
Everybody: Please feel free to implement this!

wimrijnders:
This issue has been labeled `Inactive`, because there has been no activity here for over two months.

You can:
- Close the issue yourself, so that we know we don't need to take it into account (thanks!).
- Post a new comment to keep it active. The Inactive label and this message will be removed.
- Ignore this comment. If the inactivity continues, the issue will be closed in two weeks.

wimrijnders:
Keeping this open, there appears to be more interest for this (e.g. #3261).

karenhanson:
Could this be broadened to include a click event on node labels as well (where node labels are outside of the nodes), or should I log a separate feature request?  I would like to be able to make the behavior of a node click also happen when someone clicks the node label under it.  Thanks!

wimrijnders:
@karenhansen I'm looking at it right now and yes, I'll take any label into account, irrespective of their positioning to the parent node/edge. This falls under the current issue, so don't bother with a new one.

I'm actually taking it a bit further than that, because in the future, due to known pending requests, I expect there to be:

- Multiple labels per node/edge
- z-ordering for nodes/edges
- option for returning all clicked on items instead of just the topmost.

It's not that these will be implemented right away, but I'd like the label click event to be ready for them.

wimrijnders:
Got it working. Need to do documentation as well before submitting. And clean up of course.

PR #3410, opened 2017-09-03
Network detect clicks on labels

wimrijnders:
Fix for #0824.

Clicks on nodes and edge labels are detected and passed to the `click` event.

- All items clicked on are returned, in reverse z-order[I]
- Adjustments take into account the rotation of edge labels
- Refactoring done in order to be able to use certain code more generally
- Adjusted example `network/labels/labelAlignment` to illustrate the click functionality
- Adjusted the documentation for the new returned property of the `click` event

The labels have their own click-detection, which means that their clicks can be detected independent from the parent node/edge.

A new property `items` is added to the parameter of the `click`-handler. This is an array, reverse sorted on 'depth'[I], which can contain the following hashes:

```js
  {nodeId:number}            // node with given id clicked on
  {nodeId:number labelId:0}  // label of node with given id clicked on
  {edgeId:number}            // edge with given id clicked on
  {edge:number, labelId:0}   // label of edge with given id clicked on
```

----

[I] True z-ordering is not implemented yet. Currently, the drawing order is used as 'depth'


docs/network/index.html
```diff
@@ -1323,6 +1323,24 @@ <h2 id="Events">Events</h2>
...
+...
+  items: [Array of click items],
+}</pre>
+
+Where the click items can be:
+<pre class="prettyprint lang-js">
+  {nodeId:number}            // node with given id clicked on
+  {nodeId:number labelId:0}  // label of node with given id clicked on
+  {edgeId:number}            // edge with given id clicked on
+  {edge:number, labelId:0}   // label of edge with given id clicked on
```
mbroad: Can't edge and node IDs be strings as well?

lib/network/modules/components/Edge.js
```diff
@@ -605,6 +614,36 @@ class Edge {
 
 
   /**
+   * Determine all visual elements of this edge instance, in which the given
+   * point falls within the bounding shape.
+   *
+   * @param {point} point
+   * @returns {Array.<edgeClickItem|edgeLabelClickItem>} list with the items which are on the point
```
mbroad: Can't this also return nodes?

lib/network/modules/components/shared/ComponentUtil.js
```diff
@@ -1,3 +1,19 @@
...
+ * These are more or less global within Network. Putting them here until I can figure out
+ * where to really put them
+ * 
+ * @typedef {string|number} id
+ * @typedef {{x: number, y: number}} point
+ * @typedef {{left: number, top: number, width: number, height: number}} rect
+ * @typedef {{x: number, y:number, angle: number}} rotationPoint
+ *   - point to rotate around and the angle in radians to rotate. angle == 0 means no rotation
+ * @typedef {{nodeId:id}} nodeClickItem
+ * @typedef {{nodeId:id, labelId:id} nodeLabelClickItem
```
mbroad: Missing closing curly

lib/network/modules/components/shared/ComponentUtil.js
```diff
@@ -49,6 +65,56 @@ class ComponentUtil {
...
+   * @param {rotationPoint} [rotationPoint] if specified, the rotation that applies to the rectangle.
+   * @returns {boolean}  true if point within rectangle, false otherwise
+   * @static
+   */
+  static pointInRect(rect, point, rotationPoint) {
+    if (rect.right <= rect.left || rect.bottom <= rect.top) {
+      return false;  // early out
+    }
+
+    if (rotationPoint !== undefined) {
```
mbroad: I prefer if we invert this if statement and do an early return, and then we can unindent the else case

lib/network/modules/components/shared/ComponentUtil.js
```diff
@@ -49,6 +65,56 @@ class ComponentUtil {
...
+      // Rotate the point the same amount as the rectangle
+      var tmp = {
```
mbroad: Can we rename all tmp variables to more descriptive names, like `candidatePoint` or `rotatedPoint`?

lib/network/modules/components/shared/Label.js
```diff
@@ -656,6 +628,50 @@ class Label {
...
+      }
+    }
+
+    var ret = {
+      left  : x,
+      top   : y,
+      width : this.size.width,
+      height: this.size.height,
+
+      // TODO: see if following required
```
mbroad: Is this meant to be addressed before merge?

docs/network/index.html
```diff
@@ -1323,6 +1323,24 @@ <h2 id="Events">Events</h2>
...
+</pre>
+
+The order of the <code>items</code> array is descending in z-order.
+Thus, to get the topmost item, get the value at index 0.
                 </td>
```
mbroad: Does this change require a follow up pr to the gh-pages branch?
If so, can you create an issue and reference it from this pr?

examples/network/labels/labelAlignment.html
```diff
@@ -25,8 +25,12 @@
 <p>Text-alignment within node labels can be 'left' or 'center', other font alignments not implemented.</p>
 <p>Label alignment (placement of label &quot;box&quot;) for nodes (top, bottom, left, right, inside) is
 planned but not in vis yet.</p>
+<p>The click event is captured and displayed, to illustrate how the clicking on labels works.
```
mbroad: Unnecessary comma

examples/network/labels/labelAlignment.html
```diff
@@ -25,8 +25,12 @@
...
+You can drag the nodes over each other, to see how this influences the click event values.
```
mbroad: Same as above

lib/network/modules/LayoutEngine.js
```diff
@@ -1743,11 +1743,13 @@ class LayoutEngine {
     if (nodeArray.length > 1) {
       if(this._isVertical()) {
         nodeArray.sort(function (a, b) {
+          if (a.x === undefined || b.x === undefined) return 0;
```
mbroad: How often do these sorting functions get used?

Might be worth extracting these sort functions out to top-level functions

lib/network/modules/components/edges/util/EdgeBase.js
```diff
@@ -415,6 +415,9 @@ class EdgeBase {
       returnValue = Math.abs(Math.sqrt(dx * dx + dy * dy) - radius);
     }
 
+    // This is **wrong**! It doesn't take into account labels which are displayed rotated.
+    // It only works for horizontal labels.
+    // In addition, it's in violation of what this method is supposed to do.
```
mbroad: Should an issue be opened?

mbroad:
How does this behave on very large graph with lots of nodes and edges? Does it freeze up when you click on things?

lib/network/modules/components/shared/ComponentUtil.js
wimrijnders: Curly! CURLY!!!! Come here, boy!

...damn curly's running away every time. Back to obedience school it is for him.

docs/network/index.html
wimrijnders: Indeed. Would a `@typedef` be acceptable here? e.g.

```
@typedef {number|string} Id
```
or

```
@typedef {number|string} NodeId
@typedef {number|string} NdgeId
@typedef {number|string} LabelId
```

lib/network/modules/components/shared/ComponentUtil.js
wimrijnders: This would mean that the return statement would be doubled in the code. I prefer this construct.

lib/network/modules/components/shared/ComponentUtil.js
wimrijnders: For small scopes I prefer short names, I believe this enhances readability. Also, `tmp` has meaning here; it's a clear indication that this is meant as a temporary variable.

lib/network/modules/components/shared/Label.js
wimrijnders: Sort of. I'll just address it now to be done with it.

docs/network/index.html
wimrijnders: I don't think so; the docs in `gh-pages` are updated from `develop` upon release.

lib/network/modules/LayoutEngine.js
wimrijnders: These will change in #3411

lib/network/modules/components/edges/util/EdgeBase.js
wimrijnders: Well spotted.

The consequence of removing this is that click will not work any more on a horizontal label in the regular click event, thus external changes. I'm a bit hesitant of removing it because users might expect it. I know, paranoia.

What's your opinion, given the comment? Can this be removed?

wimrijnders:
Good, you're critical in the review. I like that.
Next time, feel free to click the 'Changes requested' button in the review dialog.

wimrijnders:
> How does this behave on very large graph with lots of nodes and edges? Does it freeze up when you click on things?

Tried this with example `worldCupPerformance`, almost 1000 nodes and 10000 edges. I don't feel any noticeable difference in responsiveness. YMMV.

docs/network/index.html
mbroad: Imo, a top level typedef `@typedef {number|string} Id` and then
```
@typedef {Id} NodeId
@typedef {Id} EdgeId
@typedef {Id} LabelId
```
But maybe that's needless indirection.

Personally I like it, because it's descriptive and flexible for the future.

lib/network/modules/components/shared/ComponentUtil.js
mbroad: I won't fight you on this, but rather agree to disagree.

lib/network/modules/LayoutEngine.js
mbroad: K

lib/network/modules/components/edges/util/EdgeBase.js
mbroad: I don't know, do we need to follow up on rotated labels?

mbroad:
I'm glad you like a critical review. I've taken flack for it in the past.

docs/network/index.html
wimrijnders: I like it too; the only qualm that I have about it, is that it will duplicate some text within the definitions, e.g.

```
@typedef {{nodeId:NodeId}} param
```
Actually, that isn't so bad. I'm actually inclined to go with this. Will let it simmer a bit to see if I'm going with this.

lib/network/modules/components/shared/ComponentUtil.js
wimrijnders: ok

lib/network/modules/components/edges/util/EdgeBase.js
wimrijnders: Clicking on rotated labels is the new bit, previously only clicking on horizontal labels worked.

Thinking about this, I realized that the question is: Should clicking on a node/edge label also select the given node/edge?

For labels within a node or on an edge, this is a no-brainer. For outside labels, less so.
I'm inclined to just remove that code, and rebuild the node selection around given method.

Also, I'll scan the issues to see what the users are expecting. Chances are someone has someone already has raised qualms about clicking labels to select.

docs/network/index.html
wimrijnders: OK, I'm won over. It's the most complete definition for the id's.

macleodbroad-wf:
So does this run `getClickedItems` even if there's no callback handler to return data to?

wimrijnders:
Yes, it does, but that applies to all internal event handlers.

I examined the code, what happens: all click events are caught and all internal handlers are run, always. The results of these, however, are discarded afterwards if there are no callbacks attached for given event.

There is room for optimization here, in the sense of first detecting if a given callback has been defined, then only running a handler if callback is present.

yotamberk:
Can you please resolve conflicts?

wimrijnders:
Conflicts fixed.

Issue #3403, opened 2017-08-31
Node order problem when using hierarhical layout


KimRasmussen:
When using hierarchical layout with 2 levels and > 10 items in the 2nd level, the sort order gets weird.

I have a simple example here: http://jsfiddle.net/sg27ycbn/

Here, you can see the order of nodes in the 2nd level is all wrong. If I comment out the line 

edges.add({from: 1, to: 12});

then the remaining 9 nodes connected to the root gets shown sorted in the order they were inserted in, but as soon as there are 10 of them, the order  gets weird.

I assume that is a bug and not intentional - is there any way I can fix this ? Any options that should have been set different ?

wimrijnders:
It looks perfectly fine to me; perhaps there's some extra step I need to do to reproduce?

Having said that, there are known problems with the hierarchical ordering, see e.g. #3357. But for a simply connected network like this, there should not be any problem.

KimRasmussen:
I am sorry that I was unclear in my description of the issue.

As it looks now, the nodes on the 2nd level is in this order:
2, 12, 4, 5, 6, 3, 8, 9, 10, 11, 7.

If I remove the last edges.add line, the appear in the order I inserted them (which is what I would expect them to do).
2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12

As soon as the last edge (from 1 to 12) is added, the order of all the nodes get messed up.

wimrijnders:
No, sorry, I really don't see that happening.
Perhaps it's platform-specific? I'm doing this on linux+firefox. Want OS/browser are you using?

KimRasmussen:
Ah, it seems to be browser specific. It works just fine on firefox on windows, but not on chrome.
Seems to work fine in internet explorer and edge as well - so it looks like it is only chrome that has the problem.

Didn't at all think about trying it on other browsers - sorry.

wimrijnders:
Well, I'll be. Confirmed. This is truly weird.

![untitled](https://user-images.githubusercontent.com/769037/29996700-fdd70c88-9003-11e7-80c2-34241f067546.png)

Now, why on earth would chrome do it differently?

Okay, I have something to go on now. Will keep you posted.

wimrijnders:
Found it. It has to do with the `Array.sort()` method, which I can imagine it to be browser-specific.

[...]

PR #3411, opened 2017-09-03
Network: Fixes sorting on Layout, refactoring.

wimrijnders:
Fix for #3403

Chromium has divergent behaviour on sorting of `undefined` values, a check has been added for these.

The fix itself is small, it adds a check on `undefined` in the sorting function(s).

In addition:

- Fixed off-by-one error on tree index
- Refactored away checks on direction by replacing `_isVertical()` with a Strategy pattern for direction

The latter removes a long-standing eyesore in `LayoutManager`; it's probably also faster.

There is some hope that these fixes will improve the hierarchical layout initialization, which has had
problems for quite a while.


rahyadav:
Hi @wimrijnders,

I'm facing an issue where my Hierarchical layout works fine in edge/IE. However in Chrome, It renders during stabilization and then the browser **scroll** starts **shrinking** continuously, and the graph gets lost.
Is it a familiar issue?
This happens only on all the laptops similar to mine, not on other laptops.
Something to do with screen res or windows version? Here are the os details:
image

wimrijnders:
@rahyadav I've seen exploding networks, but shrinking ones are new to me. Would you mind making a separate issue of that, with an example?

lib/network/modules/LayoutEngine.js
```diff
@@ -110,6 +110,8 @@ class HierarchicalStatus {
    * @param {String|Number} treeId
    */
   setTreeIndex(node, treeId) {
+    if (treeId === undefined) return;  // Don't bother
```
mbroad: Is this an error condition? Should some warning be elicited?

lib/network/modules/LayoutEngine.js
```diff
@@ -1281,9 +1273,9 @@ class LayoutEngine {
 
         // we get the X or Y values we need and store them in pos and previousPos.
         // The get and set make sure we get X or Y
-        if (i === 0) {pos = this._getPositionForHierarchy(this.body.nodes[parentId]);}
-        else         {pos = this._getPositionForHierarchy(childNodes[i-1]) + spacing;}
-        this._setPositionForHierarchy(childNode, pos, childNodeLevel);
+        if (i === 0) {pos = this.direction.getPosition(this.body.nodes[parentId]);}
+        else         {pos = this.direction.getPosition(childNodes[i-1]) + spacing;}
```
mbroad: Would prefer this if else to be spread across more lines, rather than 2 lines

lib/network/modules/LayoutEngine.js
```diff
@@ -1684,55 +1664,96 @@ class LayoutEngine {
...
+          node.y = this.options.hierarchical.levelSeparation * level;
+          node.options.fixed.y = true;
+        },
+
+        shift: (nodeId, diff) => {
+          this.body.nodes[nodeId].x += diff;
+        },
+      };
+    } else {
+      // See the vertical instance for comments.
```
mbroad: An alternative is to use jsdoc `@link` to link to the documentation

lib/network/modules/LayoutEngine.js
```diff
@@ -1684,55 +1664,96 @@ class LayoutEngine {
...
-    //console.log('_setPositionForHierarchy',node.id, position)
-    if (doNotUpdate !== true) {
-      this.hierarchical.addToOrdering(node, level);
-    }
+  setDirectionStrategy() {
+    var isVertical = (this.options.hierarchical.direction === 'UD'
+                   || this.options.hierarchical.direction === 'DU');
+
+    if(isVertical) {
+      this.direction = {
```
mbroad: Imo, it would be better to define a direction interface, and have two fleshed out classes that implement the direction interface. That way the code is self documenting.

lib/network/modules/LayoutEngine.js
```diff
@@ -1741,40 +1762,12 @@ class LayoutEngine {
    */
   _sortNodeArray(nodeArray) {
```
macleodbroad-wf: This doesn't seem very useful anymore, now that the sorting is handled by the strategy

mbroad:
Great idea splitting the logic out into separate strategies! It really cleans up the code nicely!

lib/network/modules/LayoutEngine.js
wimrijnders: No, it's not. `_crawlNetwork()` can be called for either a full network or a single node. In the latter case, param `tree` in `crawler()` is `undefined` and there's no need to set the tree index.

Would you thing it's better if the test is moved outside of `crawler()`?

lib/network/modules/LayoutEngine.js
wimrijnders: agree

lib/network/modules/LayoutEngine.js
wimrijnders: I don't see how I can make this work here; I've kept `direction` intentionally light-weight, so it's not really a separate class (or otherwise) that can be referred to (I think).

lib/network/modules/LayoutEngine.js
wimrijnders: That's not a bad idea. Haven't actually used interfaces in javascript because there's no technical need for it.

Google is telling me there's no such thing, even for >= ES5. It would be a hackish kind of thing.

lib/network/modules/LayoutEngine.js
wimrijnders: Well, only the test on length > 1 I guess. Removing `_sortNodeArray()` would make the `sort`'s in the strategy instances 3 lines longer each, and I wanted to avoid that to keep the sizes down for clarity.

wimrijnders:
You're a maintainer now, you should be able to click on the 'Request changes` button in the Review dialog. You have the right to use this power.

lib/network/modules/LayoutEngine.js
mbroad: I see.

lib/network/modules/LayoutEngine.js
mbroad: Ok

lib/network/modules/LayoutEngine.js
mbroad: Is the micro-optimization of skipping sorts for 1 item lists worth it? I'd just drop this code completely unless it can be shown to have an noon negligible effect on performance.

lib/network/modules/LayoutEngine.js
mbroad: Maybe a typedef describing the expectations of a sort strategy and what it's expected to implement.

mbroad: Ah, here's a good example of an abstract base class that would achieve the same end: https://stackoverflow.com/questions/597769/how-do-i-create-an-abstract-base-class-in-javascript

wimrijnders: OK. I adapted the idea to es5 classes. See how you like it.

lib/network/modules/LayoutEngine.js
wimrijnders: Agree, it's silly. In fact my argument is completely silly. Sorting moved to strategy instances.

wimrijnders:
FYI, I used following command to test the jsdoc generation:

```
> jsdoc -r -p -c ./jsdoc.json lib/network/modules/LayoutEngine.js
```

With the following `jsdoc.json`:

```
{
  "plugins": ["plugins/markdown"]
}
```

I really like the usage of `markdown` in the jsdoc comments. I hope we can standardize on that.

I had to jump through a few hoops to stop jsdoc errors from occurring, it should be obvious in the code (e.g. see `fake_use()`). Also, the conversion to classes was a bit more verbose than I would have hoped. See what you think of it.

lib/network/modules/LayoutEngine.js
```diff
@@ -35,6 +35,242 @@ var NetworkUtil = require('../NetworkUtil').default;
...
+   * @param {Object} layout reference to the parent LayoutEngine instance.
+   */
+  constructor(layout) {
+    super();
+    this.layout = layout;
+  }
+
+  /** @inheritdoc */
+  curveType() {
+    return 'horizontal';
```
wimrijnders: Note the fix here; return value must be perpendicular to 'current' direction.

macleodbroad-wf:
lgtm +1

wimrijnders:
Thanks. I just considered that the Strategy thing has become too big for the module it is in now. I'm considering to move it to a separate module, e.g. `DirectionStrategy.js`. OK by you? and OK to do this in this PR?

macleodbroad-wf:
Sounds great.

wimrijnders:
OK I'll do that then first opportunity. @yotamberk don't merge yet!

wimrijnders:
OK done. This can be merged now.

yotamberk:
Please fix the coflicts and I'll merge =)

PR #3421, opened 2017-09-07
Network: Added default sizes for nodes without labels

wimrijnders:
I noticed during testing that display of nodes without label text has changed[I], specifically the nodes with internal labels. The nodes with external labels are unchanged. In particular, an `ellipse` node can vanish if there is no label text applied:

shapes_prepost

Current PR supplies a basic size for nodes with internal, empty labels.
Note that this is not the smallest size possible; a tiny text of e.g. one character can produce smaller labels.

----

[I]: This is due to changes in the label code, where for an empty text screen width and height are now both calculated to be zero.

### DIFF
--- a/lib/network/modules/CanvasRenderer.js
+++ b/lib/network/modules/CanvasRenderer.js
@@ -149,19 +149,28 @@ class CanvasRenderer {
     // This is not something that will happen in normal operation, but we still need
     // to take it into account.
     //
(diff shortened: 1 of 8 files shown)


wimrijnders:
Note that I sneaked in a bugfix on the height calculation of multifont labels. The essence should be obvious from the code.

Issue #3412, opened 2017-09-03
Edge Endpoint "|"


iXanthos:
Hello, I would like to ask the addition of a new endpoint, the "|".

I saw #1993 , so I will explain briefly where this is used.

In biology, the up-regulation or stimulation is pointed out in the graphs by the use of the standard arrow
"->".
The down-regulation or inhibition, is depicted by the use of this endpoint "-| ".
It is a standard to separate up-regulation from down-regulation.
I've seen the code and it seems pretty easy (for you) to install an extra endpoint, so I hope I gave you a decent reason to add it.

tl;dr: Please add an edgepoint that is used by biologists in their graphs.

Thanks in advance.

wimrijnders:
Γεια σου φίλε! Σπάνια βλέπω Έλληνες εδώ.

OK, the endpoint requests keep on stacking up. Would you mind given a graphical example of how the 'down-regulation' endpoint should look like? I suppose the current arrow is OK for your purposes.

I'll make a point up bumping endpoints to a higher priority. Unfortunately, my time is limited and I am incapable of keeping up with all the issues here, so forgive me if it should take longer than you would wish.

> I've seen the code and it seems pretty easy (for you) to install an extra endpoint,

I actually disagree with this, it's not easy enough. I intend to refactor this code to make it dirt-easy, so that you could have made the PR yourself instead of requesting the issue for this.

wimrijnders:
Related to #3118.

iXanthos:
Γεια σου και σένα!

Yeah, here I have attached what I would like to use. It shows the connection of 2 nodes in a graph. As stated above, the normal arrow is already used in the bibliography to signal the up-regulation, so it cannot be used for the down one as well, since it has to be distinct.

I called it easy, but used the keyword "seems", since I have no idea how the library actually works in its core, so pardon me.

![updownreg](https://user-images.githubusercontent.com/22717082/30005786-d0ac2b04-90f2-11e7-8e20-ef45479687f5.png)

wimrijnders:
OK, that is clear enough. Is the shape of the arrow also important? I mean the gap that is in there.

I'll see what I can do, time permitting. In any case it's on the list.

iXanthos:
No no, the arrow is just your already implemented one, the default. I just added it in order to give you a general idea of what the connections in those kinds of graphs look like. What is missing is the down-regulation part. 

Thanks for the help.

wimrijnders:
[...]

Issue #3118, opened 2017-05-31
Enhance the ends of edges to allow a Crows Foot symbol instead of arrow/circle


EaleyP:
This is a request for enhancement to the edges in networks.
As of now, the ends of edges can only be arrows or circles. Can a feature be added that allows us to choose to put a Crows foot (even a simple single 3 leg symbol would be useful) at the ends of edges instead of the shapes?

bradh:
#3101 about a label variation appears related.

wimrijnders:
This is definitely a feature request. I'm inclined to add this, but unfortunately there are more pressing issues to tackle. So it might take a while. Otherwise, feel free to have a go at it yourself! I'm happy to guide you through it.

For clarity, what kind of symbol do you mean? Something like the following?

![crows_foot](https://cloud.githubusercontent.com/assets/769037/26673555/9673570c-46bd-11e7-995f-625a810fe6d3.png)

PR #3422, opened 2017-09-07
Network: Consolidate code for generating endpoints

wimrijnders:
This is a preliminary refactoring of the endpoint definitions, in preparation for fixing #3412, #1993 and #3118.

All code related to the drawing of arrow types for edges has been moved to new module `EndPoints` and refactored.

The intention is to make it as simple as possible to create new endpoints. My personal hope is that it becomes so simple, that users take it upon themselves to add new endpoints.

There is room for further simplification; any thoughts on this are welcome.

### DIFF
--- a/lib/network/modules/components/edges/util/EndPoints.js
+++ b/lib/network/modules/components/edges/util/EndPoints.js
@@ -4,9 +4,63 @@
  * The intention is to make it as simple as possible to add new endpoint types.
  * As long as the endpoint classes are simple and not too numerous, they will be contained within this module.
  *
- * @typedef {x:number, y:number} Point
- * @typedef {type:string, point:Point, angle:number, length:number} ArrowData
+ * @typedef {{x:number, y:number}} Point
+ * @typedef {{type:string, point:Point, angle:number, length:number}} ArrowData
  */
+
+class EndPoint {
+
+  /**
+   * Apply transformation on points for display.
+   *
+   * The following is done:
+   * - multiply the (normalized) coordinates by the passed length
+   * - rotate by the specified angle
+   * - offset by the target coordinates
+   *
+   * @param {Array<Point>} points
+   * @param {ArrowData} arrowData
+   * @static
+   */
+  static transform(points, arrowData) {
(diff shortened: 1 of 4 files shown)


mbroad:
Looks good

lib/network/modules/components/edges/util/EndPoints.js
```diff
@@ -0,0 +1,177 @@
...
+   * The following is done:
+   * - rotate by the specified angle
+   * - multiply the (normalized) coordinates by the passed length
+   * - offset by the target coordinates
+   *
+   * @param {Array<Point>} points
+   * @param {ArrowData} arrowData
+   * @static
+   */
+  static transform(points, arrowData) {
```
yotamberk: static won't work on browsers that don't support ES6

wimrijnders: Well, we *do* use babel to convert `es2015 syntax` to 'regular' javascript. It's this line in the `package.json`:

```
    "babel-preset-es2015": "^6.24.1",
```

All `es2015` (aka `ES6`) syntax is transpiled to 'regular' javascript upon compilation with `gulp`. To give an example:

```js
class Node {
...
  static parseOptions(parentOptions, newOptions, allowDeletion = false, globalOptions = {}) {
...
  }
};
```

... becomes in the target `vis.js`:

```js
  (0, _createClass3['default'])(Node, [{
...
  }], [{
    key: 'parseOptions',
    value: function parseOptions(parentOptions, newOptions) {
      var allowDeletion = arguments.length > 2 && arguments[2] !== undefined ? arguments[2] : false;
      var globalOptions = arguments.length > 3 && arguments[3] !== undefined ? arguments[3] : {};
...
  }]);
  return Node;
}();
```

`static` just plain disappears in the target file `vis.js` We've been done the transpilation for some time now. I really think we're OK here. 

----

...if I'm missing something, please enlighten me.

yotamberk: Ok, you're right again =)

PR #3431, opened 2017-09-10
The next fix on Travis unit test failure

wimrijnders:
This is the next escalation on the war against the Travis unit tests failing (which came into being by yours truly). By accident, I could recreate the unit test failure on my development machine. This led to a more directed effort to squash the bug.
    
The insight here is that test `(window === undefined)` fails, but `(typeof window === 'undefined')` succeeds. This undoubtedly has to do with the special status `window` has as a global object.
    
Changes:
    
- Added catch-clause in `CanvasRenderer._requestNextFrame()`, fixed local source errors
- Added test on presence `window` in `Canvas._determinePixelRatio()`
- small fix: raised timeout for the network `worldCup2014` unit test

### DIFF
--- a/lib/network/modules/Canvas.js
+++ b/lib/network/modules/Canvas.js
@@ -373,11 +373,19 @@ class Canvas {
      throw new Error("Could not get canvax context");
     }
 
-    return (window.devicePixelRatio || 1) / (ctx.webkitBackingStorePixelRatio ||
+    var numerator = 1;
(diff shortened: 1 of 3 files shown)


macleodbroad-wf:
+1

wimrijnders:
It's conceptually a small difference between the checks, but I was blind to it.

Note that there now is no need for the try/catch, and I'm silently rolling it back in the conflict fixes I'm submitting now for other PR's.

Issue #3408, opened 2017-09-02
network - glyph - node font.multi fontAwesome icon blank


dockstreet:
Hi,

I tried to specify a node from the multifont example with a FontAwesome glyph, but the icon I was trying to use showed up blank? 

```js
{ id: 3, font: { multi: 'html', face: 'FontAwesome' }, label: ' \uf286 <b>This</b> is an\n<i>html</i> <b><i>multi-</i>font</b> <code>label</code>', x: 40, y: 40 },
```

Is it expected or a bug? 

Or perhaps there is another way to combine an image with text in an node without SVG? kind regards!!!

mbroad:
When you resize or interact with the graph in some way, does the icon begin to appear?

dockstreet:
![image](https://user-images.githubusercontent.com/1080803/30061734-91d7adf2-9216-11e7-8493-bc6833d61eb9.png)

wimrijnders:
Frankly, I'm amazed that you got that far. You might be the first to try it.
I don't think this is implemented, but I'll have a look.

sawatts:
Coincidentally I have just been trying to do this myself - setting the italic and bold italic faces to fontawesome so that I can add icons in two different colours to the label along with regular text.

Didn't work for me either.

wimrijnders:
Er. Okay. Uhm. I got it.

![download](https://user-images.githubusercontent.com/769037/30069117-13ccbffc-9260-11e7-864a-705127b5fc5c.png)

Eventually. all I did was upgrade `FontAwesome` from `v4.3.0` to `v4.5.0`:

```html
  <link rel="stylesheet" href="http://maxcdn.bootstrapcdn.com/font-awesome/4.5.0/css/font-awesome.min.css">
```
.... because I accidentally noticed that some icons *were* working.

Go ahead, facepalm. It is that kind of moment.

sawatts:
Not seeing FontAwesome glyphs in multi labels for FA v4.7.0.  They are fine as node icons and in regular HTML constructs.

I had the nodes (option default) `font.ital.face` to `FontAwesome`, `font.ital.mod` to empty string, `font.multi` to `true.  The node-specific label then contains `<i>\uf0c2/i> some other text` for a cloud glyph (amongst others).

Can't say I've used multi-font labels before though - so something may be wrong.  However the "blank boxes" were the correct colour at least.

wimrijnders:
FA 4.7 working for me:

[...]

PR #3442, opened 2017-09-13
Strapping down of Extend-routines in util.js

wimrijnders:
This is internal work, prompted by the investigation in #3408. Submitted here separately in order to keep the PR size(s) down. The reason to do this was to better understand the code, this PR records the results of the investigation for posterity.

**Changes:**
- Some refactoring and fixes of the routines in `util.js` that are used for copying data between objects
- Added (header) commenting for these routines
- Added unit tests for these routines.

- Extra: added unit test for invariability of font properties, because changing these has big consequences for the code in which they are used (paranoia by admission)

**Caveat:** There still is functionality present which might be considered buggy. Please scan in `test/util.test.js` for the word 'bug' and give me your opinion on these. Is it something that should be fixed?

### DIFF
--- a/lib/network/modules/CanvasRenderer.js
+++ b/lib/network/modules/CanvasRenderer.js
@@ -44,13 +44,12 @@ function _initRequestAnimationFrame() {
 let util = require('../../util');
 
 /**
- * @class CanvasRenderer
+ * The canvas renderer
  */
 class CanvasRenderer {
   /**
    * @param {Object} body
    * @param {Canvas} canvas
-   * @constructor CanvasRenderer
    */
   constructor(body, canvas) {
     _initRequestAnimationFrame();
@@ -292,7 +291,7 @@ class CanvasRenderer {
    * Redraw all nodes
    *
    * @param {CanvasRenderingContext2D}   ctx
-   * @param {Boolean} [alwaysShow]
+   * @param {boolean} [alwaysShow]
    * @private
    */
   _resizeNodes() {
@@ -322,7 +321,7 @@ class CanvasRenderer {
    * Redraw all nodes
    *
    * @param {CanvasRenderingContext2D} ctx  2D context of a HTML canvas
-   * @param {Boolean} [alwaysShow]
+   * @param {boolean} [alwaysShow]
    * @private
    */
   _drawNodes(ctx, alwaysShow = false) {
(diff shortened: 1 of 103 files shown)


lib/network/modules/CanvasRenderer.js
```diff
@@ -166,9 +168,10 @@ class CanvasRenderer {
         }
       }
     } catch(e) {
-      console.warning("Got exception with message: '" + e.message() + "'");
-      if (e.message() === 'window is undefined') {
-        console.warning("'" + e.message() + "' happened again");
+      var msg = e.message;
+      //console.warn("Got exception with message: '" + msg + "'");
```
yotamberk: Remove comment

test/util.test.js
```diff
@@ -2,10 +2,345 @@ var assert = require('assert');
...
+
+  it('performs fillIfDefined() as advertized', function () {
+    var a = this.a;
+    var b = this.b;
+
+    util.fillIfDefined(a, b);
+    checkExtended(a, b);
+
+    // NOTE: if allowDeletion === false, null values are copied over!
+    //       This is due to existing logic; it might not be the intention and hence a bug
```
yotamberk: Umm, I don't think it's a good behavior. It does seem like a bug that should be addressed. But on the other-hand, I usually take the approach that if a field is undefined, it should not be assigned a value of `undefined` whilst `null` is a legit value in a field and should be passed on in the extend. If that's the behavior of the function, I'm ok with it.

wimrijnders: I agree with you; for me the trade-off is between fixing it and keeping it like this so as not to break things elsewhere. I'll have a look if I can get rid of it, or confirm that the current working is good.

lib/network/modules/CanvasRenderer.js
wimrijnders: Ah, yes, that one. I hope the travis unit test thing is truly dead now.

wimrijnders: I'm getting rid of the catch-block, it's not required any more.

wimrijnders:
@yotamberk Please check **Caveat** comment in first post. There is stuff I encountered during making unit test of which I'm not sure to fix. It's *almost* legitimate but not consistent.

PR #3451, opened 2017-09-18
Revert "Group redraw performance"

yotamberk:
Reverts almende/vis#3409

this causes a big problem in changing ranged items....
I'll fix this first before remerging


wimrijnders:
Crap. Here is me the reviewer and I let it pass. Sorry (but my knowledge of `timeline` is not enough to have seen this coming.

yotamberk:
It was my fault. I thought i was pedantic about it and only after the merge i found this bug... 
Ill try doing it in stages to understand where i fell

PR #3474, opened 2017-09-22
network - make negative scaleFactor reverse middle arrow correctly

MightyPork:
This fix makes it possible to use negative `scaleFactor` (eg. -1) to flip the middle arrow.

The bug is with this line: `15 * scaleFactor + 3 * lineWidth` - if scale factor is negative, lineWidth is applied in the wrong direction and the arrow gets smaller instead of larger.

Old behavior - wrong scaling when reversed (gets worse when selected):
screenshot_20170922_120556
screenshot_20170922_120606

New behavior - scales correctly
screenshot_20170922_120703

### DIFF
--- a/lib/network/modules/components/edges/util/EdgeBase.js
+++ b/lib/network/modules/components/edges/util/EdgeBase.js
@@ -538,6 +538,7 @@ class EdgeBase {
       }
     }
 
+    if (position === 'middle' && scaleFactor < 0) lineWidth *= -1; // reversed middle arrow
     let length = 15 * scaleFactor + 3 * lineWidth; // 3* lineWidth is the width of the edge.
 
     var xi = arrowPoint.x - length * 0.9 * Math.cos(angle);


wimrijnders:
Hi there! Great of you that you are willing to contribute.

Can you please base the pull request on the `develop` branch?

Also, I'm not really sure how this actually reverses the arrow. I'll wait for the change to `develop` in order to check - but maybe you can explain.

MightyPork:
hmmm it got all messed up when I rebased it to develop, maybe better to start over ??

It's really just the one line.

What reverses the arrow is not this change, but you using negative scale factor. But without my change, the scaling is wrong and the arrow becomes tiny when reversed.

MightyPork:
ah I think I fixed the rebase :)

wimrijnders:
Yes, please make a new PR and close this one. (Even if it is one line!)

MightyPork:
no need now, did a hard reset to develop, cherry-picked my commit and force-pushed it

wimrijnders:
There is still enough room there for me to stand in amazement....

PR #3475, opened 2017-09-22
Improve Item redraw and initial draw performance

yotamberk:
Based on the work done in #3409, I've improved the performance of the timeline redraw time by 2-5 times faster! Initial draw has been reduced to less than 1 seconds for 1,000 items across 10 groups.
default

This PR includes parallel redrawing of items in the same group avoiding "read,write,read,write" reflow/repaint delays of the browser, reducing it to a single repaint per all items' action in redrawing queue.
 
Further work needs to been done for:
- parallel redrawing of items across different groups.
- move the redraw queue of each item type to a general redraw function under the `Item.js` file (since the redraw function is identical to all.)
- Improve even more unneeded repaints occurring in initial drawing of the timeline and redraws called in events preceding initial draw.

### DIFF
--- a/examples/timeline/other/stressPerformance.html
+++ b/examples/timeline/other/stressPerformance.html
@@ -3,8 +3,8 @@
 <head>
   <title>Timeline | Performance example</title>
   <script src="http://cdnjs.cloudflare.com/ajax/libs/moment.js/2.8.4/moment.min.js"></script>
-  <script src="../../dist/vis.js"></script>
-  <link href="../../dist/vis-timeline-graph2d.min.css" rel="stylesheet" type="text/css" />
+  <script src="../../../dist/vis.js"></script>
+  <link href="../../../dist/vis-timeline-graph2d.min.css" rel="stylesheet" type="text/css" />
   
 </head>
 <body>
--- a/examples/timeline/other/stressPerformance.html
+++ b/examples/timeline/other/stressPerformance.html
@@ -1,0 +1,78 @@
+<!DOCTYPE HTML>
+<html>
+<head>
+  <title>Timeline | Performance example</title>
+  <script src="http://cdnjs.cloudflare.com/ajax/libs/moment.js/2.8.4/moment.min.js"></script>
+  <script src="../../dist/vis.js"></script>
+  <link href="../../dist/vis-timeline-graph2d.min.css" rel="stylesheet" type="text/css" />
+  
+</head>
+<body>
+    <div id="visualization"></div>
+</body>
+
+
+<script>
+    var now = moment();
+    var groupCount = 20;
+    var itemCount = 400;
+
+    function genName(length) {
+        var chars = [];
+        for (var i = 0; i<= length; i++) {
+        chars.push(String.fromCharCode(65 + (Math.random()*26)));
+        }
+        return chars.join('');
+    }
+
+    // create a data set with groups
+    var groups = new vis.DataSet();
+    for (var g = 0; g < groupCount; g++) {
+        groups.add({
+        id: g, 
+        content: genName(8)
+        });
+    }
+
+    var types = ['point', 'range', 'box', 'background']
+    // create a dataset with items
+    var items = new vis.DataSet();
+    for (var i = 0; i < itemCount; i++) {
+        var start = now.clone().add(Math.random() * 180, 'days');
+        var end = start.clone().add(Math.random() * 30, 'days');
+        var group = Math.floor(Math.random() * groupCount);
+        items.add({
+        id: i,
+        type: types[Math.floor(Math.random()*types.length)],
+        group: group,
+        content: ''+i,
+        start: start,
+        end: end
+        });
+    }
+
+    // create visualization
+    var container = document.getElementById('visualization');
+    var options = {
+        width: '100%',
+        //autoResize: true,
+        /*
+        margin: {
+        axis: 0,
+        item: 2
+        },
+        */
+        showCurrentTime: false,
+        //locale: 'en',
+        stack: true,
+        selectable: true,
+        editable: true,
+        //groupOrder: 'content',
+    };
+
+    var timeline = new vis.Timeline(container);
+    timeline.setOptions(options);
+    timeline.setGroups(groups);
+    timeline.setItems(items);
+  </script>
+</html>
--- a/examples/timeline/other/stressPerformance.html
+++ b/examples/timeline/other/stressPerformance.html
@@ -1,78 +1,66 @@
 <!DOCTYPE HTML>
+<!-- 
+  This example is used mainly for developers to test performance issues by controlling number of groups, 
+  number of items and their types.
+ -->
 <html>
+
 <head>
-  <title>Timeline | Performance example</title>
+  <title>Timeline | Stress Performance example</title>
   <script src="http://cdnjs.cloudflare.com/ajax/libs/moment.js/2.8.4/moment.min.js"></script>
   <script src="../../../dist/vis.js"></script>
   <link href="../../../dist/vis-timeline-graph2d.min.css" rel="stylesheet" type="text/css" />
-  
 </head>
+
 <body>
-    <div id="visualization"></div>
+  <div id="visualization"></div>
 </body>
 
-
 <script>
-    var now = moment();
-    var groupCount = 20;
-    var itemCount = 400;
+  var now = moment();
+  var groupCount = 20;
+  var itemCount = 400;
 
-    function genName(length) {
-        var chars = [];
-        for (var i = 0; i<= length; i++) {
-        chars.push(String.fromCharCode(65 + (Math.random()*26)));
-        }
-        return chars.join('');
-    }
+  // create a data set with groups
+  var groups = new vis.DataSet();
+  for (var g = 0; g < groupCount; g++) {
+    groups.add({
+      id: g,
+      content: "group " + g
+    });
+  }
 
-    // create a data set with groups
-    var groups = new vis.DataSet();
-    for (var g = 0; g < groupCount; g++) {
-        groups.add({
-        id: g, 
-        content: genName(8)
-        });
-    }
+  var types = ['point', 'range', 'box', 'background']
+  // create a dataset with items
+  var items = new vis.DataSet();
+  for (var i = 0; i < itemCount; i++) {
+    var start = now.clone().add(Math.random() * 180, 'days');
+    var end = start.clone().add(Math.random() * 30, 'days');
+    var group = Math.floor(Math.random() * groupCount);
+    items.add({
+      id: i,
+      type: types[Math.floor(Math.random() * types.length)],
+      group: group,
+      content: '' + i,
+      start: start,
+      end: end
+    });
+  }
 
-    var types = ['point', 'range', 'box', 'background']
-    // create a dataset with items
-    var items = new vis.DataSet();
-    for (var i = 0; i < itemCount; i++) {
-        var start = now.clone().add(Math.random() * 180, 'days');
-        var end = start.clone().add(Math.random() * 30, 'days');
-        var group = Math.floor(Math.random() * groupCount);
-        items.add({
-        id: i,
-        type: types[Math.floor(Math.random()*types.length)],
-        group: group,
-        content: ''+i,
-        start: start,
-        end: end
-        });
-    }
+  // create visualization
+  var container = document.getElementById('visualization');
+  var options = {
+    width: '100%',
+    showCurrentTime: false,
+    stack: true,
+    selectable: true,
+    editable: true,
+  };
 
-    // create visualization
-    var container = document.getElementById('visualization');
-    var options = {
-        width: '100%',
-        //autoResize: true,
-        /*
-        margin: {
-        axis: 0,
-        item: 2
-        },
-        */
-        showCurrentTime: false,
-        //locale: 'en',
-        stack: true,
-        selectable: true,
-        editable: true,
-        //groupOrder: 'content',
-    };
+  var timeline = new vis.Timeline(container);
+  timeline.setOptions(options);
+  timeline.setGroups(groups);
+  timeline.setItems(items);
+</script>
 
-    var timeline = new vis.Timeline(container);
-    timeline.setOptions(options);
-    timeline.setGroups(groups);
-    timeline.setItems(items);
-  </script>
 </html>
--- a/lib/timeline/Timeline.js
+++ b/lib/timeline/Timeline.js
@@ -469,13 +469,29 @@ Timeline.prototype.getItemRange = function () {
     }
     var factor = interval / this.props.center.width;
 
-    // calculate the date of the left side and right side of the items given
-    util.forEach(this.itemSet.items, function (item) {
-      if (item.groupShowing) {
-        item.show();
-        item.repositionX();
-      }
+    var redrawQueue = {};
+    var redrawQueueLength = 0;
 
+    // collect redraw functions
+    util.forEach(this.itemSet.items, function (item, key) {
+      if (item.groupShowing) {
+        var returnQueue = true;
+        redrawQueue[key] = item.redraw(returnQueue);
+        redrawQueueLength = redrawQueue[key].length;
+      }
+    })
+
+    if (redrawQueueLength) {
+      // redraw all regular items
+      for (var i = 0; i < redrawQueueLength; i++) {
+        util.forEach(redrawQueue, function (fns, key) {
+          fns[i]();
+        });
+      }
+    }
+
+     // calculate the date of the left side and right side of the items given
+    util.forEach(this.itemSet.items, function (item) {
       var start = getStart(item);
       var end = getEnd(item);
       var startSide;
--- a/lib/timeline/Timeline.js
+++ b/lib/timeline/Timeline.js
@@ -481,7 +481,8 @@ Timeline.prototype.getItemRange = function () {
       }
     })
 
-    if (redrawQueueLength) {
+    var needRedraw = redrawQueueLength > 0;
+    if (needRedraw) {
       // redraw all regular items
       for (var i = 0; i < redrawQueueLength; i++) {
         util.forEach(redrawQueue, function (fns) {
--- a/lib/timeline/Timeline.js
+++ b/lib/timeline/Timeline.js
@@ -484,7 +484,7 @@ Timeline.prototype.getItemRange = function () {
     if (redrawQueueLength) {
       // redraw all regular items
       for (var i = 0; i < redrawQueueLength; i++) {
-        util.forEach(redrawQueue, function (fns, key) {
+        util.forEach(redrawQueue, function (fns) {
           fns[i]();
         });
       }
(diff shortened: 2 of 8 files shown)


lib/timeline/component/item/BackgroundItem.js
```diff
@@ -49,54 +49,48 @@ BackgroundItem.prototype.isVisible = function(range) {
...
+    background.appendChild(this.dom.box);
   }
   this.displayed = true;
+}
 
-  // Update DOM when item is marked dirty. An item is marked dirty when:
-  // - the item is not yet rendered
-  // - the item's data is changed
-  // - the item is selected/deselected
+BackgroundItem.prototype._updateDirtyDomComponents = function() {
```
wimrijnders: How about retaining the previous comment block as a header comment for this function?

I'm surprised that no header comment passes linting; The linter freaks out if I leave a method uncommented.

lib/timeline/component/item/BackgroundItem.js
```diff
@@ -105,16 +99,73 @@ BackgroundItem.prototype.redraw = function() {
...
+BackgroundItem.prototype._getDomComponentsSizes = function() {
+  // determine from css whether this box has overflow
+  this.overflow = window.getComputedStyle(this.dom.content).overflow !== 'hidden';
+  return {
+    content: {
+      width: this.dom.content.offsetWidth
+    }
+  }
+}
+BackgroundItem.prototype._updateDomComponentsSizes = function(sizes) {
```
wimrijnders: Separate by empty line. I personally prefer two empty lines between functions. This comment applies to the whole file.

lib/timeline/component/item/BackgroundItem.js
```diff
@@ -105,16 +99,73 @@ BackgroundItem.prototype.redraw = function() {
...
 
-    // determine from css whether this box has overflow
-    this.overflow = window.getComputedStyle(dom.content).overflow !== 'hidden';
+  this.dirty = false;
+}
 
-    // recalculate size
-    this.props.content.width = this.dom.content.offsetWidth;
-    this.height = 0; // set height zero, so this item will be ignored when stacking items
+BackgroundItem.prototype._repaintDomAdditionals = function() {
```
wimrijnders: What use is this empty function?

lib/timeline/component/item/BackgroundItem.js
```diff
@@ -105,16 +99,73 @@ BackgroundItem.prototype.redraw = function() {
...
+/**
+ * Repaint the item
+ * @param {boolean} [returnQueue=false]  return the queue
+ * @return {boolean} the redraw queue if returnQueue=true
+ */
+BackgroundItem.prototype.redraw = function(returnQueue) {
+  var sizes
+  var queue = [
+    // create item DOM
+    this._createDomElement.bind(this),
```
wimrijnders: Ah, right. Now the new methods above come together. This is a definitive improvement (wrt readability).

lib/timeline/component/item/BackgroundItem.js
```diff
@@ -105,16 +99,73 @@ BackgroundItem.prototype.redraw = function() {
...
+    // - the item is not yet rendered
+    // - the item's data is changed
+    // - the item is selected/deselected
+    this._updateDirtyDomComponents.bind(this),
+
+    (function() {
+      if (this.dirty) {
+        sizes = this._getDomComponentsSizes.bind(this)();
+      }
+    }).bind(this),
```
wimrijnders: See remark on arrow function `() => {}` elsewhere in this review. For sequences of functions like this, it really helps.

wimrijnders: Oh, crap. I understand now, arrow functions are not present in `timeline` because it uses old syntax.

....but.... AFAIK the babel conversion from ES6 runs over *all code*. I may be wrong.

lib/timeline/component/item/BackgroundItem.js
```diff
@@ -105,16 +99,73 @@ BackgroundItem.prototype.redraw = function() {
...
-    this.props.content.width = this.dom.content.offsetWidth;
-    this.height = 0; // set height zero, so this item will be ignored when stacking items
+BackgroundItem.prototype._repaintDomAdditionals = function() {
+}
 
-    this.dirty = false;
+/**
+ * Repaint the item
+ * @param {boolean} [returnQueue=false]  return the queue
+ * @return {boolean} the redraw queue if returnQueue=true
```
wimrijnders: And if `returnQueue === false`? *Something* is returned there.

lib/timeline/component/item/BoxItem.js
```diff
@@ -58,63 +58,57 @@ BoxItem.prototype.isVisible = function(range) {
...
+    axis.appendChild(this.dom.dot);
   }
   this.displayed = true;
+}
 
-  // Update DOM when item is marked dirty. An item is marked dirty when:
-  // - the item is not yet rendered
-  // - the item's data is changed
-  // - the item is selected/deselected
+BoxItem.prototype._updateDirtyDomComponents = function() {
```
wimrijnders: As with `BackGroundItem.updateDirtyDomComponents()`, please retain commenting. It is useful.

lib/timeline/component/item/BoxItem.js
```diff
@@ -126,41 +120,107 @@ BoxItem.prototype.redraw = function() {
...
+BoxItem.prototype.redraw = function(returnQueue) {
+  var sizes
+  var queue = [
+    // create item DOM
+    this._createDomElement.bind(this),
+
+    // append DOM to parent DOM
+    this._appendDomElement.bind(this),
+
+    // update dirty DOM. An item is marked dirty when:
```
wimrijnders: No, not comment here. I really think that placing it with the method definition is better.

lib/timeline/component/item/BoxItem.js
```diff
@@ -126,41 +120,107 @@ BoxItem.prototype.redraw = function() {
...
-  this._repaintDeleteButton(dom.box);
+  this._repaintDeleteButton(this.dom.box);
+}
+
+/**
+ * Repaint the item
+ * @param {boolean} [returnQueue=false]  return the queue
+ * @return {boolean} the redraw queue if returnQueue=true
+ */
+BoxItem.prototype.redraw = function(returnQueue) {
```
wimrijnders: Identical to `backgroundItem.redraw()`. Any way to DRY this?

lib/timeline/component/item/PointItem.js
```diff
@@ -48,49 +48,109 @@ PointItem.prototype.isVisible = function(range) {
...
     if (!foreground) {
       throw new Error('Cannot redraw item: parent has no foreground container element');
     }
-    foreground.appendChild(dom.point);
+    foreground.appendChild(this.dom.point);
   }
   this.displayed = true;
+}
+
+// PointItem.prototype._updateDirtyDom = function() {
```
wimrijnders: If not used, remove. Otherwise, add motivational comment why this is here.

lib/timeline/component/item/PointItem.js
```diff
@@ -48,49 +48,109 @@ PointItem.prototype.isVisible = function(range) {
...
+//     } else {
+//       this.dom.dot.style.left = (this.props.dot.width / 2) + 'px';
+//     }
+
+//     this.dirty = false;
+//   }
+// }
+
+PointItem.prototype._updateDirtyDomComponents = function() {
+  // An item is marked dirty when:
```
wimrijnders: Make json-style method comment of this.

lib/timeline/component/item/PointItem.js
```diff
@@ -104,40 +164,108 @@ PointItem.prototype.redraw = function() {
...
-  this._repaintDeleteButton(dom.point);
+  this._repaintDeleteButton(this.dom.point);
+}
+
+/**
+ * Repaint the item
+ * @param {boolean} [returnQueue=false]  return the queue
+ * @return {boolean} the redraw queue if returnQueue=true
+ */
+PointItem.prototype.redraw = function(returnQueue) {
```
wimrijnders: Strike three: identical to `BoxItem` and `BackgroundItem`. It's not a question any more: DRY this.

lib/timeline/component/item/BackgroundItem.js
```diff
@@ -105,16 +99,73 @@ BackgroundItem.prototype.redraw = function() {
     // update class
     var className = (this.data.className ? (' ' + this.data.className) : '') +
         (this.selected ? ' vis-selected' : '');
-    dom.box.className = this.baseClassName + className;
+    this.dom.box.className = this.baseClassName + className;
+  }
+}
+BackgroundItem.prototype._getDomComponentsSizes = function() {
```
wimrijnders: Put empty lines between methods. Please review the whole PR for this.

lib/timeline/component/item/BackgroundItem.js
wimrijnders: Why is the empty method needed?

lib/timeline/component/item/RangeItem.js
```diff
@@ -109,27 +104,91 @@ RangeItem.prototype.redraw = function() {
...
+
+    // update dirty DOM. An item is marked dirty when:
+    // - the item is not yet rendered
+    // - the item's data is changed
+    // - the item is selected/deselected
+    this._updateDirtyDomComponents.bind(this),
+
+    (function() {
+      if (this.dirty) {
+        sizes = this._getDomComponentsSizes.bind(this)();
```
wimrijnders: ? Why the bind here? Isn't it bound already (after function def)?

lib/timeline/component/item/RangeItem.js
```diff
@@ -109,27 +104,91 @@ RangeItem.prototype.redraw = function() {
...
+}
+
+
+/**
+ * Repaint the item
+ * @param {boolean} [returnQueue=false]  return the queue
+ * @return {boolean} the redraw queue if returnQueue=true
+ */
+RangeItem.prototype.redraw = function(returnQueue) {
+  var sizes
```
wimrijnders: `;` - If you had DRY'd this before, you'd need to change it only once.

lib/timeline/component/item/RangeItem.js
```diff
@@ -109,27 +104,91 @@ RangeItem.prototype.redraw = function() {
...
   this._repaintDragRight();
+}
+
+
+/**
+ * Repaint the item
+ * @param {boolean} [returnQueue=false]  return the queue
+ * @return {boolean} the redraw queue if returnQueue=true
+ */
+RangeItem.prototype.redraw = function(returnQueue) {
```
wimrijnders: Strike four: Image Samuel L. Jackson holding a gun against your head and what he would say about DRY-ing this.

examples/timeline/other/stressPerformance.html
```diff
@@ -0,0 +1,78 @@
...
+    <div id="visualization"></div>
+</body>
+
+
+<script>
+    var now = moment();
+    var groupCount = 20;
+    var itemCount = 400;
+
+    function genName(length) {
```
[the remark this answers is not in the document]
wimrijnders: Previous comment is misplaced; it should have been in the HTML section.

lib/timeline/Timeline.js
```diff
@@ -489,7 +505,6 @@ Timeline.prototype.getItemRange = function () {
         endSide = end   + (item.getWidthRight() + 10) * factor;
       }
 
-
       if (startSide < min) {
         min = startSide;
         minItem = item;
```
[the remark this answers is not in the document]
yotamberk: Agree. The sign is because I'm subtracting with right while adding with left: 
```
if (this.options.rtl) {
        startSide = start - rhs * factor;
        endSide = end + lhs * factor;
      } else {
        startSide = start - lhs * factor;
        endSide = end + rhs * factor;
      }
```

lib/timeline/component/Group.js
```diff
@@ -210,19 +210,41 @@ Group.prototype._didMarkerHeightChange = function() {
   var markerHeight = this.dom.marker.clientHeight;
   if (markerHeight != this.lastMarkerHeight) {
     this.lastMarkerHeight = markerHeight;
-    util.forEach(this.items, function (item) {
+    var redrawQueue = {};
```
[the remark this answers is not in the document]
yotamberk: I'll do it in another iteration. 
It's a bit different for each case. Let me take care of it in next PRs

lib/timeline/component/Group.js
```diff
@@ -210,19 +210,41 @@ Group.prototype._didMarkerHeightChange = function() {
...
     return true;
   }
 }
 
 Group.prototype._calculateGroupSizeAndPosition = function() {
-  var foreground = this.dom.foreground;
-  this.top = foreground.offsetTop;
-  this.right = foreground.offsetLeft;
-  this.width = foreground.offsetWidth;
+  var offsetTop = this.dom.foreground.offsetTop
```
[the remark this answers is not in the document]
yotamberk: it's not superfluous.
This avoids the browser from read,write,read,write issues. 
If you put together all the read actions (reading properties from the dom), the browser will get these properties in the next frame while avoiding thrashing (going to read a property, changing the dom, reading another property and changing the dom, and again reading a property and changing the dom - each read and write is a reflow - which is heavy).

lib/timeline/component/Group.js
```diff
@@ -237,13 +259,31 @@ Group.prototype._redrawItems = function(forceRestack, lastIsVisible, margin, ran
       // show all items
       var me = this;
       var limitSize = false;
-      util.forEach(this.items, function (item) {
+
```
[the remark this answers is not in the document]
yotamberk: I'l take care of it next PR

lib/timeline/component/Group.js
```diff
@@ -727,14 +767,30 @@ Group.prototype._updateItemsInRange = function(orderedItems, oldVisibleItems, ra
     });
   }
 
-  // finally, we reposition all the visible items.
+  var redrawQueue = {};
```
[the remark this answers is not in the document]
yotamberk: same

lib/timeline/component/item/BackgroundItem.js
```diff
@@ -49,54 +49,48 @@ BackgroundItem.prototype.isVisible = function(range) {
   return (this.data.start < range.end) && (this.data.end > range.start); 
 };
 
-/**
- * Repaint the item
- */
-BackgroundItem.prototype.redraw = function() {
-  var dom = this.dom;
-  if (!dom) {
+BackgroundItem.prototype._createDomElement = function() {
```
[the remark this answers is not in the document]
yotamberk: No good reason really. I'll implement it in a future PR. I much prefer ES6 but I was still confused on what is supported and what is not, and what I can update and what should stay this way,

lib/timeline/component/item/BackgroundItem.js
yotamberk: Ummm I'm not sure how this passes but I assumed if it's not yelling at me, it's ok.

lib/timeline/component/item/BackgroundItem.js
yotamberk: because all the other items contain 6 actions, I wanted to insure that all items have the same number of actions (and that the same function type is done in parallel - this is a preparation to making the redraw queue for all the items defined in the `Item.js` prototype) + it's a space keeper for future implementations of dom additionals to background items (I've been planning to add)

lib/timeline/component/item/BoxItem.js
yotamberk: Yes, in future PRs. This is too difficult right now.

lib/timeline/component/item/PointItem.js
yotamberk: Will be done in future PR

lib/timeline/component/item/RangeItem.js
yotamberk: because `_getDomComponentsSizes` needs the `this`

lib/timeline/component/Group.js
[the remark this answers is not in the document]
wimrijnders: OK, TIL.

lib/timeline/component/item/BackgroundItem.js
```diff
@@ -91,6 +92,10 @@ BackgroundItem.prototype._appendDomElement = function() {
 }
 
 BackgroundItem.prototype._updateDirtyDomComponents = function() {
+  // update dirty DOM. An item is marked dirty when:
+  // - the item is not yet rendered
+  // - the item's data is changed
+  // - the item is selected/deselected
```
wimrijnders: That's actually not what I meant. I meant a comment block like at line 132. Oh well, at least you DRY'd it.

wimrijnders:
As a reviewer, I'm not too happy with the resolution of this issue. I really do think that 4xduplication is a bug.

But I do realize that you're anxious to get this in.....and I already said that the fix is good.
Still, I don't want you to get off that lightly. I'm going to make an issue for you for the DRY'ing, so you don't forget.

yotamberk:
@wimrijnders Thanks.
I understand this still requires more work. It will be DRYed with further PRs

Areson:
@yotamberk @wimrijnders Since I've been working with the `develop` branch post the merging of this issue (for reference, the one I am working with is #3501), I've noticed that this causes the order of stacked items to no longer be consistent. Under the old redraw logic for groups (and in general?) the vertical ordering of items was consistent as the graph was scrolled horizontally. After this change the order of items appears to jump around erratically as you scroll horizontally, which really wrecks the usability of the timeline. I think I've narrowed it down to what causes it, but I don't know why.

Here is the old method. If I revert back to just this (leaving all other changes intact) then the vertical positioning is stable again:

```javascript
util.forEach(this.items, function (item, key) {
        if (!item.displayed) {     
          item.redraw();            
          me.visibleItems.push(item);
        }
        item.repositionX(limitSize);
      });
```

If I make one adjustment to this to move back towards the code from the PR, the vertical position starts jumping around:
```javascript
util.forEach(this.items, function (item, key) {
        if (!item.displayed) {       
          item.redraw();            
          me.visibleItems.push(item);
        }
      })

 for (var i = 0; i < this.items.length; i++) {
    this.items[i].repositionX(limitSize);
 }

yotamberk:
Great! Can you please submit a PR with this revert?

lib/timeline/component/Group.js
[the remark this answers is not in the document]
wimrijnders: [Is this relevat](http://wilsonpage.co.uk/preventing-layout-thrashing/)?

If so, very interesting. This is something that definitely should be checked everywhere in the code, to improve performance.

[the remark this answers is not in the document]
Areson: Cool idea. At least initially it might be worth considering replacing the various places `setTimeout` is used for animation frames with calls to `requestAnimationFrame` before going deeper. For some items though (like #3513) the library is relying on the fact that the DOM will get updated in order for it to perform later calculations for things like stacking. This would still work with the idea from the article but would likely require quite a bit of refactoring.

Issue #3297, opened 2017-07-26
Microsoft Edge error: IndexSizeError


madhukarhere:
Hi,

I am trying to create a graph with image nodes. It is working fine in Firefox and Chrome but not working in IE 11 and IE edge. When I try to debug in IE Edge it is showing the below error

> IndexSizeError

 and pointing to below code in line 35466 in Vis.js
`ctx.drawImage(this.canvas, from[0], from[1], from[2], from[3], _to[0], _to[1], _to[2], _top[3]).`

Below is the configuration section:
```
var container = document.getElementById("groups");
    var data = {
        nodes: nodes,
        edges: edges
    };
    var options = {
        interaction: {
            hover: true,
            tooltipDelay: 10,
            dragNodes: false,// do not allow dragging nodes
            zoomView: true, // do not allow zooming
            dragView: true  // do not allow dragging
        },
        nodes: {size:30},
        edges: {
            width: 2,
            shadow: true,
            color: edge_color
        },
        groups: {
            orgs: {
                font: {
                    size: 16,
                    color: "#000000"
                }
                shadow: {enabled: true, borderRadius: 35},
            },
            groups: {
                shape: 'icon',
                icon: {
                    face: 'FontAwesome',
                    code: '\uf007',
                    size: 80,
                    color: '#aa00ff'
                },
                shadow: {enabled: true},
                size: 35,
                font: {
                    size: 16,
                    color: "#000000"
                },
            },
            sub_groups: {
                shape: 'icon',
                icon: {
                    face: 'FontAwesome',
                    code: '\uf007',
                    size: 35,
                    color: '#aa00ff'
                },
                shadow: {enabled: true},
                size: 40,
                font: {
                    size: 16,
                    color: "#000000"
                },
            },
            more:{
                shape: "circle",
                font: {
                    size: 45
                },
                borderWidth: 2,
                shadow: {enabled: true},
                color: get_rand_color()
            }
        },
        physics:{barnesHut:{gravitationalConstant:-2000}},

    };

    network = new vis.Network(container, data, options);

    network.on("click", function (params) {
        params.event = "[original event]";
        var s_nodes = params["nodes"];
        var pointer = params["pointer"];
        var dom_cords = pointer["DOM"];
        if(s_nodes[0] !== "more_"+group_id){
            generate_toop_tip(s_nodes[0],dom_cords["x"], dom_cords["y"]);
        } else{
            remove_tooltip();
            show_more_orgs(group_id);
        }
    });

[...]

PR #3477, opened 2017-09-22
IE11 svg image fixes

### LINKED ISSUE #3356: IE11 Bug? Network: Using svg as "image" doesn't work in all zoom levels
There seems to be a bug in IE11 when using the network with shape 'image' and svgs.
In IE11, the image is not visible initially, only after zooming out to a very high zoom level do the images pop up.

Here is a fiddle underlining the problem:
https://jsfiddle.net/zzhb8wkb/7/

Any ideas?

justinharrell:
fixes #3356 and #3297

IE would get IndexSizeErrors from drawImage from fractional heights or widths, now they are floored to avoid the issue.

Also _isImageOk was preventing SVG images from loading in IE due to naturalWidth being 0 just as width is. This can't be fixed up like width can because naturalWidth is read only. _isImageOk seems to be unecessary since Image.load handles onload and on error first so no need to double check if the image is ok.

### DIFF
--- a/lib/network/CachedImage.js
+++ b/lib/network/CachedImage.js
@@ -36,17 +36,28 @@ class CachedImage {
     this.width  = w;
     this.height = h;
 
+    var h2 = Math.floor(h/2);
+    var h4 = Math.floor(h/4);
+    var h8 = Math.floor(h/8);
+    var h16 = Math.floor(h/16);
+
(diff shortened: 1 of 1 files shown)


ipavlic:
When will this be available in a version on npm?

Issue #2677, opened 2017-02-05
Network clustering: Cannot read property 'font' of undefined


AoDev:
Hi, I have basically copied the example of clustering [here](https://github.com/almende/vis/blob/master/examples/network/other/clustering.html).

Just using my own data.

when initiating the graph
```
var container = document.getElementById('#network')
this.graph = new vis.Network(container, {}, graphOptions)
```

When getting the data (before, I add a group property to each node)
```
    this.graph.setData(graphData)

    var groups = ['group1', 'group2', 'group3']
    var clusterOptionsByData
    for (var i = 0; i < groups.length; i++) {
      var group = groups[i]
      clusterOptionsByData = {
        joinCondition (childOptions) {
          return childOptions.group === group
        },
        clusterNodeProperties: {
          id: 'cluster:' + group,
          borderWidth: 3,
          shape: 'dot',
          color: '#000',
          label: 'group:' + group,
          font: {
            color: '#000',
            size: 14
          }
        }
      }

      this.graph.cluster(clusterOptionsByData)
    }
```

My graph config
```
const graphOptions = {
  physics: false,
  edges: {
    arrows: 'to',
    physics: false,
    scaling: {
      min: 1,
      max: 5,
      label: {
        enabled: false
      }
    },
    smooth: {
      enabled: false,
      type: 'cubicBezier',
      forceDirection: 'vertical',
      roundness: 0.5
    }
  },
  nodes: {
    shape: 'dot',
    scaling: {
      min: 10,
      max: 30,
      label: {
        enabled: false
      }
    },
    font: {
      strokeWidth: 2,
      strokeColor: '#fff',
      multi: 'html'
    }
  },
  layout: {
    improvedLayout: true,
    hierarchical: {
      enabled: true,
      levelSeparation: 150,
      nodeSpacing: 200,
      treeSpacing: 200,
      blockShifting: true,
      edgeMinimization: true,
      parentCentralization: true,
      direction: 'UD',        // UD, DU, LR, RL
      sortMethod: 'directed'   // hubsize, directed
    }
  }
}
```

Stacktrace:
```
TypeError: Cannot read property 'font' of undefined
    at Label.propagateFonts (vis.js:34775)
    at Cluster.setOptions (vis.js:34226)
    at Cluster.Node (vis.js:34126)
    at new Cluster (vis.js:43598)
    at NodesHandler.create (vis.js:33809)
    at ClusterEngine._cluster (vis.js:42959)
[...]

PR #3486, opened 2017-09-26
Network: Fix handling of multi-fonts 

### LINKED ISSUE #3408: network - glyph - node font.multi fontAwesome icon blank

wimrijnders:
Fixes #3408, #2677.

This PR serves to get the handling of the font options right. The fixing of the issue is more or less a side-effect. As a special note, with this PR the group options for fonts are finally handled correctly.

Guiding principles:
- Option values should *not* be changed inline. They should remain as entered.
- Priority order for options should be respected: `node, group, default node, global default`.

## Changes
- Got rid of `Label#parseOptions()`. This changes the option values, and we want to keep them as
  specified.
- Adjusted usage of groups, at least for the font options in groups.

  Group font options were merged into the options of a given node instance, and the label
  handling requires these to be separate. *In fact, it is questionable if the group options
  should be merged into the node instance options at all.*

- Unit tests for the setting of font options in all manners have been added.
- **Special Note:** The added unit tests picked up an unrelated bug in `util.deepExtend()` and `util.selectiveDeepExtend()`;  added an extra check on `null` there.

### Further changes

- Removed member `edgeOptions` in module `EdgesHandler`, it's just a reference to `this.options`
- Removed member `nodeOptions` in module `NodesHandler`, it was useless for the same reason
- Added/edited commenting extensively
- Completely rewrote `Label.propagateFonts()`, additional notes :
  - Special handling in the old code of multi-font `mono` has been dropped; if users want to
  have a non-proportional font for `mono`, that's their problem.
  - Special handling of `ratio` has been dropped. This does not have any effect on the outcome. The value of `ratio` was always `1` in all the cases I have conceived to test.

## Final Comment

If anything is to be concluded from this fix, it's that the option setup and handling within
`Network` is unnecessarily complicated. This urgently needs to be looked at.

In particular, the **prototype chaining of options appears to be deficient**. While the idea is 
elegant in itself, the prototype order is not sufficient to handle all option sequences; in fact,
the code is rife with exceptions on it. Prototype chaining needs to be replaced with something
probably less elegant but better suited to the task.


lib/network/modules/Groups.js
```diff
@@ -75,14 +75,17 @@ class Groups {
...
-   * @return {Object} group      The created group, containing all group options
+   * Get group options of a groupname.
+   * If groupname is not found, a new group may be created.
+   *
+   * @param {*}       groupname     Can be a number, string, Date, etc.
+   * @param {boolean} [create=true] If true, create a new group
+   * @return {Object} The found or created group
    */
-  get(groupname) {
+  get(groupname, create = true) {
```
yotamberk: I personally like more declarative variables. I think it should be: `create` => `shouldCreate`

lib/network/modules/NodesHandler.js
```diff
@@ -1,9 +1,8 @@
 let util = require("../../util");
 let DataSet = require('../../DataSet');
 let DataView = require('../../DataView');
-
 var Node = require("./components/Node").default;
-var Label = require("./components/shared/Label").default;
+
```
yotamberk: extra line

lib/network/modules/components/shared/Label.js
```diff
@@ -178,141 +202,178 @@ class Label {
...
 
+   /**
+    * Return all options values for the given multi-font.
+    *
+    * All available option objects are trawled in the set order to construct the option values.
+    *
+    * @param {MultiName} multiName sub path for the mod-font
+    * @returns {ModOptions}
+    */
+    var getFontOptions = (multiName) => {
```
yotamberk: Why do you use `var`?  Why not use `const` and `let`? 
(for consistency)

lib/network/modules/components/shared/Label.js
```diff
@@ -584,7 +645,14 @@ class Label {
         }
       }
     }
-    ctx.font = (values.mod + " " + values.size + "px " + values.face).replace(/"/g, "");
+
+    var fontString = "";
+    if (values.mod !== undefined && values.mod !== "") {  // safeguard for undefined - this happened
+      fontString += values.mod = " ";
```
yotamberk: I think you mean `fontString += values.mod + " "` ? It seems really weird to me if you meant to assign `values.mod = " "` and simultaneously add a space to `fontString`

lib/network/modules/components/shared/Label.js
wimrijnders: A quick response on this one in passing:

I prefer `var` because chrome has issues dealing with `const` and `let`, it can't seem to display the values correctly inline (usually shows `undefined` even though it has a value). I find this bothersome to the point of forsaking them, even though `let` would be better.

If you have a solution for this, let me know.

yotamberk: But it seems that you douse `let` in the rest of the file. How does that work?

wimrijnders: I really try to use `let`. But if it gets in the way of debugging, I revert it to `var`.

lib/network/modules/Groups.js
wimrijnders: :+1:

lib/network/modules/NodesHandler.js
wimrijnders: You mean that there are two empty lines? I actually prefer this, for better visual separation. I do it between methods as well, and in other places where I think a bigger distinction between blocks is useful.

lib/network/modules/components/shared/Label.js
wimrijnders: Oh well, I can change this since I'm done with debugging.

lib/network/modules/components/shared/Label.js
wimrijnders: Indeed. Good catch.

lib/network/modules/components/shared/Label.js
```diff
@@ -178,141 +202,178 @@ class Label {
...
-     * @return {*|null} value if found, null otherwise.
-     */
-    var getP = function(options, mod, property) {
-      let opt = pathP(options, mod);
+   /**
+    * Collect all own-property values from the font pile that aren't multi-font option objectss.
+    *
+    * @returns {object} object with all current own basic font properties
+    */
+    let getBasicOptions = () => {
```
mbroad: I'm seeing a lot of nested function calls and wondering if it is a micro optimization to bring them up a level, so that they're initialized with the class, rather than initialized with every call to the function that encapsulates them.

test/Label.test.js
```diff
@@ -418,6 +435,686 @@ describe('Network Label', function() {
...
+          bold: {
+            color: 'black'
+          }
+        }
+      },
+    });
+
+    assert.equal(modBold(0).color, 'black');    // nodes default
+    assert.equal(modBold(1).color, 'black');    // more specific bold value overrides group value
+    assert.equal(modBold(2).color, 'black');    // idem
```
mbroad: Idem?

mbroad: Why are these three changed, and the others aren't? I'm missing something here...

test/Label.test.js
```diff
@@ -418,6 +435,686 @@ describe('Network Label', function() {
...
+    assert.equal(modBold(1).color, 'black');    // more specific bold value overrides group value
+    assert.equal(modBold(2).color, 'brown');    // bold group value overrides bold node value
+    assert.equal(modBold(3).color, 'green');    // unchanged
+    assert.equal(modBold(4).color, 'orange');   // unchanged
+
+
+    //
+    // Same initialization as previous with a color set for the default node font
+    //
+    data.nodes = new vis.DataSet(dataNodes);  // Need to reset nodes, changed in previous
```
mbroad: This isn't really a reset is it?

It's a new dataset, but the since dataNodes is a mutable array of mutable objects, and not a function who's return value is a new array of objects, the nodes have not been reset.

test/Label.test.js
```diff
@@ -418,6 +435,686 @@ describe('Network Label', function() {
...
+    // console.log(fontOption(1));
+
+    assert.equal(modBold(1).color, '#343434');  // Reverts to default 
+    assert(!fontOption(1).multi);               // idem 
+    assert.equal(modBold(6).color, 'blue');     // unchanged
+    assert(fontOption(6).multi);                // unchanged
+
+    done();
+  });
+
```
mbroad: Great addition of tests! Well done!

I would suggest, however, refactoring these very large tests into a number of much smaller tests, in separate test classes where required, and making use of `before` and `after` to set up the state you need.

wimrijnders:
@mbroad review :+1, you're being sharp.

lib/network/modules/NodesHandler.js
```diff
@@ -144,7 +142,9 @@ class NodesHandler {
       throw 'Internal error: mass in defaultOptions of NodesHandler may not be zero or negative';
     }
 
-    util.extend(this.options, this.defaultOptions);
+    //this.options = {};
+    //util.extend(this.options, this.defaultOptions);  // Candidate for the Hall of Shame!
```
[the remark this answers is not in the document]
wimrijnders: Heh. For the epitaph, actually. Major stupidity was performed here. I do this often in my own code, so that I remember not to do that thing every time I encounter the comment.

But this is shared code, and keeping it in is just vanity. Hereby gone.

lib/network/modules/components/shared/Label.js
```diff
@@ -2,11 +2,24 @@ let util = require('../../../../util');
 let ComponentUtil = require('./ComponentUtil').default;
 let LabelSplitter = require('./LabelSplitter').default;
 
+/**
+ * @typedef {'bold'|'ital'|'boldital'|'mono'|'normal'} MultiName
+ *
```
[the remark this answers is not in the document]
wimrijnders: Ok, although I do want to convey that this is the list of names of special fonts that can be used when multi-font is enabled.

Will you settle for `MultiFontStyle`?

lib/network/modules/components/shared/Label.js
```diff
@@ -2,11 +2,24 @@ let util = require('../../../../util');
...
+ * The allowed specifiers of multi-fonts.
+ */
+
+/**
+ * @typedef {{color:string, size:number, face:string, mod:string, vadjust:number}} ModOptions
```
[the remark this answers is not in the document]
wimrijnders: :+1:

lib/network/modules/components/shared/Label.js
```diff
@@ -16,59 +29,69 @@ class Label {
...
+   * Init the font Options structure.
    *
-   * @param {Object} parentOptions
-   * @param {Object} newOptions
-   * @param {boolean} [allowDeletion=false]
-   * @static
+   * Member fontOptions serves as an accumulator for the current font options.
+   * As such, it needs to be completely separated from the node options.
+   *
+   * @param {Object} options
```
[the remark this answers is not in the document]
wimrijnders: Other options, whatever the user specified or otherwise the defaults.

lib/network/modules/components/shared/Label.js
```diff
@@ -16,59 +29,69 @@ class Label {
...
+   * @private
    */
-  static parseOptions(parentOptions, newOptions, allowDeletion = false) {
-    if (Label.parseFontString(parentOptions, newOptions.font)) {
-      parentOptions.vadjust = 0;
-    }
-    else if (typeof newOptions.font === 'object') {
-      util.fillIfDefined(parentOptions, newOptions.font, allowDeletion);
+  initFontOptions(options) {
```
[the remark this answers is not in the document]
wimrijnders: Well, sort of. The underscore prefix was already present when I first came here.
There's no written convention about it. I would personally like them to signify 'methods internal to the project', but I haven't even proposed this.

The docs are vague when it comes to programming conventions. This is actually something that needs to be addressed (although, now, for many things you can just say 'it needs to pass linting').

lib/network/modules/components/shared/Label.js
```diff
@@ -16,59 +29,69 @@ class Label {
...
+    let newFont = options.font;  // Only this member used!
+
+    // Get the structure right
+    util.forEach(options.font, (prop, n) => {
+      if (typeof prop === 'object') {
```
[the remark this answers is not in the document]
wimrijnders: Er. Oops. Seldom does it occur that someone's paranoia overshadows mine.

I'm rewriting this to use the `MultiFontStyle` list as in the jsdoc typedef. We know exactly which multi-fonts should be there, might as well use this knowledge.

lib/network/modules/components/shared/Label.js
```diff
@@ -178,141 +202,178 @@ class Label {
...
+   *   - search in option group 'bold' in local properties
+   *   - search in main font option group in local properties
+   *
+   * ---------------------------------------------------------------------
+   * @param {Array.<object>} pile  sequence of option objects to consider.
+   *                               First item in list assumed to be the newly set options.
    */
-  propagateFonts(options, groupOptions, defaultOptions) {
-    if (!this.fontOptions.multi) return;
+  propagateFonts(pile) {
```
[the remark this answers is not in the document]
wimrijnders: Every time the options or data change. In the regular flow, it shouldn't be called at all (which is something that I should actually confirm)

lib/network/modules/components/shared/Label.js
```diff
@@ -178,141 +202,178 @@ class Label {
...
+    let fontPile = [];   // sequence of font objects to consider, order important
+    let mods = ['bold', 'ital', 'boldital', 'mono'];
```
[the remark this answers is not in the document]
wimrijnders: :+1:

lib/network/modules/components/shared/Label.js
wimrijnders: OK. My long-term planning is to isolate this pile handling code in a separate class. This is the approach that I want to use to replace the prototype chaining in options. I thought it was too early to make a class of it now.

But fair enough, I can define them as private methods for the time being. Wrt. 'micro optimization', it's an easy task to raise them to class-level, and every teeny bit helps.

lib/network/modules/components/shared/Label.js
```diff
@@ -178,141 +202,178 @@ class Label {
...
       }
 
+      return result;
+    }
+
+
+    //
+    // Main body of the method
+    //
+
```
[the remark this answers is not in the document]
wimrijnders: ok. The idea was to indicate where the actual function code began, after the definitions of the local function. But I just moved them away, so....

test/Label.test.js
wimrijnders: ['Idem'](http://www.dictionary.com/browse/idem) : Same as previous comment in this column.

wimrijnders: Overview:
- The priority of the options are:  `node, group, default node, global default`.
- `black` is in 'default node'.
- multi-font options, if defined, take precedence. The main font options are fall-backs for the multi-font options.

Specific:
- Node 0 has no 'node' or 'group' options, hence uses 'default node'.
- Nodes 1 has `group2`, which has a default font but no explicit bold font. The bold definition in 'default node' trumps this.
- Node 2 has `group1`, the same logic as with Node 1 applies here.
- Nodes 3 and 4 have bold fonts defined in the node instance. These trump everything.

Welcome to the wonderful world of multi-font options. Precisely this kind of thing is what made fixing the issue so hard. My head still hurts.

test/Label.test.js
wimrijnders: Yes, but `dataNodes` is not changed anywhere, so the same original values are being entered in the `DataSet` instance, and `DataSet` copies values internally.

Crap, let me confirm that.

test/Label.test.js
wimrijnders: I guess; they're kind of lengthy, that's true. I'll have a think about how to carve this up.

lib/network/modules/components/shared/Label.js
```diff
@@ -2,11 +2,44 @@ let util = require('../../../../util');
...
+ *  - prototypes are totally ignored. The idea is to add option objects used as prototypes
+ *    to the pile, in the correct order.
+ */
+
+
+/**
+ * List of special styles for multi-fonts
+ * @private
+ */
+const multiFontStyle = ['bold', 'ital', 'boldital', 'mono'];
```
mbroad: `normal` is omitted?

test/Label.test.js
mbroad: Excellent! The only other suggestion I'll make is to follow the mocha convention for `pending`, or write the entire test and use `skip`, as described here: https://mochajs.org

This is a nit pick, and I'm approving as-is, with or without my suggestion.

wimrijnders:
`pending` and `skip` conventions are new to me. I suppose the intention would be to get rid of the TODO's at the top of the unit test file(s)?

OK, I think this is a minor thing. I will leave it as is, but thanks for pointing it out.

wimrijnders:
@justinharrel, since your comment on `widthConstraint` appears to have been deleted, I take it that the wrapping is not an issue? Just checking.

justinharrell:
I accidentally commented on commit rather than PR, its still there. widthConstraint still does not seem to work.

Should I just make a new issue?

wimrijnders:
What another failing case? Sure, make a new issue. If you can give a test-case that would be welcome.

PR #3488, opened 2017-09-26
Network: Adjust documentation for arrows.middle.scaleFactor

wimrijnders:
Addendum to #3474.

Updated the documentation, so that users can know about flipping the middle arrow with a negative scale factor.

### DIFF
--- a/docs/network/edges.html
+++ b/docs/network/edges.html
@@ -285,7 +285,9 @@ network.setOptions(options);
             <td class="indent">arrows.middle</td>
             <td>Object or Boolean</td>
             <td><code>Object</code></td>
-            <td>Exactly the same as the to object but with an arrowhead in the center of the edge.</td>
+            <td>In general, the same as the to object but with an arrowhead in the center of the edge.
+                The only difference is that the direction of the arrow can be flipped by using a negative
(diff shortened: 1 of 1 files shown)


docs/network/edges.html
```diff
@@ -285,7 +285,9 @@
             <td class="indent">arrows.middle</td>
             <td>Object or Boolean</td>
             <td><code>Object</code></td>
-            <td>Exactly the same as the to object but with an arrowhead in the center of the edge.</td>
+            <td>In general, the same as the to object but with an arrowhead in the center of the edge.
+                The only difference is that the direction of the arrow can be flipped by using a negative
+                value for <code>arrows.middle.scaleFactor</code>.</td>
```
mbroad: Consider dropping superfluous wording and wrapping the word to in single quotes:
```
<td>Similar to the 'to' object, but with an arrowhead in the center of the edge. The direction of the arrow can be flipped by using a negative value for <code>arrows.middle.scaleFactor</code>.</td>
```

mbroad:
One nit. But this is fine as is if you don't like my suggestion.

wimrijnders:
I'm going with your suggestion; documentation should be as clear and concise as feasible.

Issue #3425, opened 2017-09-07
MS IE and Edge slow when stabilizing is finished


Head:
For example, open http://visjs.org/examples/network/exampleApplications/lesMiserables.html on Edge (latest version 40.15063.0.0).
Then move the whole graph: Utterly smooth.
Now wait a bit until all nodes are stabilized in their positions.
Move again and have a bad time.

OACDesigns:
I'm also experiencing this on Edge 15.15063
It becomes virtually unusable as soon as stabalisation has completed.

wimrijnders:
So I've got a Window 10 machine running here now, to check this and other issue on Edge. It's a 1.4GHz Celeron, so not the fastest thing on the planet.

I ran the given example on Edge and in all honesty I have to say it's working quite well. I do notice, however, a difference with the usage of input devices:

- The touch pad on my Logitech keyboard works smoothly as a baby's bottom. I can drag to my heart's content and all is responsive.
- With the USB mouse, however, there is a significant lag while dragging, with the dragged node chasing the mouse cursor in discrete steps. As soon as I stop dragging, everything is smooth again.

Could this perhaps be a driver problem?

I'm wondering now what the symptons are that you are experiencing. Does it look like what I experienced with the USB mouse? Or is it something else more pronounced? What particular experience makes it virtually unusable and/or gives me a bad time?

I've also tried the `worldCupPerformance` example, with is a heavy-weight with about 1000 nodes and 10000 edges. Here, I do see significant difference with my i7. But I think that that can be forgiven with such a big network.

----
Stabilization, as the term is known within `Network`, is the step that is performed *before* anything is shown on screen, to assign decent starting positions for the nodes. I'm assuming you don't mean that, but that the nodes have stopped moving on screen.

wimrijnders:
@OACDesigns, how do I check the Edge version number? Have no clue and Bing isn't helping.

Head:
Yep, the moment the nodes have stopped moving it becomes laggy. Using a normal mouse and a high end gamer PC, latest "everything".

I made a video to demonstrate it (also showing where to find the version no): https://youtu.be/_xSYcFyVnGw

wimrijnders:
Thanks for the video, that helped.

What I'm seeing is that movement is smooth as long as the nodes are moving. When the movement stops, the full network drag is 'stuttering' (to give it a name). There are large pauses between movement and that movement is abrupt.

I also note:
- zoom in/out is responsive
[...]

PR #3489, opened 2017-09-26
IE performance improvements #3425

justinharrell:
So the main issue with #3425 is onDrag was calling the synchronous _redraw instead of the aysnc _requestRedraw like the other interaction events. This didn't seem to be issue in Chrome but in IE it destroyed frame rate when dragging the view around.

The reason it didn't not cause an issue when physic simulation was still running is because the _redraw handler in CanvasRenderer does not actually call redraw if rendering is active.

The other changes to Label.js improve performance across the board based on profiling in IE and Chrome, which is where I started with the issue. I also cleaned up some code that was simply not correct and made it a little better for possible further optimizations in the future.

### DIFF
--- a/lib/network/modules/InteractionHandler.js
+++ b/lib/network/modules/InteractionHandler.js
@@ -393,7 +393,7 @@ class InteractionHandler {
         let diffY = pointer.y - this.drag.pointer.y;
 
         this.body.view.translation = {x:this.drag.translation.x + diffX, y:this.drag.translation.y + diffY};
-        this.body.emitter.emit('_redraw');
+        this.body.emitter.emit('_requestRedraw');
       }
     }
   }
--- a/lib/network/CachedImage.js
+++ b/lib/network/CachedImage.js
@@ -11,7 +11,6 @@
  * @param {Image} image
  */
 class CachedImage {
-  constructor() {  // eslint-disable-line no-unused-vars
   /**
    * @ignore
    */  
--- a/lib/network/CachedImage.js
+++ b/lib/network/CachedImage.js
@@ -11,6 +11,7 @@
  * @param {Image} image
  */
 class CachedImage {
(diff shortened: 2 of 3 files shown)


wimrijnders:
:heart: Thanks for looking into this!

lib/network/modules/InteractionHandler.js
```diff
@@ -393,7 +393,7 @@ class InteractionHandler {
         let diffY = pointer.y - this.drag.pointer.y;
 
         this.body.view.translation = {x:this.drag.translation.x + diffX, y:this.drag.translation.y + diffY};
-        this.body.emitter.emit('_redraw');
+        this.body.emitter.emit('_requestRedraw');
```
wimrijnders: That's it???? This resolves the IE issue??? Wow....

wimrijnders:
Esteemed comitter, your changes and fixes are totally and uniformly excellent. I am exceedingly impressed.

There's only one thing that prevents me from immediately approving, and that is that the PR appears to be building on an old version of the code. Would you mind merging with latest develop and see what changes are needed?

Sincerely, thank you for your efforts.

justinharrell:
Is that any better? The 0700f19da436ced753a6fddf249f660afcd64173 can be ignored if thats the issue, it was a merge artifact fixed up in 50db76fbc2cbf9e47671c71945000bf7c21b26da

wimrijnders:
I found the problem: it's that I'm an idiot. I was comparing to a branch that hasn't been merged yet (#3486).

That means that I consider your changes to be perfect, I can't find a fault in them. It also means that one of use will have to do merge+conflicts soon. Probably me, because I'm hereby approving this.

wimrijnders:
@mbroad That's a fair question. I'll scan the code to see if it makes sense to convert to `_requestRedraw()` elsewhere.

PR #3492, opened 2017-09-28
Network: Prevent image loading for hidden cluster nodes

### LINKED ISSUE #3483: Option image must be defined for node type 'image'
Hi vis.js team,

I think I found a bug with the latest version of the library (4.20.1).

When you render nodes of 'image' shape while leaving the shape definition globally set on options, it results in the error:
Option image must be defined for node type 'image', line 36469.

This happens with more than 150 nodes.

I can confirm this was introduced in 4.20.0. The 4.19.1 works fine, and 4.20.1 is still broken.

Here is the working example (with version 4.19.1):
https://jsfiddle.net/os1zn7qz/

And here is the same with the newer library (4.20.1), which fails to load:
https://jsfiddle.net/b1gc1qe4/

I can get the version 4.20.1 to work if I remove the global setting and place it instead within the node setting:
https://jsfiddle.net/oe9qvLsj/

The code (in case Fiddle fails):

```html
<html>

<head>
    <link href="https://cdnjs.cloudflare.com/ajax/libs/vis/4.20.1/vis.min.css" rel="stylesheet" />
    <script src="https://cdnjs.cloudflare.com/ajax/libs/vis/4.20.1/vis.min.js"></script>
    <style>
    #network_visualisation {
        height: 100%;
        width: 100%;
    }
    </style>
</head>

<body>
    <div id="network"></div>
    <script>
    var limit = 15;
    var nodes = [],
        edges = [];
    for (var i = 0; i < limit; i++) {
        nodes.push({
            "id": "id" + i.toString(),
            "label": "node" + i.toString(),
            // "shape": "image",
            "image": "https://cdn4.iconfinder.com/data/icons/iconsimple-logotypes/512/app_store-128.png",
        });
    }

    for (var i = 0; i < limit; i += 2) {
        var randId = "id" + (Math.floor(Math.random() * 99) + 1).toString();
        edges.push({
            "to": randId,
            "from": "id" + i.toString(),
            "label": "edge" + i.toString()
        })
    }

    var network = new vis.Network(document.getElementById('network'), {
        nodes: new vis.DataSet(nodes),
        edges: new vis.DataSet(edges)
    }, {
        nodes: {
            shape: 'image',
        },
    });
    </script>
</body>
</html>
```

With the sample code above:
- as is - 4.20.1 breaks
- change to 4.19.1 - works fine
- 4.20.1 - add "shape" property to each node and remove it from options

wimrijnders:
Fixes #3483.

If `shape: 'image'` was defined as option, this was used for initializing hidden cluster nodes during layout initialization. However, because no images were being passed, this led to an exception.

This fix prevents shape `image` from being used for hidden cluster nodes. In addition, some extra fields for these nodes are now overridden for some performance improvement.

### DIFF
--- a/lib/network/modules/LayoutEngine.js
+++ b/lib/network/modules/LayoutEngine.js
@@ -561,13 +561,32 @@ class LayoutEngine {
       if (positionDefined < 0.5 * indices.length) {
         let MAX_LEVELS = 10;
         let level = 0;
-        let clusterThreshold = 150;
-        // Performance enhancement, during clustering edges need only be simple straight lines.
+        let clusterThreshold = 150;  // TODO add this to options
+
+        //
+        // Define the options for the hidden cluster nodes
         // These options don't propagate outside the clustering phase.
+        //
+        // Some options are explicitly disabled, because they may be set in group or default node options.
+        // The clusters are never displayed, so most explicit settings here serve as performance optimizations.
+        //
+        // The explicit setting of 'shape' is to avoid `shape: 'image'`; images are not passed to the hidden
+        // cluster nodes, leading to an exception on creation.
+        //
+        // All settings here are performance related, except when noted otherwise.
+        //
         let clusterOptions = {
+          clusterNodeProperties:{
+            shape: 'ellipse',       // Bugfix: avoid type 'image', no images supplied
+            label: '',              // avoid label handling
+            group: '',              // avoid group handling
+            font: {multi: false},   // avoid font propagation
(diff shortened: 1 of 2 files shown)


mbroad:
No tests?

wimrijnders:
No, I don't regard this as necessary; this is something that will never go wrong again.

This is the part where you have authority as a reviewer: if you really do not agree, you can put your foot down and I will comply.

PR #3497, opened 2017-09-29
Add template for document generation with 'jsdoc'.

wimrijnders:
In essence, this PR defines the subdirectory `docs` as a `jsdoc` template.

**Benefits:**
- allows the usage of partials, in order to DRY common parts of the html files.
- makes available the `jsdoc`-comments, for addition into the documentation.
- enables extraction of data from the source code. For example, the list of edge endpoints `['arrow', 'bar', 'circle']` can now be extracted from the source and inserted into the documentation on generation.

----
In this initial version, the only file that has been changed is `docs/data/dataset.html`. Here, partials have been added to illustrate how common page elements can be DRY'd.

The template has been set up in such a way, that resource files will be copied and that html files can pass through unchanged if no special template tags (`<?js...?>`) are used. This allows for a gradual transition of the html files to templates.

A `README.md` has been added to subdirectory `docs`, with notes on how to use the `jsdoc`-generated comment data.

----
**Usage:**  `gulp docs`

- The result files are placed in subdirectory `gen/docs/`.

**NOTE:** The release procedure will have to be adjusted by adding this gulp command.
The docs-files will then have to be taken from `gen/docs`.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -1,0 +1,239 @@
+# Usage
+
+```
+jsdoc -c jsdoc.json -r -t docs -d gen ../github/vis/lib
+
+```
+
+- `-c`: use this config file for `jsdoc`
+- `-r`: Recurse into source directory `../github/vis/lib`
+- `-t`: Use template in path `docs`
+- `-d`: Generated html files put in `gen`
+- Source files to parse by `jsdoc` taken from `../github/vis/lib`
+
+
+## Notes
+
+The template generation is set up so that:
+
+- Files ending in `.tmpl` are skipped
+- All non-html files are plain copied
+- html-files *can* contain `<?js ?>` tags, but this is not required
+
+The only file that has been changed is `docs/data/dataSet.html`.
+In here, partials have been added to illustrate how common page elements can be DRY'd.
+
+
+## Intention
+
+The `docs` directory is treated as a `jsdoc` template, in which the html-files are the template files.
+This allows for a gradual adaptation of the html-files to templates; unchanged html-files will pass 
+through `jsdoc` unchanged.
+
+The added value of using `jsdoc` for documentation generation, is that
+the complete documentation information, as collected by `jsdoc` from the source, is
+available for usage. This way, it's possible to insert technical notes from the source
+code into the documentation.
+
+----
+
+# Usage of and Notes on Source Code
+
+  ## Parameters of `publish()`
+
+  ### Parameter `taffyData`
+
+  A table containing *all* data collected from the source code, related to jsdoc generation.
+  See below for more info and example outputs.
+
+  ### Parameter `opt`
+
+  Example of `opt` variable:
+
+```
+{
+  "_":["../github/vis/lib/network/"],
+  "configure":"jsdoc.json",
+  "recurse":true,
+  "template":"/home/wim/projects/jsdoc/default",
+  "destination":"./out/",
+  "encoding":"utf8"
+}
+```
+
+  ### Parameter `tutorial`
+
+  This does not appear to be of use for the generation of `vis.js` documentation.
+
+  Example of `tutorial` variable:
+
+```
+{
+  "longname":"",
+  "name":"",
+  "title":"",
+  "content":"",
+  "parent":null,
+  "children":[],
+  "_tutorials":{}
+}
+```
+
+  ## Global variable `env`
+
+  This contains addition info for the current execution of `jsdoc`. Example of `env` variable:
+
+```
+{
+  "run":{"start":"2017-09-16T05:06:45.621Z","finish":null},
+  "args":["-c","jsdoc.json","-r","-t","default","../github/vis/lib/network/"],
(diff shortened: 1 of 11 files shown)


docs/README.md
```diff
@@ -0,0 +1,240 @@
...
+  The standard way of using this appear to be:
+  ```
+  <?js
+    var data = obj;   // Whatever docData is
+    var self = this;
+  ?>
+  ```
+ 
+  But it also appear to be possible to use the elements of docData directly:
+  ```
```
mbroad: You can get proper syntax highlighting by adding the name to your code block.

ie, 
<pre>
```javascript
var docData =...
```
<pre>

mbroad: json is also a valid key word for syntax highlighting

wimrijnders: yeah, I knew that, it just didn't occur to me to use it here.

```js
 var msg = 'js works as well.';
```

<pre>And now I learnt from you how to display ```js itself!</pre>

wimrijnders: Please do keep on reviewing! And yes try it out, please, if only to see if it works on machines other than mine.

docs/README.md
```diff
@@ -0,0 +1,240 @@
...
+- Files ending in `.tmpl` are skipped
+- All non-html files are plain copied
+- html-files *can* contain `<?js ?>` tags, but this is not required
+
+
+## Intention
+
+The `docs` directory is treated as a `jsdoc` template, in which the html-files are the template files.
+This allows for a gradual adaptation of the html-files to templates; unchanged html-files will pass 
+through `jsdoc` unchanged.
```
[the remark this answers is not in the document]
wimrijnders: ok. It should be readable as text as well, of course.

gulpfile.js
```diff
@@ -232,5 +233,27 @@ gulp.task('lint', function () {
...
+
+  // Not sure if this is the best way to handle 'cb'; at least it works.
+  var hasError = false;
+  var onError = function(error) {
+    if (error !== undefined && error !== null) {
+      console.error('Error while running task: ' + error);
+      hasError = true;
+      cb();
+    }
+  }
```
macleodbroad-wf: missing semi-colon

gulpfile.js
```diff
@@ -232,5 +233,27 @@ gulp.task('lint', function () {
...
+
+  rimraf(__dirname + '/' + targetDir, onError);  // Clean up previous generation
+
+  if (!hasError) {
+    var params = '-c ./jsdoc.json -r -t docs -d ' + targetDir;
+    child_exec('node ./node_modules/jsdoc/jsdoc.js ' + params + ' lib', undefined, cb);
+  };
```
macleodbroad-wf: unnecessary semi-colon

macleodbroad-wf:
I found that the index.html files did not work as I would have expected them to? Maybe this is a misunderstanding on my part.

gulpfile.js
wimrijnders: ok

gulpfile.js
wimrijnders: Due to this, I linted `gulpfile.js` and to my delight, I discovered that it's actually possible. What do you think, should it officially be added to the linting step?

wimrijnders:
The top index.html has this redirect in it which possibly reconnects to the main site. But all the other, lower-level, index.html's should be just fine.

The top-level index.html can be excused; do you have trouble with the other ones as well?

mbroad:
Maybe I'm misunderstanding how to use this.

If I navigate to dataset.html directly, everything looks fine.

If I open the index.html file, I'm redirected, but file not found.

If I copy and paste the entire file structure into the docs the gh-pages branch, replacing existing files, then the index pages work, but show file listings.

It is entirely possible that this is just user error.

In any case the generated dataset.html looks great.

wimrijnders:
You're not misunderstanding anything; that's how the html files are set up. These pages are made to be integrated into the web site, as in `gh-pages`. So, locally the redirect goes nowhere. Another thing I notice is that the links in the menu's are hardcoded, i.e. they take you to somewhere on `http://visjs.org`.

> show file listings

Perhaps you mean that they *don't* show?

As for the greatness of the generated page, it should be just identical to the original version. Next step is to isolate all the common parts in partials - obvious candidates: the HTML at top and bottom, the menu's. And there are plenty more duplicates of data in the doc's themselves, e.g. font settings in both nodes and edges.

Hoping for some help in converting them....

mbroad:
And will that also mean eventually incorporating documentation on the web page that is generated from some of the jsdoc in the source code itself?

wimrijnders:
Oh yes, definitely. That is in fact the major point of this exercise. The rest is just nice consequences.

mbroad:
Excellent!

wimrijnders:
Since the whole point of using `jsdoc` for document generation is to be able to use the header comments, I've added a proof of concept. Diffing should make everything clear.

Issue #620, opened 2015-02-06
Stacking options for groups/subgroups/items


billbither:
I would like to specify more stacking options. For example some items (or group of items) I'd like to stack, while others I would not. 

As you can see with the below image, the bottom row shouldn't stack as it represents continuous time of different conditions, while the subgroup above are single events that I'd like to stack.
![image](https://cloud.githubusercontent.com/assets/5885418/6074663/73aaa0a8-ad91-11e4-9e21-47cfd071329d.png)

AlexDM0:
@josdejong maybe we could look into setting stacking per group?

josdejong:
@AlexDM0 I guess you mean stacking per *subgroup*, stacking works already fine for groups ([see example](http://visjs.org/examples/timeline/05_groups.html)). Yes, that makes sense.

AlexDM0:
Hi,

I think you misunderstood, I ment to select on a group to group basis if you want to use automatic stacking OR the subgroups.

Regards

josdejong:
Ah ok, so you can choose per group whether you want stacking or subgroups or no stacking. Sounds good.

bbseven:
Thanks, that will be a great enhancement!

carueda:
:+1:

Atala:
@josdejong did you have time to work on this?
@bbseven did you found a solution/hack for this ?

josdejong:
I haven't yet worked on it, it's still on the list. Unless we want a hacky solution, it will involve a serious refactoring of the subgroups.

Atala:
Thanks, I will have a look a little bit later.

simo9000:
I just ran into the need for this feature. Can't wait until it's added.

odick:
Hi all, not sure if this would help
I got the solution from [stackoverflow] (http://stackoverflow.com/questions/33310120/vis-js-timeline-dont-stack-items-with-no-time-overlap)
```
var options = {
    margin: {
        item : {
            horizontal : 0
        }
    }
};
```
result:

![pale moon 2016-01-04 22 48 43](https://cloud.githubusercontent.com/assets/1612324/12092180/72f1b5b0-b336-11e5-930d-6205451e3292.png)

simo9000:
I just tried this and it is a perfectly acceptable solution for my specific
problem.

Thanks.

scipilot:
[...]

PR #3501, opened 2017-09-29
Subgroup stacking

Areson:
This pull request is in reference to issue #620. Enables stacking to be controlled within each subgroup, where some will not stack and others will. When stacking within subgroups is enabled, subgroup stacking must also be on, and the stacks are sorted _with respect to the subgroup they are in_, and each subgroup stack will not overlap with the stack of another subgroup (e.g., the subgroups themselves are stacked, and the contents within each subgroup can also be stacked or overlapping).

I found a need in a project I was doing to have some items in a group shown on a timeline without stacking because they were sequential, but other items within the same group needed to be stacked as they were overlapping. I've updated the subgroup example to allow toggling between subgroup stacking (the original functionality) and this new, within-subgroup stacking.

### DIFF
--- a/docs/timeline/index.html
+++ b/docs/timeline/index.html
@@ -455,6 +455,15 @@ var groups = [
       </td>
     </tr>
     <tr>
+      <td>subgroupStack</td>
+      <td>Object</td>
+      <td>none</td>
+      <td>Enables stacking within individual subgroups. Example: <code>{'subgroup0': true, 'subgroup1': false, 'subgroup2': true}</code> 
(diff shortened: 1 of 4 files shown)


Areson:
@yotamberk I'm not seeing the `keyword-spacing` directive in the .eslintrc file, just FYI.

I've made updates to fix the issues, as well as added support for the <code>subgroupStack</code> parameter to take a true/false value, as you suggested.

I looked into the example. It didn't work for me either until I did a build of the project. Additionally, you'll need to make sure that Stacking is turned on before the subgroup stacking works. Let me know if you still have issues.

PR #3503, opened 2017-09-29
Add CDN link

LukasDrgon:
I added a [jsDelivr CDN link](https://www.jsdelivr.com/package/npm/vis) to your readme, so that people can use files directly without downloading them. jsDelivr is also the [fastest opensource CDN](https://www.cdnperf.com/) available and built specifically for production usage.

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,6 +31,12 @@ Install via bower:
 
     $ bower install vis
 
+Link via jsDelivr CDN:
+
+```
+<script src="https://cdn.jsdelivr.net/npm/vis@4/dist/vis.min.js"></script>
+```
+
 Link via cdnjs: http://cdnjs.com
 
 Or download the library from the github project:


mbroad:
I would like to re-evaluate once jsDelivr comes out of beta.

jimaek:
What makes you think its in beta? Its stable and can be used in production without any problems.

mbroad:
My mistake. I clicked on your CDN Perf link, which has a Beta banner across the top.

jimaek:
Oh ok then. 
jsDelivr was launched in 2012 and currently serves 19billion hits per month :)

mojoaxel:
vis js is already on cdnjs:
https://cdnjs.com/libraries/vis

@jimaek Thanks anyway.

jimaek:
Well we have more features, like download stats per day per file per version, instant updates by pulling from NPM, auto-minification, combination of files, 4 CDNs instead of 1 and more.

But ok, no problem

Issue #2851, opened 2017-03-09
Timeline focus() doesn't center vertically


instacyber:
When I call focus() on a specific item ID, it centers the timeline horizontally on this item (i.e. by date) as expected. 

However, when constraining a timeline with a maxHeight and rendering a timeline with more groups than can fit in the view window, calling focus() does not pull the vertical scroll in a way that also displays the selected item if it is not currently visible.

Is this expected behaviour? Examination of the function code in vis.js suggests that it doesn't attempt any vertical correction, which I believe would be helpful to completely enact a "focus" on a specific item.

yotamberk:
You're right! Thanks for reporting

rbkn:
Related:  It looks like getVisibleItems() doesn't take into account vertical visibility, only horizontal.

In ItemSet.prototype.getVisibleItems there's a TODO comment to that effect too.

Areson:
I added this to the duplicate, so I'll add this here as well. This is an [example workaround](https://jsfiddle.net/Areson/z2rsyhow/) with jQuery. It's not heavily dependent on jQuery, so depending on what type of modification this requires it could be adapted to fix this issue.

PR #3504, opened 2017-09-30
Vertical focus

Areson:
This pull request is an attempt to satisfy #2851. Adds vertical scrolling support to the `focus` method by hooking into the existing animation for setting the range. The `setSelection` example was updated with a second example showing vertical scrolling when using `focus`.

### DIFF
--- a/examples/timeline/interaction/setSelection.html
+++ b/examples/timeline/interaction/setSelection.html
@@ -26,6 +26,17 @@
 </p>
 <div id="visualization"></div>
 
+<br/>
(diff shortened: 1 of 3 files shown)


Areson:
Updated 1 and 3. I'll get to 2 later today.

Areson:
I've updated the example to try to be a bit more clear about what it is demonstrating and what the buttons do. Let me know if it is more in line with what you wanted to see, and if there is a different direction you want it to go let me know.

Issue #3505, opened 2017-09-30
Drag and drop objects in to items


yotamberk:
As a user, I would like to be able to be able to drag objects and drop them on to items and call a function once succeeds. 
This is different than the current ability to drag and drop external items from the timeline, and adding in as timeline items.

PR #3506, opened 2017-09-30
Drag object in to item

yotamberk:
Addresses #3505 

It's a feature I really need for my work.
I found it really useful and A LOT of potential for further features for apps using Timeline

### DIFF
--- a/examples/timeline/other/drag&drop.html
+++ b/examples/timeline/other/drag&drop.html
@@ -132,7 +132,7 @@
     editable: true,
     orientation: 'top',
     onDropObjectOnItem: function(objectData, item, callback) {
-      if (!objectData) { return; }
+      if (!item) { return; }
       alert('dropped object with content: "' + objectData.content + '" to item: "' + item.content + '"');
     }
   };
--- a/examples/timeline/other/drag&drop.html
+++ b/examples/timeline/other/drag&drop.html
@@ -157,7 +157,6 @@
       item.start = new Date();
       item.end = new Date(1000*60*10 + (new Date()).valueOf());
     }
-
     event.dataTransfer.setData("text", JSON.stringify(item));
   }
 
--- a/examples/timeline/other/drag&drop.html
+++ b/examples/timeline/other/drag&drop.html
@@ -132,7 +132,8 @@
     editable: true,
     orientation: 'top',
     onDropObjectOnItem: function(objectData, item, callback) {
-      alert('dropped object with content: "' + objectData.content + '" to item: "' + item.content + '"')
+      if (!objectData) { return; }
+      alert('dropped object with content: "' + objectData.content + '" to item: "' + item.content + '"');
     }
   };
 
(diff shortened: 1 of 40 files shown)


examples/timeline/other/drag&drop.html
```diff
@@ -27,6 +27,27 @@
...
+      width: 150px;
+      color: #1A1A1A;
+      background-color: #D5DDF6;
+      border: 1px solid #97B0F8;
+      border-radius: 2px;
+      margin-bottom: 5px;
+      padding: 5px 12px;
+    }
+    li.object-item:before {
+      content: "≣";
```
macleodbroad-wf: What is this for?

lib/timeline/component/ItemSet.js
```diff
@@ -2070,69 +2084,65 @@ ItemSet.prototype._onAddItem = function (event) {
...
+  newItemData = this._cloneItemData(newItemData);     // convert start and end to the correct type
+  this.options.onAdd(newItemData, function (item) {
+    if (item) {
+      me.itemsData.getDataSet().add(item);
+      if (event.type == 'drop') {
+        me.setSelection([item.id]);
       }
-    });
-  }
+      // TODO: need to trigger a redraw?
```
macleodbroad-wf: Was this TODO to be addressed prior to merge? Or would this be expected to be determined after merge?

examples/timeline/other/drag&drop.html
yotamberk: so the example has a small symbol (which usually indicates it's draggable)

lib/timeline/component/ItemSet.js
yotamberk: I was there before this PR. Still not sure if a redraw trigger is needed. 
I'm trying to fully understand the redrawing flows to understand how to improve the flow

lib/timeline/component/item/PointItem.js
```diff
@@ -87,7 +87,6 @@ PointItem.prototype._appendDomElement = function() {
   }
   this.displayed = true;
 }
-
```
wimrijnders: empty line was good here.

lib/timeline/component/ItemSet.js
```diff
@@ -2070,69 +2084,65 @@ ItemSet.prototype._onAddItem = function (event) {
 
   var me = this;
   var snap = this.options.snap || null;
-  var item = this.itemFromTarget(event);
```
wimrijnders: Heh. I see. First of all `item` not used in this code block. Second, passing an existing item is silly since you're creation a new one.

examples/timeline/other/drag&drop.html
```diff
@@ -171,7 +171,6 @@
     event.dataTransfer.setData("text", JSON.stringify(objectItem));
   }
 
-
   var items = document.querySelectorAll('.items .item');
```
wimrijnders: OK, never mind. I feel silly about commenting on 'too many empty lines' anyway. There's probably a reason why you want it like this.

yotamberk:
Thanks again @wimrijnders !
I really do appreciate your reviews! I learn every time.

Issue #3468, opened 2017-09-20
[Network] physics simulation runs multiple times but stabilizationIterationsDone event only fires once


rjbriody:
Chrome Version 60.0.3112.113 (Official Build) (64-bit)
OSX
Vis version 4.20.1

This might be easiest to describe w/ an example https://jsfiddle.net/nxs3poo6/, but here's the deal...

- Fire up a network w/ physics enabled.
- graph is displayed. physics animation happens. eventually, graph stops moving.
- `stabilizationIterationsDone` event is fired.
- augment the network (add another node and edge) using the underlying DataSet.
- physics starts again. `startStabilizing` event is fired.
- graph stops moving
- (and this is where the problem occurs) **`stabilizationIterationsDone` is never fired**

wimrijnders:
Confirmed. Looking at the docs, I see this for `stabilizationIterationsDone`:

> Fired when the 'hidden' stabilization finishes. This does not necessarily mean the network is stabilized; it could also mean that the amount of iterations defined in the options has been reached.

Now, I have no idea what is meant by 'hidden', and why it is between quotes. Taking it at face value, I interpret this as meaning 'a stabilization triggered by a data change`, i.e. all stabilizations except the initial one on startup. Which means that current implementation is exactly the wrong way around.

I'd rather get rid of 'hidden' in the docs and just trigger the event always. Will get back to this.

wimrijnders:
I'm having second thoughts on this one. `stabilizationIterationsDone` is meant specifically for the stabilization on startup. If you need an event for the completion of interim stabilization, there is `stabilized`.

I propose that you handle event `stabilize` instead. Closing this issue, if you disagree with this please comment here and we'll discuss further.

rjbriody:
I agree, and in fact have transitioned to using the `stabilized` event. I was originally confused by the docs as it seemed like `stabilized` would only be called when the network was successfully stabilized and not when the simulation stopped early due to reaching the number of max iterations.

A possible docs improvement could be to change...

stabilized
> Fired when the network has stabilized or when the stopSimulation() has been called. The amount of iterations it took could be used to tweak the maximum amount of iterations needed to stabilize the network. Passes an object with properties structured as:
 
to something like this

> Fired when the network has stabilized, **when the amount of iterations defined in the options has been reached,** or when the stopSimulation() has been called. The amount of iterations it took could be used to tweak the maximum amount of iterations needed to stabilize the network. ...

wimrijnders:
[...]

PR #3507, opened 2017-09-30
Network: Cleanup and refactoring PhysicsEngine

wimrijnders:
Some work done in preparation for #3468, while examining the code for better understanding. The logic remains unchanged.

Note that this does not solve #3468, it's just a preliminary cleanup.

### DIFF
--- a/lib/network/modules/PhysicsEngine.js
+++ b/lib/network/modules/PhysicsEngine.js
@@ -6,8 +6,8 @@ var HierarchicalSpringSolver = require('./components/physics/HierarchicalSpringS
 var CentralGravitySolver = require('./components/physics/CentralGravitySolver').default;
 var ForceAtlas2BasedRepulsionSolver = require('./components/physics/FA2BasedRepulsionSolver').default;
 var ForceAtlas2BasedCentralGravitySolver = require('./components/physics/FA2BasedCentralGravitySolver').default;
-
 var util = require('../../util');
+var EndPoints = require('./components/edges/util/EndPoints').default; // for debugging with _drawForces()
 
 
 /**
@@ -121,14 +121,13 @@ class PhysicsEngine {
       this.stopSimulation(false);
       this.body.emitter.off();
     });
-    // this event will trigger a rebuilding of the cache everything. Used when nodes or edges have been added or removed.
     this.body.emitter.on("_dataChanged", () => {
-      // update shortcut lists
+      // Nodes an/or edges have been added or removed, update shortcut lists.
       this.updatePhysicsData();
     });
 
     // debug: show forces
-    // this.body.emitter.on("afterDrawing", (ctx) => {this._drawForces(ctx);});
+    //this.body.emitter.on("afterDrawing", (ctx) => {this._drawForces(ctx);});
   }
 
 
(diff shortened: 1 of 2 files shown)


lib/network/modules/PhysicsEngine.js
```diff
@@ -121,14 +121,13 @@ class PhysicsEngine {
       this.stopSimulation(false);
       this.body.emitter.off();
     });
-    // this event will trigger a rebuilding of the cache everything. Used when nodes or edges have been added or removed.
     this.body.emitter.on("_dataChanged", () => {
-      // update shortcut lists
+      // Nodes an/or edges have been added or removed, update shortcut lists.
```
yotamberk: `an/or` => and/or

wimrijnders: OK. Can you hold off merging until @mbroad has reviewed, please?

lib/network/modules/components/physics/BarnesHutSolver.js
```diff
@@ -62,17 +64,27 @@ class BarnesHutSolver {
...
     }
   }
 
 
   /**
+   * @param {Object} parentBranch
+   * @param {Node} node
+   * @private
+   */
+  _getForceContributions(parentBranch, node) {
```
mbroad: A function beginning with `_get` that has no return value...
Now I know this isn't your fault, as you're being consistent with `_getForceContribution`, but this seems wrong.

wimrijnders:
Yeah, tough one to unit test on, because everything should remain as is. I consider the examples in the unit tests running to be a good indication of fitness - actually all tests that run physics.

I did a visual check by running `worldldCupPerformance.html` with `develop` and this PR side by side.

OK, sooooo....think about a possible unit test here. The premise is `physics unchanged`: so how about setting a given seed and checking after a fixed number of iterations if the output is identical to previous?

The drawback I see to that idea is that it's most likely a single-shot test. If *anything* changes in physics, you can throw the test away.

mbroad:
I just meant testing the inputs, outputs, and side effects of individual functions in isolation.

lib/network/modules/PhysicsEngine.js
yotamberk: Of course, that's exactly why I waited

wimrijnders:
Actually, I think a full test is easier to do at this point. This would be an indication that the whole physics thing works as before.

lib/network/modules/PhysicsEngine.js
```diff
@@ -121,14 +121,13 @@ class PhysicsEngine {
...
       this.updatePhysicsData();
     });
 
     // debug: show forces
-    // this.body.emitter.on("afterDrawing", (ctx) => {this._drawForces(ctx);});
+    //this.body.emitter.on("afterDrawing", (ctx) => {this._drawForces(ctx);});
```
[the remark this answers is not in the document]
wimrijnders: No reason whatsoever. My fingers were probably thinking for me. I actually prefer the space.
Ah yes, I enabled that to test the call, then I added the //'s back again.

lib/network/modules/components/physics/BarnesHutSolver.js
wimrijnders: OK. Well, there is a force contribution to be gotten, only it's not returned but added to param `parentBranch`. I can't think of any better names right now.

wimrijnders:
It's actually quite instructive to enable the debug routine for `BarnesHutSolver`. It gives a visual indication of what is actually happening:

download

download 1

Apparently, the optimization here is to subdivide the area and deal with the forces in each area separately. I can sort of intuit that this works.

PR #3508, opened 2017-09-30
Vertical Focus - Bugfix on Final Vertical Scroll Position

Areson:
Bugfix for #3504.

@yotamberk Sorry for the quick second pull request on this. I realized once I put everything back together that I made an error with where the `setTimeout` was located for ensuring the final scroll position. It works fine in the example, but a more complex timeline with a heavier redraw (like the one I have in my own project) needs to have this in the proper location.

### DIFF
--- a/lib/timeline/Timeline.js
+++ b/lib/timeline/Timeline.js
@@ -428,32 +428,38 @@ Timeline.prototype.focus = function(id, options) {
       }
     };
     
+    // Enforces the final vertical scroll position
+    var setFinalVerticalPosition = function() {
+      var finalVerticalScroll = getItemVerticalScroll(me, item);
+
+      if (finalVerticalScroll.shouldScroll && finalVerticalScroll.itemTop != initialVerticalScroll.itemTop) {
+        me._setScrollTop(-finalVerticalScroll.scrollOffset);
+        me._redraw();
+      }
+    };
+
     // Perform one last check at the end to make sure the final vertical
     // position is correct
     var finalVerticalCallback = function() {
-      var finalVerticalScroll = getItemVerticalScroll(me, item);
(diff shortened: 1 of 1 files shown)


lib/timeline/Timeline.js
```diff
@@ -428,32 +428,38 @@ Timeline.prototype.focus = function(id, options) {
...
     };
 
+    // Perform one last check at the end to make sure the final vertical
+    // position is correct
+    var finalVerticalCallback = function() {
+      // Double check we ended at the proper scroll position
+      setFinalVerticalPosition();
+
+      // Let the redraw settle and finalize the position.      
+      setTimeout(setFinalVerticalPosition, 100);
```
yotamberk: Try `requestFrameAnimation` instead of `setTimeout`

Areson: I don't think that will accomplish what I'm trying to do here. The core issue that I noticed is the timeline resizes groups as you scroll horizontally, and if that resize happens just after the vertical scroll finishes then the "final" position the user sees is not the item that they asked to have focused. I don't have an event I can hook into for that at the moment, so the best I've got is  a small delay after the vertical scroll has finished to essentially do a "double check". `requestAnimationFrame` would likely call this too soon, which is why I chose `setTimeout`. If you have a suggestion for an event or something else I could check that would be more reliable I'll gladly use that.

yotamberk: mmm Not really. I'll approve it for now. I'l think about it more though.

Areson: I'm wanting to try and do some tracing of the various events that are being emitted so I can understand where and when the recalculations come from post-focus. If I find something l'll let you know.

PR #3511, opened 2017-10-01
Network: Add extra check on null value during label handling

wimrijnders:
This fixes an oversight in #3486.
Also added unit tests, not only for null labels, but for all weird label values I could thing of.


wimrijnders:
'Stupid' is the word I would use. Missing something so fundamental as that....

yotamberk:
Nah, we improve all the time. If we were perfect on the first time all the time, we wouldn't have work =P

Issue #3500, opened 2017-09-29
Edges in dataset may no longer be visible even if valid nodes exist.


justinharrell:
Based on internal state changes made in #3330 unconnected edges that later become connected due to node dataset modifications no longer appear on the graph until the edge itself is updated.

This is a change in behavior, previously unconnected / invalid edges, those which the from or to node id does not exist, where not drawn on the graph. However when a node was updated or added that completed the connection it was displayed in the graph without further intervention.

The old behavior allowed out of order operations, such that edges could be added before their respective nodes and nodes could be removed and re-added with edges following them properly. Now the developer must update any affected edge if a node is updated completing the connection or always add nodes before their edges.

Discussion in #3330 determined this as a bug and not intended behavior.

I am working on a fix and will submit a PR.

wimrijnders:
And an extra :+1: for picking it up yourself.

justinharrell:
No problem it allows me to get a better understanding of that area in the code and I happen to have time to devote to it.

PR #3516, opened 2017-10-02
Fix issue #3500

justinharrell:
This fixes unconnected edges not showing back up if nodes are modified or added that connect them.

I tried a couple approaches, this could probably be optimized some with deeper changes, but I choose to keep the fix as simple as possible.

### DIFF
--- a/lib/network/modules/EdgesHandler.js
+++ b/lib/network/modules/EdgesHandler.js
@@ -422,12 +422,23 @@ class EdgesHandler {
   }
 
   /**
-   * Scan for missing nodes and remove corresponding edges, if any.
-   *
    * There is no direct relation between the nodes and the edges DataSet,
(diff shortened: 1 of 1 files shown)


wimrijnders:
Fixes #3500

wimrijnders:
While the code fix is 100% good, I neglected to observe that there is no regression test for this fix. Do you have the energy to add that?

I'm thinking of some network where a node is removed and then added again. The test would be that the edges for that node have reconnected.

If you don't feel like adding that, I'm prepared to do it myself. It's my oversight.

justinharrell:
I pushed a change with a test and some white-space cleanup into the fix branch, would that be a new PR?

wimrijnders:
Sincerely thanks for your effort. I've never even tried pushing to a merged PR before, is it possible?
Let's not tart fate. A new PR would do nicely.

PR #3520, opened 2017-10-04
Network: Retain constraint values in label font handling

### LINKED ISSUE #3517: widthConstraint no longer has any effect
After PR #3486 widthConstraint no longer seems to have any effect.

I noticed this in our networks after rebasing and confirmed by navigating to: vis/test/network/maximumWidthEdgeCase.html

Before:

widthconstraint 1

After:

widthconstraint 2

wimrijnders:
Fixes #3517.

Due to changed logic in the label font handling, the option values for `widthConstraint` and `heightConstraint` were overwritten.

The fix is in effect a reversal of two code lines: parsing constraint options should come *after* parsing (multi)font options.

Further changes:

- Additional 1-liner fix: constraint values were not copied for edge-instance specific options.
- Small refactoring of `Label#constrain()` in order to separate concerns
- Added unit test for regression testing of this issue.

All of which entices me to make the curious observation that two changed lines of source code has led to a +-150 line regression test.

----
Attentive observers may note that the automatic line breaks are now positioned differently, e.g. in the [widthHeight example](http://visjs.org/examples/network/nodeStyles/widthHeight.html).

This is due to a fix in the line width calculation in #3486: current stable does not take the spaces into account when calculating width.

### DIFF
--- a/test/Label.test.js
+++ b/test/Label.test.js
@@ -1414,7 +1414,6 @@ describe('Shorthand Font Options', function() {
    * where the associated issue (i.e. widthConstraint values not copied) was most poignant.
    *
    * NOTE: boolean shorthand values for widthConstraint and heightConstraint do nothing.
-   * perhaps TODO: check shorthand values for heightConstraint.
    */
   it('Sets the width/height constraints in the font label options', function (done) {
     var nodes = [
--- a/test/Label.test.js
+++ b/test/Label.test.js
@@ -1407,4 +1407,178 @@ describe('Shorthand Font Options', function() {
 
     done();
   });
+
+
+  /**
+   * 
+   * The test network is taken verbatim from example `network/nodeStyles/widthHeight.html`,
+   * where the associated issue (i.e. widthConstraint values not copied) was most poignant.
+   *
+   * If you want to be really pedantically paranoid about it, widthConstraint should also 
+   * be tested when set in groups. perhaps TODO.
+   *
+   * perhaps TODO: check boolean shorthand values for widthConstraint.
+   * perhaps TODO: check shorthand values for heightConstraint.
+   */
+  it('Sets the width/height constraints in the font label options', function (done) {
+    var nodes = [
+      { id: 100, label: 'This node has no fixed, minimum or maximum width or height', x: -50, y: -300 },
+      { id: 210, widthConstraint: { minimum: 120 }, label: 'This node has a mimimum width', x: -400, y: -200 },
+      { id: 211, widthConstraint: { minimum: 120 }, label: '...so does this', x: -500, y: -50 },
+      { id: 220, widthConstraint: { maximum: 170 }, label: 'This node has a maximum width and breaks have been automatically inserted into the label', x: -150, y: -150 },
+      { id: 221, widthConstraint: { maximum: 170 }, label: '...this too', x: -100, y: 0 },
+      { id: 200, font: { multi: true }, widthConstraint: 150, label: '<b>This</b> node has a fixed width and breaks have been automatically inserted into the label', x: -300, y: 50 },
+      { id: 201, widthConstraint: 150, label: '...this as well', x: -300, y: 200 },
+      { id: 300, heightConstraint: { minimum: 70 }, label: 'This node\nhas a\nminimum\nheight', x: 100, y: -150 },
+      { id: 301, heightConstraint: { minimum: 70 }, label: '...this one here too', x: 100, y: 0 },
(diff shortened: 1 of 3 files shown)


test/Label.test.js
```diff
@@ -1408,6 +1408,180 @@ describe('Shorthand Font Options', function() {
...
+    //
+    // Check the internal constraint values
+    //
+
+    let assertConstraints = (item, fontOptions, expectedMaxWidthDefault) => {
+      if (item.widthConstraint === undefined) {
+        assert.equal(fontOptions.minWdt, -1);
+        assert.equal(fontOptions.maxWdt, expectedMaxWidthDefault);
+      } else {
+        let constraint = item.widthConstraint;
```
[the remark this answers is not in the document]
wimrijnders: Perhaps that's what you mean: In the unit test, since you know exactly what value you are expecting, so can do a direct test on input and output. So, for e.g. node 100, following should suffice:

```js
  assert.equal(fontOptions.minWdt, -1);
  assert.equal(fontOptions.maxWdt, 200);
  assert.equal(fontOptions.minHgtt, -1);
  assert.equal(fontOptions.valign, 'middle');
```

So trading in the conditionals for repetitive but flat code. Is that the clue? I'm agreeing to that. It's not that I don't know how to do that, e.g. `it('respects the font option precedence`)`.

In fact, I'm not even going to wait for your confirmation that that is what you mean to change to this. I've been following the source code logic too much here (notably `Label#constrain()`).

wimrijnders:
> Not necessarily as part of the ticket, but it seems like we have a lot of work to do around documenting and testing the order and application of options.

Oh man. Surely you couldn't have missed me lamenting about the state of network option handling in the lobby. The direction I'd like to go, is to simplify option handling as much as possible so that the actual intent becomes obvious. 'We're not there yet' is an euphemism in this context.

wimrijnders:
The consequence of making it more linear is that it's now obvious that there is too much checking of the same thing.

I'm going let the idea of using the example go and make an 'enhanced subset' of it, so to speak, which also addresses the TODO's in the header comment.

Also, I notice that internal font option fields `constrainWidth` and `constrainHeight` are not tested, will add these as well.

test/Label.test.js
```diff
@@ -1408,6 +1408,150 @@ describe('Shorthand Font Options', function() {
...
+      { nodeId: 401, minWdt:  -1, maxWdt: 200, minHgt: 100, valign: 'middle'},
+      { nodeId: 402, minWdt:  -1, maxWdt: 200, minHgt: 100, valign: 'bottom'},
+    ];
+
+
+    // For edge labels, only maxWdt is set. We check the rest anyway, be it for
+    // checking incorrect settings or for future code changes.
+    //
+    // There is a lot of repetitiveness here. Perhaps using a direct copy of the
+    // example should be let go.
```
mbroad: I agree wholeheartedly with this comment.

Test set ups should be made up of the minimal required code to satisfy preconditions for your tests.

mbroad:
> This also means that boolean can/should be removed from the option definitions. This would be a breaking API change, though the least malevolent I can imagine. Users in their right mind wouldn't be using the value anyway, and worst of all, it does nothing.

Should it really be removed, or just actually set?

mbroad:
> Well, crap. It appears that constrainWidth and constrainHeight are never set. @mbroad, would you mind verifying this?

Verified. They are both set to `false`, but nothing else modifies them.

test/Label.test.js
wimrijnders: OK. The thing is, I've already let go of the example and changed the test accordingly. I'll remove this comment and add some extra explanation.

wimrijnders:
> Verified. They are both set to `false`, but nothing else modifies them.

Thanks. I'll consider them artifacts and remove them and `Label#adjustSize()` which is the only place were they are accessed and which thus does nothing.

wimrijnders:
> Should it really be removed, or just actually set?

Had a think about this: I can image a scenario where the `widthConstraint`/`heightConststraint` gets overridden in the more specific options (e.g. defines in the global options, overridden in group or node instance).

So setting to `false` actually has potential use. So it needs to be handled as well. What to do with setting to `true`, however? That is useless......

Anyway, I'll add the handling in.

wimrijnders:
Wrt. fully enabling `width/heightConstraint`: I'm digging myself into a hole here.

The consequence is that suddenly the margins are used again - *option `margin` has not been active in a long time*. So the changes are propagating.

Rather than fix this as well, I'll leave the margins as is, make a note and create an issue for it.

wimrijnders:
Oh, merged already ☹️. I was hoping to get some changes in still. Never mind, I'll add them in #3536.

Issue #3529, opened 2017-10-05
On initial timeline draw complete


yotamberk:
As a user, I need an event/function that fires when a timeline initial draw is completed

PR #3530, opened 2017-10-05
On timeline loaded

yotamberk:
Solves #3529

### DIFF
--- a/docs/timeline/index.html
+++ b/docs/timeline/index.html
@@ -886,7 +886,7 @@ function (option, path) {
       <td>Callback function triggered when an item is about to be added: when the user double taps an empty space in the Timeline. See section <a href="#Editing_Items">Editing Items</a> for more information. Only applicable when both options <code>selectable</code> and <code>editable.add</code> are set <code><code>true</code></code>.
       </td>
     </tr>
-
+    
     <tr>
       <td>onAddGroup</td>
       <td>function</td>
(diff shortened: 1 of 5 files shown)


docs/timeline/index.html
```diff
@@ -944,6 +944,14 @@ <h2 id="Configuration_Options">Configuration Options</h2>
     </tr>
 
     <tr>
+      <td>onInitialDrawComplete</td>
```
wimrijnders: Small niggle on my part: The options are arranged alphabetically, this new entry and `onUpdate` are not.

examples/timeline/interaction/eventListeners.html
```diff
@@ -34,7 +34,8 @@
 
     var container = document.getElementById('visualization');
     var options = {
-      editable: true
+      editable: true,
+      onInitialDrawComplete: function() { return logEvent('Timeline initial draw completed', {}) },
```
wimrijnders: `return` not necessary. `logEvent` returns nothing.

lib/timeline/Core.js
```diff
@@ -1003,7 +1003,6 @@ Core.prototype._redraw = function() {
   } else {
     this.redrawCount = 0;
   }
-  this.initialDrawDone = true;
```
wimrijnders: Doesn't this need to be set? It's the only place in the module where it goes to `true`.

OK, got it. It's set in `Timeline._activateOnInitialDrawDone()`. I find it a bit confusing that this is in a derived class.

lib/timeline/Timeline.js
```diff
@@ -202,6 +202,16 @@ function Timeline (container, items, groups, options) {
 // Extend the functionality from Core
 Timeline.prototype = new Core();
 
+Timeline.prototype._activateOnInitialDrawDone = function() {
+  if (!this.initialDrawDone) {
+    this.initialDrawDone = true;
+    setTimeout(() => {
+      if (this.options.onInitialDrawComplete) {
```
wimrijnders: Why bother with `setTimeOut` if `onInitialDrawComplete` is not present? Better to move this condition out of `setTimeOut`.

lib/timeline/Timeline.js
```diff
@@ -160,22 +161,21 @@ function Timeline (container, items, groups, options) {
...
           var range = me.getItemRange();
         }
 
         var start = me.options.start != undefined ? me.options.start : range.min;
         var end   = me.options.end   != undefined ? me.options.end   : range.max;
-        me.setWindow(start, end, {animation: false});
-      }
-      else {
-        me.fit({animation: false});
+        me.setWindow(start, end, {animation: false}, this._activateOnInitialDrawDone.bind(this));
```
wimrijnders: Strictly speaking, calling it here is a lie; `_redraw()` is done at the end of this method. `initialDrawDone` should be set to true *after* that call, same applies to the callback.

lib/timeline/Timeline.js
```diff
@@ -411,21 +421,22 @@ Timeline.prototype.focus = function(id, options) {
  *                                    provided to specify duration and easing function.
  *                                    Default duration is 500 ms, and default easing
  *                                    function is 'easeInOutQuad'.
+ * @param {Function} callback
```
wimrijnders: `@param {function} [callback]`

- casing, callback optional

examples/timeline/interaction/eventListeners.html
wimrijnders: ;

lib/timeline/Core.js
yotamberk: `Timeline._activateOnInitialDrawDone()` was incorrect. I removed it.

docs/timeline/index.html
```diff
@@ -886,71 +886,71 @@ <h2 id="Configuration_Options">Configuration Options</h2>
...
       <td>Callback function triggered repeatedly when an item is being moved. See section <a href="#Editing_Items">Editing Items</a> for more information. Only applicable when both options <code>selectable</code> and <code>editable.updateTime</code> or <code>editable.updateGroup</code> are set <code><code>true</code></code>.
       </td>
     </tr>
-
+    
     <tr>
       <td>onRemove</td>
       <td>function</td>
       <td>none</td>
       <td>Callback function triggered when an item is about to be removed: when the user tapped the delete button on the top right of a selected item. See section <a href="#Editing_Items">Editing Items</a> for more information. Only applicable when both options <code>selectable</code> and <code>editable.remove</code> are set <code><code>true</code></code>.
```
wimrijnders: Er, why the double `<code>` tags? It happens more often in the docs.

lib/timeline/Core.js
```diff
@@ -111,6 +111,11 @@ Core.prototype._create = function (container) {
       this._redraw();
     }
   }.bind(this));
+  this.on('rangechanged', function () {
```
wimrijnders: Yes, this is a better place.

lib/timeline/Timeline.js
```diff
@@ -173,9 +174,19 @@ function Timeline (container, items, groups, options) {
...
         var end   = me.options.end   != undefined ? me.options.end   : range.max;
-        me.setWindow(start, end, {animation: false}, this._activateOnInitialDrawDone.bind(this));
+        me.setWindow(start, end, {animation: false});
       } else {
-        me.fit({animation: false}, this._activateOnInitialDrawDone.bind(this));
+        me.fit({animation: false});
+      }
+    }
+
+    if (!me.initialDrawDone && me.initialRangeChangeDone) {
```
wimrijnders: Ah yes right. If I was just a bit clearer in my head I should have remarked on that. Make sense, not directly connected to the while-loop.

yotamberk:
@wimrijnders seems like that `<code><code>` was in many places... thanks for noticing!

Issue #3323, opened 2017-07-31
Timeline: Control the scale(s) and step(s) of both major and minor gridlines, independently


ShawnMercado:
I've been using this library for a few weeks now, and while it works well for most use cases, not having direct control over when gridlines is drawn can limit the usefulness or force the user to compromise.

I'm pretty sure what I'm asking for is not a feature right now, but if I'm wrong about that please let me know.

First, I think at the very least the `showMajorLabels` option, when false, should prevent drawing major gridlines.  Currently, for instance, if you set your timeline to the week scale, there will still be lines drawn at the month boundaries even when hiding the major labels.  Months do not follow naturally from weeks, and there are plenty of use cases where all you may care about is weeks, and all the month lines do is add clutter.

(This is a problem I'm trying to find a workaround for right now, so if anyone has any suggestions, please let me know).

But even better would be if I could control exactly when a major gridline is drawn in the first place. 

Currently, you can set a fixed scale by adding an option like this:

`options = {
    timeAxis: {
        scale: 'day',
        step: 1,
}`

It would be far more useful if I could set this both for the minor lines and the major lines.  So a really obvious example would be like (and this is just an example, i don't know how it would actually be implemented):

`options = {
    timeAxis: {
        minorLines: {
            scale: 'day',
            step: 1
        },
        majorLines: {
            scale: 'week',
            step: 1,
        },
    }
}`

Another example might be single days for minor lines and then 30 days for major lines (I have a client who specifically plans work in this manner, actually).  

But there's another thing, with the current implementation, you can only set one scale/step and then you lose a lot of what makes the library useful.  It would be super, duper nice if we could define an array of scale combinations somehow, and that zooming the timeline would simply cycle through these predefined scales.  I know that might be asking a lot but currently the zoom functionality is just not that useful if you have to go away from the default scaling.

ponml:
Hey @ShawnMercado, I have a PR up right now that I believe addresses the week/month major label issue: 
https://github.com/almende/vis/pull/3305
original issue: #3246

ShawnMercado:
@ponml oh cool.  that will be very helpful when available.

marcortw:
[...]

PR #3549, opened 2017-10-11
Week scale - enhance and add to auto-scaling

knokit:
**Changes:**
* ~~Makes the week scale more consistent and removes clutter:~~
  * ~~next() always adds 1 week to current date~~
  * ~~move majorLabels to the 1st week in the month~~
* Add week to auto-scaling (zooming)
  * replaces 5 days scale

**Screenshots:**
Before this PR: (**keep this behavior**)
week_scale_orig

~~After this PR:~~ (**reverted**)
week_scale_new

**Live examples:**
* [weekStyling](http://rawgit.com/knokit/vis/feature/week-scale-dist/examples/timeline/styling/weekStyling.html)
* [basicUsage](http://rawgit.com/knokit/vis/feature/week-scale-dist/examples/timeline/basicUsage.html) - zoom can be tested with this example

### DIFF
--- a/lib/timeline/TimeStep.js
+++ b/lib/timeline/TimeStep.js
@@ -73,7 +73,7 @@ TimeStep.FORMAT = {
     hour:       'HH:mm',
     weekday:    'ddd D',
     day:        'D',
-    week:       'w',
+    week:       'D',
     month:      'MMM',
     year:       'YYYY'
   },
@@ -324,7 +324,7 @@ TimeStep.prototype.setMinimumStep = function(minimumStep) {
   if (stepYear > minimumStep)             {this.scale = 'year';        this.step = 1;}
   if (stepMonth*3 > minimumStep)          {this.scale = 'month';       this.step = 3;}
   if (stepMonth > minimumStep)            {this.scale = 'month';       this.step = 1;}
-  if (stepDay*5 > minimumStep)            {this.scale = 'day';         this.step = 5;}
+  if (stepDay*5 > minimumStep)            {this.scale = 'week';        this.step = 1;}
   if (stepDay*2 > minimumStep)            {this.scale = 'day';         this.step = 2;}
   if (stepDay > minimumStep)              {this.scale = 'day';         this.step = 1;}
   if (stepDay/2 > minimumStep)            {this.scale = 'weekday';     this.step = 1;}
--- a/lib/timeline/TimeStep.js
+++ b/lib/timeline/TimeStep.js
@@ -222,7 +222,23 @@ TimeStep.prototype.next = function() {
       break;
     case 'weekday':      // intentional fall through
     case 'day':          this.current.add(this.step, 'day'); break;
-    case 'week':         this.current.add(this.step, 'week'); break;
+    case 'week':
+      if (this.current.weekday() !== 0){ // we had a month break not correlating with a week's start before
+        this.current.weekday(0); // switch back to week cycles
+        this.current.add(this.step, 'week');
+      } else if(this.options.showMajorLabels === false) {
+        this.current.add(this.step, 'week'); // the default case
+      } else { // first day of the week
+        var nextWeek = this.current.clone();
+        nextWeek.add(1, 'week');
+        if(nextWeek.isSame(this.current, 'month')){ // is the first day of the next week in the same month?
+          this.current.add(this.step, 'week'); // the default case
+        } else { // inject a step at each first day of the month
+          this.current.add(this.step, 'week');
+          this.current.date(1);
+        }
+      }
+      break;
     case 'month':        this.current.add(this.step, 'month'); break;
     case 'year':         this.current.add(this.step, 'year'); break;
     default: break;
@@ -532,7 +548,7 @@ TimeStep.prototype.isMajor = function() {
     case 'day':
       return (date.date() == 1);
     case 'week':
-      return (Math.ceil(date.date() / 7) == 1); // is 1st week in the month?
+      return (date.date() == 1);
     case 'month':
       return (date.month() == 0);
     case 'year':
--- a/lib/timeline/TimeStep.js
+++ b/lib/timeline/TimeStep.js
@@ -340,7 +340,7 @@ TimeStep.prototype.setMinimumStep = function(minimumStep) {
   if (stepYear > minimumStep)             {this.scale = 'year';        this.step = 1;}
   if (stepMonth*3 > minimumStep)          {this.scale = 'month';       this.step = 3;}
   if (stepMonth > minimumStep)            {this.scale = 'month';       this.step = 1;}
-  if (stepDay*5 > minimumStep)            {this.scale = 'week';        this.step = 1;}
+  if (stepDay*7 > minimumStep)            {this.scale = 'week';        this.step = 1;}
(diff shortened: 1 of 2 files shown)


lib/timeline/TimeStep.js
```diff
@@ -73,7 +73,7 @@ TimeStep.FORMAT = {
     hour:       'HH:mm',
     weekday:    'ddd D',
     day:        'D',
-    week:       'w',
+    week:       'D',
```
marcortw: I don't quite understand why to choose the "Day of Month" as default instead of "Locale week of year" when local-aware moment.js is defined as a requirement. Besides, it seems to be more safe to choose the local-aware moment.js week, you never know what happens within these locales.

lib/timeline/TimeStep.js
```diff
@@ -222,23 +222,7 @@ TimeStep.prototype.next = function() {
...
-        nextWeek.add(1, 'week');
-        if(nextWeek.isSame(this.current, 'month')){ // is the first day of the next week in the same month?
-          this.current.add(this.step, 'week'); // the default case
-        } else { // inject a step at each first day of the month
-          this.current.add(this.step, 'week');
-          this.current.date(1);
-        }
-      }
-      break;
+    case 'week':         this.current.add(this.step, 'week'); break;
```
marcortw: I absolutely disagree with this change, which causes the first day of the month to completely disappear. In fact, I believe that showing the start of the month is mandatory, it is at least for me and so far I haven't seen anybody complaining. Week starts do not necessarily align with the start of the month. With this change, there is wrong information displayed, e.g. giving the user the impression that the month starts right with the start of the week, as is visible in your initial comment for this PR #3549, e.g. in September 2016, (ISO) week 37 started on the 12th of September. Also it is not consistent, because July looks like it was 5 weeks long. That's why I went through all the hassle with this admittedly ugly piece of code. But please leave it as it is, and maybe apply your own patch or make it configurable if you don't like it.

lib/timeline/TimeStep.js
knokit: The reason for that is to be consistent with the other formats, so when your zooming it doesn't jump from number of the week in the year to day of the month.
This can then be overridden to fit specific cases using ``format: { minorLabels: {week: 'w'} }`` as shown in this example: [weekStyling](http://rawgit.com/knokit/vis/feature/week-scale-dist/examples/timeline/styling/weekStyling.html)

lib/timeline/TimeStep.js
knokit: I know that this changes a bit your initial implementation, but it's more generic and consistent with the rest of the code, i.e. creates a regular scale that iterates only weeks (as we are creating a week scale), without exceptions.

The rest is just a matter of label formatting, if the _majorLabel_ seems a bit misleading it can be changed for  _D MMM YYYY_ instead. But again, this can be modified using ``format: { majorLabels: {week: 'D MMM YYYY'} }``

It's not a matter of disliking the current solution, it's instead an attempt to make it more consistent with the rest of the code, usable for other cases and customizable.

wimrijnders:
@knokit Can you please fix the build? Travis reports the following:

```
/home/travis/build/almende/vis/test/PointItem.test.js:15
  var now = moment();
            ^

TypeError: moment is not a function
    at Suite.<anonymous> (/home/travis/build/almende/vis/test/PointItem.test.js:13:13)
    at Object.create (/home/travis/build/almende/vis/node_modules/mocha/lib/interfaces/common.js:114:19)
    at context.describe.context.context (/home/travis/build/almende/vis/node_modules/mocha/lib/interfaces/bdd.js:44:27)
    at Object.<anonymous> (/home/travis/build/almende/vis/test/PointItem.test.js:10:1)
    at Module._compile (module.js:570:32)
    at loader (/home/travis/build/almende/vis/node_modules/babel-register/lib/node.js:144:5)
    at Object.require.extensions.(anonymous function) [as .js] (/home/travis/build/almende/vis/node_modules/babel-register/lib/node.js:154:7)
    at Module.load (module.js:487:32)
    at tryModuleLoad (module.js:446:12)
    at Function.Module._load (module.js:438:3)
    at Module.require (module.js:497:17)
    at require (internal/module.js:20:19)
    at /home/travis/build/almende/vis/node_modules/mocha/lib/mocha.js:231:27
    at Array.forEach (native)
    at Mocha.loadFiles (/home/travis/build/almende/vis/node_modules/mocha/lib/mocha.js:228:14)
    at Mocha.run (/home/travis/build/almende/vis/node_modules/mocha/lib/mocha.js:514:10)
    at Object.<anonymous> (/home/travis/build/almende/vis/node_modules/mocha/bin/_mocha:480:18)
    at Module._compile (module.js:570:32)
    at Object.Module._extensions..js (module.js:579:10)
    at Module.load (module.js:487:32)
    at tryModuleLoad (module.js:446:12)
    at Function.Module._load (module.js:438:3)
    at Module.runMain (module.js:604:10)
    at run (bootstrap_node.js:383:7)
    at startup (bootstrap_node.js:149:9)
    at bootstrap_node.js:496:3
```

knokit:
@wimrijnders, not sure how can I fix this. I haven't change anything connected with the tests and they all passed when I ran them locally.

wimrijnders:
OK understood. This has happened before. The thing to do here is to retrigger the travis test by submitting a new commit.

I suggest you merge the PR with the latest `develop` branch and resubmitting. Someone has to do that anyway at some point. If that's not enough, make a tiny change somewhere and submit that.

wimrijnders:
Yay! 👍

knokit:
@wimrijnders, thanks! :+1:

wimrijnders:
Thank *you* for getting this done!

lib/timeline/TimeStep.js
marcortw: I see your point and agree that this approach is more consistent. But I'd like to suggest to at least change the major label default to D MMM YYYY if you think this change is absolutely necessary. Otherwise you get the impression of a wrong start of the month. For me, this change makes the week scale more or less useless because I really need and want to see the beginning of the month. What imho is really needed is a more flexible/independent approach to draw the major and minor labels and lines as requested in https://github.com/almende/vis/issues/3323. Please reconsider.

yotamberk: I tend to agree with @marcortw . The weeks don't always align with the month so the current behavior is much preferred to me.

knokit: Ok, I have to agreed with you guys. It's rather confusing. I'll revert this change and keep the current behavior. ;)

Can we still add the _week_ scale to the auto-scale? It could replace the 5 days scale (or keep both?) and by default its _minorLabels_ would display the "day of the month". What do you guys think?

wimrijnders:
Travis test failing not your fault. I had this zany idea to add a stress test to the unit tests, and sometimes it exceeds the time out. Remedy is same as before; otherwise I'll adjust the timeout in question.

lib/timeline/TimeStep.js
marcortw: I'm fine with that, thanks! I initially had the weeks in the auto-scale but then left it out because I thought that people might be confused by the week number, it's probably a bit of information that is not used by many people.

PR #3565, opened 2017-10-13
Network: DRY code in Label for parsing markup

wimrijnders:
This gets rid of a major eyesore for me. The accumulator object was identical for HTML and  Markdown.

In addition, the parsing has been refactored. Common elements have been DRY'd and the logic of the parsing has been made more comprehensible.

Unit tests for parsing HTML tags and Markdown are already present. These were used to verify that this refactoring is good.


macleodbroad-wf:
Suggest pre-compiling regexes as per performance tip on https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_Expressions

See attached patch.

[precompile-regex.diff.zip](https://github.com/almende/vis/files/1385036/precompile-regex.diff.zip)

wimrijnders:
OK. Went with this and expanded on it. Also code changes to make it work nice, some code cleanup.

This is starting to look a lot like [memoization](https://en.wikipedia.org/wiki/Memoization).

wimrijnders:
BTW, superfluously, long live unit tests. With the markup testing already in place, changing this was with 100% reliability.

wimrijnders:
I'm refraining with further refactoring. I still see some good thing to do here, some other time.

wimrijnders:
Any other good ideas? Looking forward to your review.

PR #3568, opened 2017-10-13
Network: Prevent crash when dataChanged is triggered during initialization

### LINKED ISSUE #3562: EdgesHandler._addMissingEdges Cannot read property 'forEach' of null
On upgrading from 4.20.1 to 4.21. I'm now seeing this error in the console.
image

This appears to be caused by https://github.com/almende/vis/pull/3516 and discussed in https://github.com/almende/vis/issues/3500 

Exception is raised at this line - https://github.com/almende/vis/blob/master/lib/network/modules/EdgesHandler.js#L468

Can someone explain if these changes are meant to cause this error and does it require different data structures now?

wimrijnders:
Fixes #3562.

Options `hidden` and `physics` can emit a `_dataChanged` event within `setOptions()`.
If this happens when setting options during the initialization of the `Network` instance, this leads
to an error thrown, because there is no DataSet instance  connected yet to the instance.

This bug was introduced in `v4.21.0`. Unit tests have been added for this case.

### DIFF
--- a/test/Network.test.js
+++ b/test/Network.test.js
@@ -369,6 +369,27 @@ describe('Network', function () {
   });
 
 
+  it('does not crash when dataChanged is triggered when setting options on first initialization ', function() {
+    // The init should succeed without an error thrown.
+    var options = {
+      nodes: {
+        physics: false   // any value here triggered the error 
+      }
+    };
+    createSampleNetwork(options);
+
+    // Do the other values as well that can cause this./
+    // 'any values' applies here as well, expecting no throw
+    options = {edges: {physics: false}};
+    createSampleNetwork(options);
+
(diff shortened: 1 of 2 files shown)


test/Network.test.js
```diff
@@ -369,6 +369,27 @@ describe('Network', function () {
...
+    var options = {
+      nodes: {
+        physics: false   // any value here triggered the error 
+      }
+    };
+    createSampleNetwork(options);
+
+    // Do the other values as well that can cause this./
+    // 'any values' applies here as well, expecting no throw
+    options = {edges: {physics: false}};
```
Stexxen: Should this physics option should be true here?

PR #3569, opened 2017-10-13
Network: Cleanup @typedef's in jsdoc comments

wimrijnders:
This is a cleanup of the jsdoc commenting. No code has been changed, only comments.

- Replaced a really big and recurring `@param` definition with a `@typedef`.
- moved all `@typedef`'s to central location `network/typedefs.js`.


wimrijnders:
@mbroad care to string this through that ts API generation tool, to see if it all works?

wimrijnders:
@mbroad, asking again: could you test this PR with the ts conversion tool? Would really like to know if all is good.

PR #3571, opened 2017-10-14
Network: Handle null data gracefully

wimrijnders:
During testing I discovered that passing `null` for network data leads to a thrown error.
This adds a guard to prevent it, plus unit tests for regression.


wimrijnders:
Crap, I accidentally added a change for #3570 here. I hope it's OK, it shouldn't affect the outcome.

PR #3573, opened 2017-10-14
Timeline: Fix for majorLabels showing wrong dates

knokit:
_majorLabels_ were being rendered with wrong date (rendered date of the next step in the scale).

**Screenshot of the issue:** (check it live [here](https://jsfiddle.net/knokit/4y30hrjb/))
timeline_majorlabels

**Changes:**
* ``TimeStep.getCurrent()`` returns a clone of current date, in order to avoid accidental modifications outside of TimeStep class
* Removed some unused variables from ``TimeAxis._repaintLabels()`` and render ``majorLabels`` with the correct date

### DIFF
--- a/lib/timeline/TimeStep.js
+++ b/lib/timeline/TimeStep.js
@@ -279,7 +279,7 @@ TimeStep.prototype.next = function() {
  * @return {Moment}  current The current date
  */
 TimeStep.prototype.getCurrent = function() {
-  return this.current;
+  return this.current.clone();
 };
 
 /**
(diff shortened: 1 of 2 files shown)


yotamberk:
Thanks for the PR but I don't see what the bug is. Where do you see the wrong date?

knokit:
Using the _"D MMM YYYY"_ format for the ``majorLabels``, you can see that they show a different dates than the ones showed in the respective _minorLabels_, as visible in the screenshot above. Should this be the correct behaviour?

Example:
**_minorLabel_** shows **1** (of April 2018), while **_majorLabel_** shows **6 Apr 2018** when it should be "1 Apr 2018"

yotamberk:
Uh ha! Ok I see it now.
Well done!

PR #3603, opened 2017-10-22
Unit tests: upgrade jsdom to v11 to fix breaking unit tests for Network

wimrijnders:
The handling of `getContext()` in `jsdom v9` was broken, and caused unit tests for `Network` to fail.
The issue has been fixed in `jsdom v11`, so I upgraded it.

This issue also deals with the fallout of this upgrade:

- Modules dependent on `jsdom` needed to be upgraded as well, notably `jsdom-global`.
- Replaced `mocha-jsdom` with `jsdom-global`. The former has not been updated to deal with the latest `jsdom` versions.
- DRY'd the `before`-code to the init method of `canvas-mock`, it was getting too unwieldly.
- Added some extra definitions to `.gitignore`, not necessarily related to current issue

**Further:**

`jsdom v11` issues a 'helpful' message that `getContext()` is not supported without `canvas`. But we've got the canvas mock in place now, which replaces that. The jsdom message does not break the unit test but is an eyesore (exception output), so effort has been made to bury it.


wimrijnders:
To be clear about it: this fix is required to make `Network` unit tests work properly.

test/canvas-mock.js
```diff
@@ -87,4 +90,54 @@ function mockify(window) {
...
+        //console.error('all is well');
+      } else {
+        // All other messages we pass through
+        console.error(msg);
+      }
+    }
+  };
+
+  // Couldn't get the [event handlers](https://github.com/tmpvar/jsdom/#virtual-consoles)
+  // for `jsdom` to work, so using the global catch instead.
```
mbroad: Indent level is incorrect

test/canvas-mock.js
```diff
@@ -87,4 +90,54 @@ function mockify(window) {
...
 }
 
+
+/**
+ * Initialize the mock, jsdom and jsdom_global for unit test usage.
+ *
+ * In particular, this takes care of a 'helpful' message about support of `getContext()`, which in
+ * practice is a complete eyesore. It's not a problem in our case, because right here we're using
+ * our own canvas mock to deal with it.
+ * The unit test run fine with or without this message, it just gets in the way.
```
mbroad: This comment could be as short as "Supresses getContext warning, ..."
And in the "..." Post the original message

test/canvas-mock.js
```diff
@@ -87,4 +90,54 @@ function mockify(window) {
...
+
+  // jsdom uses a virtual console to output its messages. The default virtual console is
+  // overridden here so that we can (selectively) suppress output.
+  const virtualConsole = new jsdom.VirtualConsole();
+
+  // This defines a simple 'mock' console output.
+  // Only 'error' is overridden which for current purposes is fine. The other calls we'll deal
+  // with when they crash, in the event that they get used.
+  let myConsole = {
+    error: (msg) => {
```
mbroad: Is Warning something we can also log?

test/canvas-mock.js
wimrijnders: the original message is an exception trace at least 15 lines long, would prefer not to add that in full. The first line of that is in the code.

test/canvas-mock.js
wimrijnders: Yes, sure. But it's not used here so YAGNI until you do.

test/canvas-mock.js
mbroad: Ok I still think this comment could be shortened considerably

test/canvas-mock.js
mbroad: Oh I see. I misunderstood and thought we were suppressing warnings by not including them. But now I understand that until a warning is invoked by a test, it's pointless to add it to the mock.

wimrijnders:
I'll fix the comments, don't approve yet.

test/canvas-mock.js
wimrijnders: mock upon mock upon mock....pretty soon it's going to be mocks all the way down.

test/canvas-mock.js
wimrijnders: OK will shorten.

wimrijnders:
@mbroad Fixed the comments, have a look again?

PR #3604, opened 2017-10-22
Moved navbar elements in all html documentation files to partial

wimrijnders:
This consolidates the common menu elements in all documentation files to a partial.

### DIFF
--- a/docs/tmpl/navbar.tmpl
+++ b/docs/tmpl/navbar.tmpl
@@ -13,7 +13,7 @@
                         <span class="icon-bar"></span>
                         <span class="icon-bar"></span>
                     </button>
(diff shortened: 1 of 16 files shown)


docs/tmpl/navbar.tmpl
```diff
@@ -14,7 +13,7 @@
                         <span class="icon-bar"></span>
                         <span class="icon-bar"></span>
                     </button>
-                    <a class="navbar-brand hidden-sm" href="./index.html">vis.js</a>
+                    <a class="navbar-brand hidden-sm" href="./index.html">vis.js!</a>
```
mbroad: !?

wimrijnders: 😆 I used that to check that the partial templates were indeed loading. Overisight, hereby fixed.

PR #3621, opened 2017-10-27
Fix outdated showcase

macleodbroad-wf:
Still not sure how we lost the showcase updates. I think it had something to do with release process... In any case, this fixes it.

Should I make a PR for these changes against develop and master as well?

@yotamberk @wimrijnders


yotamberk:
Should be done also to `develop`

PR #3622, opened 2017-10-27
Fixes issue 3397 - Updates url for gephi's json exporter plugin

macleodbroad-wf:
Not sure how this commit was reverted, but here it is again.

Should I make this PR against master & develop as well?


yotamberk:
To `develop`

PR #3626, opened 2017-10-28
Adds code-coverage badge, code coverage reports, and test-cov script …

macleodbroad-wf:
…to master.


yotamberk:
This will need to be done in `develop` too. please submit another one to develop so we don't forget. =)

Issue #2584, opened 2017-01-15
Webpage: Missing Example Screenshots


mojoaxel:
I updated the example overview pages in the gh-pages branch. There are still a view screenshots missing.

Tooa:
@mojoaxel Is this still an issue or has this been fixed? For me, all examples seem to have a screenshot.

mojoaxel:
I added the missing the timeline screenshots and updated all existing with 276175e8c41c3ddc7c0aa2a00136aecaac15f1d3.
Some screenshots are still missing for [network](http://visjs.org/network_examples.html)

PR #3647, opened 2017-11-05
Fixes Issue #2584 (again)

macleodbroad-wf:
Fixes Issue #2584 (again)

This was already approved and went in once, but was somehow reverted...
Does this need to be PR'd somewhere else as well to prevent from being reverted again??

R: @yotamberk


yotamberk:
This needs to be added to `develop` branch

PR #3652, opened 2017-11-08
BUGFIX: No group info returned to mouse events when cursorY == group top

mimafogeus2:
When the mouse was over a group-label's very top pixels, no group id was returned in mouse events.
This was due to checking if the cursor's Y coordinate is higher than the group element's top position, and not equal or higher.


yotamberk:
solves #3653

PR #3687, opened 2017-11-27
Patch 1

tasmandrake:
This change allows for add node mode to be disabled and the manipulation toolbar to be displayed when undefined or null are passed into the callback function.


mojoaxel:
@tasmandrake I'm sorry but you have to base your changes on the develop and not the master branch. Your PR would merge the master back to develop witch we sadly can not do at the moment.
Please make y new PR based on the current develop branch. Sorry for the inconvenience!

PR #3688, opened 2017-11-28
Fix for partial hidden large pop ups

carlos-olr:
I got a scenario where the pop ups get partial hided because of their content plus the location of the pop up div. So I changed the parent element that do the "appendChild" of the div and got a nice result.

I put two images here to show the new behavior

With current version:
without_fix

And with my code
with_fix

So, that is! Thanks for the vis.js

Att
Abs


yotamberk:
Thanks for this! I agree tooltips should be added to body.
Can you please resubmit this PR to the `develop` branch? We cannot accept this change directly to the `master`,

yotamberk:
I'm sorry but this won't do. 
Your change was done on a master branch which pushed changes  that are not good. 
Can you please do your changes again on a `develop` forked branch and resubmit?

carlos-olr:
I crated a new [pull request](https://github.com/almende/vis/pull/3726).

att
abs

PR #3694, opened 2017-11-29
Move showManipulationToolbar outside if statement

tasmandrake:
This change allows for add node mode to be disabled and the manipulation toolbar to be displayed when undefined or null are passed into the callback function.


yotamberk:
Awesome! Thanks for the PR

PR #3707, opened 2017-12-04
Add data attributes to line element and dot element for BoxItem

jallers:


yotamberk:
Thank you for the PR.
Can you explain why you add this?
Is there an issue this relates to?

PR #3717, opened 2017-12-10
[Timeline] Year quarters

francescogc:
Hello again :) I think the title is pretty much self explanatory: this PR adds year quarters (Q1 2018, Q2 2018...) to the timeline axis.

image

My only concern is that not everyone is interested in having quarters displayed (since it's more of a 'business' thing) so I don't know, maybe there should be some sort of flag in the options, disabled by default, that enables quarters in the axis.

Let me know if you like it!

### DIFF
--- a/docs/graph2d/index.html
+++ b/docs/graph2d/index.html
@@ -1002,7 +1002,7 @@ function (option, path) {
             <td class="indent">timeAxis.scale</td>
             <td>String</td>
             <td>none</td>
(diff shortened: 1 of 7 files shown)


francescogc:
@yotamberk thanks for the tip, now quarters are hidden by default. In order to show them, you need to add something like this in the timeline options: ` format: { minorLabels: { quarter: '[Q]Q' } }` which is totally fine with me.

PR #3726, opened 2017-12-12
Fix for partial hidden large pop ups

carlos-olr:
I got a scenario where the pop ups get partial hided because of their content plus the location of the pop up div. So I changed the parent element that do the "appendChild" of the div and got a nice result.

I put two images here to show the new behavior

With current version:
antes

And with my code
depois

So, that is! Thanks for the vis.js

Att
Abs

p.s: Sorry for the delay to submit one the correct branch.

### DIFF
--- a/lib/shared/Popup.js
+++ b/lib/shared/Popup.js
@@ -18,7 +18,7 @@ class Popup {
     // create the frame
     this.frame = document.createElement('div');
     this.frame.className = 'vis-tooltip';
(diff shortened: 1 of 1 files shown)


yotamberk:
One of the tests seems to be failing. Can you please fix it so we can accept this PR?

yotamberk:
@carlos-olr Can you please fix the test? I'd really like to add this in

PR #3728, opened 2017-12-13
fixed memory leak because eventlisteners weren't destroyed

### LINKED ISSUE #3645: vis memory leak with angular2/4
Hey there, first of all thank you for making and working on this project, it's still great!

I've got a little problem here though, probably my last issue. I'm working with Angular2/4, and vis (yes, I'm using vis as an imported script in index.html). I implemented a component, where I have to use it like: declare var vis: any; on the top of the component file, so I can use the vis functions.
With this method it works great, except one thing, every time I reload the component (not with f5, with like normal routing, imagine the following: I use vis to redraw subnetworks, and I've got like 300 subnetworks, with different routes / URLs), my whole application gets laggy more and more because there's a common bug in vis, if you reload initialize a new network, the previous one get stuck in the memory. So if I check out like 5-6 different subnetworks on my application, the result is a laggy app in the end, because the components reload, with the new initialized vis (because of the different URLs). Do you have any idea what might cause the memory leak problem?

Thank you in advance.

lennartboeckx:
In Angular (2+) there was a memory leak because the hammers with their event listeners weren't destroyed. Fixed it by destroying the hammer before they were set to null.
fixes #3645

### DIFF
--- a/lib/timeline/Core.js
+++ b/lib/timeline/Core.js
@@ -527,6 +527,7 @@ Core.prototype.destroy = function () {
     }
   }
   this.listeners = null;
+  this.hammer && this.hammer.destroy();
   this.hammer = null;
 
   // give all components the opportunity to cleanup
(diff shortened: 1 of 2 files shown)


TheMalloum:
+1

PR #3795, opened 2018-01-29
[Timeline] Vertical group labels

frastefanini:
Hi everyone,
this PR adds a global option called `groupLabelDirection` and if you set it to `vertical` guess what happens next...

image

Pretty straightforward. This time I decided **not** to introduce a group-level parameter because imo if you override only some groups behaviour the timeline turns into a mess (but let me know what you think).

Also for the CSS positioning I used flexbox, which is supported by almost everyone nowadays. Here's the full compatibility report: https://caniuse.com/#search=flex

Cheers,
Francesco

### DIFF
--- a/docs/timeline/index.html
+++ b/docs/timeline/index.html
@@ -687,6 +687,13 @@ function (option, path) {
     </tr>
 
     <tr>
+      <td>groupLabelDirection</td>
+      <td>String</td>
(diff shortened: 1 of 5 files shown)


yotamberk:
Awesome feature! Thanks!

tatsujb:
still not integrated?

PR #3814, opened 2018-02-09
fixed bug that did not allocate the correct array if there is a remov…

AlexDM0:
Hi guys,

I encountered an issue where I have a vis dataset loaded into a graph2d and on removal (.clear() method) of all items, the graph did not redraw correctly. On updating afterwards it threw an Error.

Took quite a bit of digging but it turns out the optimization of the memory allocation does not account for deleted items. the onRemove method forwards to the onUpdate method which should not be done anymore if the group update blindly expects the items to be updated, not deleted.

This is more of an indication, you don't need to accept the request. Normally, this request would go to develop but I really didnt feel like comparing develop branches :)

Cheers

### DIFF
--- a/lib/timeline/component/LineGraph.js
+++ b/lib/timeline/component/LineGraph.js
@@ -338,7 +338,7 @@ LineGraph.prototype._onAdd = function (ids) {
   this._onUpdate(ids);
 };
 LineGraph.prototype._onRemove = function (ids) {
-  this._onUpdate(ids);
+  this._updateAllGroupData(ids, undefined, true);
 };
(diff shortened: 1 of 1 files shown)


Tooa:
Can you provide an example showing the redraw regression issues between 4.21 and 4.18? We may open a ticket with your example then. Someone might pick it up to investigate.

AlexDM0:
Anything with updating a large dataset of say 4000 items multiple times a second works. Try to tax the 4.21 and then try it with 4.18.

I don’t have an example on hand.

________________________________

PR #3818, opened 2018-02-13
[Timeline] Allow normal page scrolling over timeline

### LINKED ISSUE #3734: Feature request: option to use mousewheel for scrolling window even when over timeline-DIV
Hi team,
I currently use the timeline component of visjs to display a historical overview of a wineyards which works perfect. I disabled scrolling by mousewheel without altKey (which means i use ALT as zoomKey, which works perfect) when i now scroll down the page using the mousewheel and the mousepointer is over the timeline-DIV page scrolling stops, because its captured by the timeline component.

It would be nice to have an option now to enable page scrolling by mousewheel especially when a zoomKey is specified.

my configuration:
var options = {
	locale: 'de',
	clickToUse: false,
	height: '700px',
	zoomable: true,
	moveable: true,
	selectable: true,
	zoomKey: 'altKey',
};

I am using the latest version 4.21.0 and didn't find an option.
To use the lib for our needs I commented out line 9619: event.preventDefault(); in function onMouseWheel() to have scrolling page by mousewheel.

Thanks!
Julian

lewisjb:
Adds/fixes #3734. Since I believe that this should be the expected behaviour, I don't think it requires an extra example etc, if reviewers disagree, I can add one.

As the linked ticket states, you can't scroll the page while the mouse is over the timeline (even if the timeline is doing nothing with that scroll). This PR changes it so that as-long as the timeline isn't zooming or scrolling, then the page will scroll (Timeline won't eat the event).

### DIFF
--- a/lib/timeline/Core.js
+++ b/lib/timeline/Core.js
@@ -204,8 +204,11 @@ Core.prototype._create = function (container) {
       deltaX = event.deltaX;
     }
 
-    // prevent scrolling when zoomKey defined or activated
+    // Prevent scrolling when zooming (no zoom key, or pressing zoom key)
     if (!this.options.zoomKey || event[this.options.zoomKey]) return;
+
+    // Don't preventDefault if you can't scroll
+    if (!this.options.verticalScroll && !this.options.horizontalScroll) return;
 
     // Prevent default actions caused by mouse wheel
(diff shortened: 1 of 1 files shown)


yotamberk:
Awesome! Thanks you!

PR #3819, opened 2018-02-13
Update moment to last version

tomap:
This fixes the security issue mentionned by 
https://david-dm.org/almende/vis
> https://nodesecurity.io/advisories/532

Here is the (big) changelog:
https://github.com/moment/moment/compare/b8a7fc310eb3625e83fc0c8f1ea2840fa83c7378...moment:03073778ec18ee38e179208cd1af899e8d1848dc

And the more compact one:
https://github.com/moment/moment/blob/develop/CHANGELOG.md

### DIFF
--- a/package.json
+++ b/package.json
@@ -38,7 +38,7 @@
     "emitter-component": "^1.1.1",
     "hammerjs": "^2.0.8",
     "keycharm": "^0.2.0",
-    "moment": "^2.18.1",
+    "moment": "^2.20.1",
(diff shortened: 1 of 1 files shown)


Xesenix:
@mojoaxel Why is moment bundled with this library and not as peer dependency?

mojoaxel:
> Why is moment bundled with this library and not as peer dependency?

@Xesenix That is a good question! I suspect that has historical reasons..

PR #3846, opened 2018-02-28
Adding updateClusteredNode to network.prototype

Menighin:
Just noticed `updateClusteredNode` was not added to the `Network.prototype`. Got a hard time finding out that I could access it through `network.clustering.updateClusteredNode`.

Regards,


Menighin:
I also added other functions that were missing from the network prototype. Please check my commit.

micahstubbs:
@Menighin thanks for the sending this PR with the fix!

PR #3856, opened 2018-03-08
added Array check before edgesData.forEach since edgesData can be null

ssadams11:
Discovered a small bug in EdgesHandler where this.body.data.edges was sometimes null and thus broke _addMissingEdges by attempting a forEach on null.
That this can be null is documented in EdgesHandler.js:
...line 257
  // TODO: is this null or undefined or false?
    if (this.body.data.edges) {
      // subscribe to new dataset
...

Added a simple Array.isArray() guard clause around the forEach section.
My app now works fine on 4.21.0

### DIFF
--- a/lib/network/modules/EdgesHandler.js
+++ b/lib/network/modules/EdgesHandler.js
@@ -464,14 +464,14 @@ class EdgesHandler {
      let edges = this.body.edges;
      let edgesData = this.body.data.edges;
      let addIds = [];
-
-     edgesData.forEach((edgeData, edgeId) => {
-         let edge = edges[edgeId];
-         if(edge===undefined) {
-           addIds.push(edgeId);
-         }
-     });
-     
+     if (Array.isArray(edgesData)) {
+         edgesData.forEach((edgeData, edgeId) => {
+             let edge = edges[edgeId];
+             if (edge === undefined) {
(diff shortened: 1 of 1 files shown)


micahstubbs:
this looks like a nice improvement - I'd like to get the tests green and get this in

micahstubbs:
here are the failing tests in Travis: https://travis-ci.org/almende/vis/builds/350851817?utm_source=github_status&utm_medium=notification

screen shot 2018-09-05 at 11 55 35 am

```javascript

  140 passing (25s)
  2 failing
  1) Network Edge has reconnected edges:
      AssertionError: edge is missing from state cache
      + expected - actual
      -false
      +true
      
      at Context.<anonymous> (test/Network.test.js:646:12)
  2) Network on node.js runs example  WorlCup2014:
     Error: Timeout of 15000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves.
      at Context.<anonymous> (test/Network.test.js:1296:5)
```

micahstubbs:
@ssadams11 when you have a moment, could you run the tests locally and see if you can make update the change in this PR to make these tests green?

to run the tests locally:

```bash
cd vis
npm install
npm run test
```

PR #3880, opened 2018-03-20
Adding refreshData parameter on cluster update methods

Menighin:
Following another cluster methods that use the `refreshData` parameter.
This is really useful when you want to update several edges and clustered nodes and don't call the update methods for each time you do.

### DIFF
--- a/lib/network/modules/Clustering.js
+++ b/lib/network/modules/Clustering.js
@@ -902,13 +902,16 @@ class ClusterEngine {
   * @param {vis.Edge.id} clusteredNodeId
   * @param {object} newOptions
   */
-  updateClusteredNode(clusteredNodeId, newOptions) {
+  updateClusteredNode(clusteredNodeId, newOptions, refreshData = true) {
     if (clusteredNodeId === undefined) {throw new Error("No clusteredNodeId supplied to updateClusteredNode.");}
     if (newOptions === undefined) {throw new Error("No newOptions supplied to updateClusteredNode.");}
     if (this.body.nodes[clusteredNodeId] === undefined)   {throw new Error("The clusteredNodeId supplied to updateClusteredNode does not exist.");}
 
     this.body.nodes[clusteredNodeId].setOptions(newOptions);
-    this.body.emitter.emit('_dataChanged');
+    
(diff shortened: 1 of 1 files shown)


lib/network/modules/Clustering.js
```diff
@@ -901,22 +901,27 @@ class ClusterEngine {
   * Using a clustered nodeId, update with the new options
   * @param {vis.Edge.id} clusteredNodeId
   * @param {object} newOptions
+  * @param {boolean} refreshData
```
Tooa: Can you please add the default value i.e. `[refreshData=true]` to the jsdoc?

lib/network/modules/Clustering.js
```diff
@@ -901,22 +901,27 @@ class ClusterEngine {
...
+    if (refreshData === true) {
+      this.body.emitter.emit('_dataChanged');
+    }
   }
 
   /**
   * Using a base edgeId, update all related clustered edges with the new options
   * @param {vis.Edge.id} startEdgeId
   * @param {object} newOptions
+  * @param {boolean} refreshData
```
Tooa: See above.

Menighin:
Hi @yoavdamri  and @Tooa,

So, I don't know if we should add this parameter to the documentation just yet. The other methods that have this approach, like `clusterOutliers()` and `openCluster()` are not documented as well with this parameter. When I found out about this, I thought it is because it doesn't make sense to document this behavior if we do not document how the user can manually update the network. And to document how the user can manually update the network, I think we need an API to make use of the `emitter`. Using `network.body.emiter.emit('_dataUpdated')`, as I am using now for performance reasons, seems a lot hacky and I don't think it should be documented that way.

Let me know what you guys think.
Meanwhile I commited the JSDoc fix.

Regards

yoavdamri:
@Menighin i think you meant to add yotamberk to your post and not me.
Good luck!

micahstubbs:
@Menighin thanks for the sending this new feature!  It looks like there a was some conversation about this one earlier.

Curious, could you provide a new example that we can use to verify the new feature, or point us to an existing example that shows off the new behavior?

PR #3881, opened 2018-03-21
Making getNodesInCluster() recursive

Menighin:


lib/network/modules/Clustering.js
```diff
@@ -850,15 +850,25 @@ class ClusterEngine {
   /**
    *
    * @param {Cluster.id} clusterId
+   * @param {Recursive} recursive
    * @returns {Array.<Node.id>}
    */
-  getNodesInCluster(clusterId) {
+  getNodesInCluster(clusterId, recursive = false) {
```
Tooa: What about adapting the documentation here [1]? Is there any benefit of making the actual algorithm implementation configurable? Is this something the user might care? 

[1] http://visjs.org/docs/network/index.html?keywords=getNodesInCluster

Tooa:
Hi, thanks for your contribution. Would you please add some word on why a recursive implementation is better for this use case? Have you tested your code and the previous implementation with respect to time and space complexitiy?

lib/network/modules/Clustering.js
```diff
@@ -850,15 +850,25 @@ class ClusterEngine {
   /**
    *
    * @param {Cluster.id} clusterId
+   * @param {Recursive} recursive
```
Tooa: Shouldn't this be `boolean` instead of `{Recursive}`? I would expect something like `@param {boolean} [recursive=false]`

Correct me if I'm wrong.

Menighin:
Hi @Tooa , I will make the changes requested asap.
Meanwhile about your questions, I have not make any specific stress test on this approach.

Currently I am working on what I consider to be a medium-size level-clustered network: It starts with 3 clustered that contains another clustered nodes that contains another clustered nodes ...... until it opens the "leaf" nodes. 

The recursive method comes in hand when the user clicks a clustered node that contains another clustered node. In my case, when this interaction is made, I would like to get all the leaf nodes the clicked clustered node contains. The current implementation is only returning me the clustered nodes inside the clicked clustered node.

I'm not sure if I understood right your questions, if there's any doubt left please feel free to ask!

Regards,

micahstubbs:
@Menighin thanks for the sending this PR with the fix!

PR #3888, opened 2018-03-25
Add dir support

geminoa:
This update is for fixing TODO for 'dir' support. It is to comply with behaviour of graphviz described in [1].

[1] https://www.graphviz.org/doc/info/attrs.html#h:undir_note

PR #3891, opened 2018-03-27
Add parsing penwidth attribute

geminoa:
This update is for adding 'penwidth' attribute which defines the line width as 'width'. If both are exist, 'penwidth' is used and 'width' is discarded as described in [1].

 [1] https://www.graphviz.org/doc/info/attrs.html#d:penwidth

### DIFF
--- a/docs/network/edges.html
+++ b/docs/network/edges.html
@@ -232,7 +232,7 @@ network.setOptions(options);
             <td class="indent2">arrows.to.type</td>
             <td>String</td>
             <td><code>arrow</code></td>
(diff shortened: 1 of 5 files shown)


micahstubbs:
@geminoa it looks like this PR also includes all or most of the work in https://github.com/almende/vis/pull/3888

could you untangle them, or send a fresh PR that contains only the `penwidth attribute` changes?

micahstubbs:
@geminoa, also, could you include an example that uses the new support for the `penwidth` attribute?  thanks!

geminoa:
@micahstubbs , OK, I'll update it.

micahstubbs:
@geminoa this might be helpful - here is a view that compares the `add_dir` branch to the `add_penwidth` branch

https://github.com/geminoa/vis/compare/add_dir...add_penwidth

Issue #3792, opened 2018-01-26
[Timeline]: Horizontal and Vertical scroll by mouse wheel issue in Mozilla Firefox


gam0ny:
Hi,
Thank you for the great tool! It covers really great part of functionality.

I'm experiencing an issue with timeline scroll by mouse wheel in Mozilla Firefox browser only.
The scrolling is really slow. It's clearly visible in the sample by the link:
http://plnkr.co/edit/zGzVxkhSsAngcLMH1xhV?p=preview

Could you please assist?
OS: Windows 8
FF: 58.0 (64-bit)

scriptkiddielvl2:
vis.js // New school wheel delta (wheel event) :

`var diff = delta / 120 * (this.range.end - this.range.start) / 20;`

might become as a kind of fix :

```
if (navigator.userAgent.indexOf("Firefox") >= 0) {
var diff = delta / 3 * (this.range.end - this.range.start) / 20;
}
else var diff = delta / 120 * (this.range.end - this.range.start) / 20;
```
?!

gam0ny:
Hi! Thank you so much for the idea of the fix! It really helped. Any idea how the fix can be requested in order to be included into the nearest build?

scriptkiddielvl2:
Good it helps but I would not feel comfortable pushing this quick&dirt fix into pull request. 
Mousewheel event is deprecated, moving to wheel is probably the right fix long term but much more work to achieve. /my2cents

Sujay-shetty:
@gam0ny which line or method you inserted the @scriptkiddielvl2 code

gam0ny:
@Sujay-shetty attaching screenshot:
![image](https://user-images.githubusercontent.com/20279374/36418744-d22d64c4-1640-11e8-90fb-39ba8aba691a.png)

deltragon:
Is there any update on this?

deltragon:
The issue has been fixed, but since there hasn't been a new release with it and we only use the timeline, we have now switched to @yotamberk's https://github.com/yotamberk/timeline-plus.

PR #3911, opened 2018-04-04
Normalize mouse wheel deltas

wildyeast:
This fix normalizes the mouse wheel delta values and enables reasonable scrolling  in Firefox (see #3792). Timeline now uses the [wheel event](https://developer.mozilla.org/en-US/docs/Web/Events/wheel) if supported by the browser.

Relevant examples: `timeline/other/horizontalScroll.html` and `verticalScroll.html`

### DIFF
--- a/lib/timeline/Core.js
+++ b/lib/timeline/Core.js
@@ -176,6 +176,11 @@ Core.prototype._create = function (container) {
    * @param {WheelEvent} event
    */
   function onMouseWheel(event) {
+
+    // Reasonable default wheel deltas
+    const LINE_HEIGHT = 40;
+    const PAGE_HEIGHT = 800;
+
     if (this.isActive()) {
       this.emit('mousewheel', event);
     }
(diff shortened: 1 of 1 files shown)


deltragon:
This drops support for older IE versions (6, 7, 8) since it always uses "addEventListener" which is supported from IE 9 upwards, instead of "attachEvent" like previously.

lib/timeline/Core.js
```diff
@@ -241,15 +257,12 @@ Core.prototype._create = function (container) {
...
-  } else {
-    // IE 6/7/8
-    this.dom.centerContainer.attachEvent("onmousewheel", onMouseWheel.bind(this));
-  }
+  // Add modern wheel event listener
+  const wheelType = "onwheel" in document.createElement("div") ? "wheel" : // Modern browsers support "wheel"
+    document.onmousewheel !== undefined ? "mousewheel" : // Webkit and IE support at least "mousewheel"
+    this.dom.centerContainer.addEventListener ? "DOMMouseScroll" : // Older Firefox versions like "DOMMouseScroll"
+    "onmousewheel" // All the rest should be comfortable with "onmousewheel"
+  this.dom.centerContainer.addEventListener(wheelType, onMouseWheel.bind(this), false);
```
deltragon: Use ``attachEvent`` here in case ``addEventListener`` is not supported (IE < 9).

wildyeast: That's right. Thank you!

SpeedHighway:
I'm trying this in firefox quantum, and it isn't having any effect at all....   It still moves in micrometers when scrolling with the mousewheel, or dragging the scrollbar, etc.

deltragon:
@SpeedHighway:
Hmm
Are you sure you checked out the develop branch/rebuilt it correctly? Can confirm that the mousewheel works properly in Firefox 59 on Linux.

SpeedHighway:
Actually, no.  I just tried this fix, by itself, because we're using vis through npm  (I was trying to make as small as a change as possible.)  Let me try the entire branch.

wildyeast:
Nice! Thanks for giving it a try :+1:

yotamberk:
Hi @wildyeast ,

deltragon:
@yotamberk We have switched to your fork. Thank you for your time!

PR #3936, opened 2018-04-17
Fix visibility of scale lines after toggling left axis groups

kbasten:
Fixes visibility of horizontal scale lines after toggling groups on the left axis. Its functionality can be checked by running the following example:
https://gist.github.com/kbasten/de0fd0d2d264d0f6c8d6086f834ed959
This example does not show scale lines after hiding, then displaying the left groups in previous versions of vis.


yotamberk:

PR #3943, opened 2018-04-23
Fixed the two broken references to SweetAlert

sandebert:
The urls I used where the best ones I managed to find. (I'm sure there are better ones somewhere, but it was surprisingly difficult finding them.)

### DIFF
--- a/examples/timeline/editing/editingItemsCallbacks.html
+++ b/examples/timeline/editing/editingItemsCallbacks.html
@@ -10,8 +10,8 @@
     }
   </style>
 
-  <script src="http://t4t5.github.io/sweetalert/dist/sweetalert.min.js"></script>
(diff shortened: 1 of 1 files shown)


sandebert:
Sorry, I should have mentioned this relates to http://visjs.org/examples/timeline/editing/editingItemsCallbacks.html, which is broken due to 404s for the SweetAlert references.

sandebert:
Oh, I made a bad assumption here. I just assumed the files had changed locations and that I just needed to update the references so I just updated the urls and PR'd.

So what do you say, are these reasonable urls instead?

- https://cdn.jsdelivr.net/npm/sweetalert@1.1.3/lib/sweetalert.min.js
- https://cdn.jsdelivr.net/npm/sweetalert@1.1.3/dist/sweetalert.css

Jogai:
Yes, I would say so.

yotamberk:
hi @sandebert 

PR #3965, opened 2018-05-14
edit edge: restore original controlNodes position when dragged into nothing

dropletzz:
When editing an edge, if it's dragged into something that is not a node, the position of the control nodes will not be immediately updated, ending up with something like this. 

screen shot 2018-05-14 at 11 02 49

As far as I can tell, this PR should fix that.

PS: fiy, I am using vis at work in order to create some kind of a graph editor. This is why I would like to change the `ManipulationSystem` in order to make it UI-indipendent, or at least to add the possibility for it to be UI-indipendent, so that it could be easily integrated with custom UI elements. Any help, suggestions or insights  would be very welcome (:
EDIT: found out it can already be made independent of the menu buttons, I'll just add some functionality to it

### DIFF
--- a/lib/network/modules/ManipulationSystem.js
+++ b/lib/network/modules/ManipulationSystem.js
@@ -992,6 +992,13 @@ class ManipulationSystem {
       }
     }
     else {
+      let controlNodeFrom = this.body.nodes[this.temporaryIds.nodes[0]];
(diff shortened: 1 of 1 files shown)


micahstubbs:
@dropletzz thanks for the contribution, this looks nice. can you provide an example of this improvement in action or point us to an existing example that demonstrates the new behavior?

PR #3975, opened 2018-05-21
fix: Network edges `dashes` isArray comparison, for inheritied styling options.

travomic:
This fixes being able to declare a default edge dash-array via: `edges.dashes` in the options. 

Something in the guts of the option values converts arrays into array-like objects, that do not pass the `Array.isArray` comparison.  The change in this pull-request loosens the comparison to ensure that the dashes property exists, and has `.length`.

### DIFF
--- a/lib/network/modules/components/edges/util/EdgeBase.js
+++ b/lib/network/modules/components/edges/util/EdgeBase.js
@@ -110,7 +110,9 @@ class EdgeBase {
   _drawDashedLine(ctx, values, viaNode, fromPoint, toPoint) {  // eslint-disable-line no-unused-vars
     ctx.lineCap = 'round';
     let pattern = [5,5];
-    if (Array.isArray(values.dashes) === true) {
+
+    // Loose "isArray" comparison, to handle default inherited dashes.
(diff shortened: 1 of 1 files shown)


lib/network/modules/components/edges/util/EdgeBase.js
```diff
@@ -110,7 +110,9 @@ class EdgeBase {
   _drawDashedLine(ctx, values, viaNode, fromPoint, toPoint) {  // eslint-disable-line no-unused-vars
     ctx.lineCap = 'round';
     let pattern = [5,5];
-    if (Array.isArray(values.dashes) === true) {
```
travomic: This comparison is ONLY truthy from direct edge object property values.  The inherited values from the `options.edges` property, somehow get converted into an array-like object, that Array.isArray does not recognize, and as shown below, when using `JSON.stringify` to inspect values.dashes it spits out an empty object.

screen shot 2018-05-21 at 12 40 37 pm

austinkelleher:
I am also seeing this issue. Can we please get this merged @yotamberk @mojoaxel?

PR #3976, opened 2018-05-22
fix: Prevent trying to read the value of a null element

mokaddem:
Hello,
In one of my project, I need to dynamically update the timeline where items may loose their ``end`` time. As it is not possible to simply delete the ``end`` key then update the timeline (see steps to reproduce), I have to set it to either ``null`` or ``undefined``. Doing so can, in some case, throw an error.

This fix adds the condition that the end time is not set to ``null`` before calling the function ``valueOf()``

Steps to reproduce:
- Create a item with both ``start`` end ``end`` set
- Set ``item.end = null`` and update the timeline
    - Setting it to ``null`` is required because if we remove the ``end`` key, it will be reset to the old value during the update process
- An error is raised because ``itemData.end === null``

What are your thoughts about it?

### DIFF
--- a/lib/timeline/Timeline.js
+++ b/lib/timeline/Timeline.js
@@ -425,7 +425,7 @@ Timeline.prototype.focus = function(id, options) {
   var end = null;
   itemsData.forEach(function (itemData) {
     var s = itemData.start.valueOf();
-    var e = 'end' in itemData ? itemData.end.valueOf() : itemData.start.valueOf();
+    var e = 'end' in itemData && itemData.end !== null ? itemData.end.valueOf() : itemData.start.valueOf();
 
     if (start === null || s < start) {
       start = s;


brichbe:
I don't have a comment about the validity of this PR, but I have a comment about what was probably causing that error. It sounds like your handling of that use case was invalid.

If your timeline items initially have both `start` and `end` times set, that probably means those items' `type` is set as 'range' or 'background' (per the [docs ](http://visjs.org/docs/timeline/#items) these types require both start and end).

Later, when you remove the `end` of an item and try to update the timeline, I would expect this to cause an error. That's because you must also change the `type` of that item to a 'box' or 'point' type (again, per the docs these types only need a start).

You should try changing the item's type as described above, along with removing the item's end, before updating the timeline and see if the error no longer occurs. If so, maybe this PR isn't necessary.

mokaddem:
Thanks for your input.
Indeed, enforcing the type to ``box`` solve the visual issue (and the bug). However, if later on I retrieve the ``end`` of the item, it will be set to its previous value. e.g.:

```
var item = data.get(1); // {id: 1, start: 0, end: 10};
delete item.end; // {id: 1, start: 0};
data.update(item);
var item = data.get(1); // {id: 1, start: 0, end: 10};
```

I guess I will have to manually set items' ``type`` and when querying their ``end`` time, will have to check their ``type`` as well. Which adds an overhead and code complexity as the library already deduce the item's ``type`` by itself.

brichbe:
No, I think there's still a better way. You've reminded me that calling the DataSet.update() method actually merges the new data with the previous data (if it already existed). This makes sense for most use cases, but not for yours, since you're removing the `end` property.

Try the following:

```
// Change a range item with start/end to a box item with only start 
var item = data.get(1);
data.remove(item);
delete item.end;
item.type = 'box';
data.add(item);

mokaddem:
Actually, switching from a range to a box is done in a function generating item by item (but not adding them to the dataset). That's why I didn't wanted to remove/add.
Still, this solution is cleaner than checking the type. :)

But I find it weird that the constructor supports ``end`` set to ``null``, but updating with a ``null`` causes errors. So, if it is good practice to remove then add for this case, you can close this PR.

PR #3978, opened 2018-05-26
Replace babel-preset-es2015 with babel-preset-env

bradh:
This replaces the previous custom module with the more "framework-like" env approach, as shown at https://babeljs.io/env/

I admit that the main reason for this was the make the `npm install` step slightly less chatty. It makes no difference (that I know of) to anything other than custom builds.


yotamberk:
Thanks for the contribution!

PR #3990, opened 2018-06-07
Fix ordering for arbitrarily nested groups

### LINKED ISSUE #1934: [Timeline]Support group hierarchy
Does the group label support hierarchy which looks like http://dhtmlx.com/docs/products/dhtmlxGantt/01_basic.html

or is there a way base on what we have to let user toggle a group(subgroup)?

trainman419:
Fix ordering for arbitrarily nested groups. Fixes #1934


followyouhome:
Could anybody review this one?

yotamberk:
Hi @trainman419 ,

PR #3991, opened 2018-06-07
Network interaction: fire (de)select event when node is dragged

dropletzz:
The node being dragged would be selected but no (de)select events would be fired.
In addition, now, if multiselect is on and an unselected node is dragged while pressing Ctrl it will be added to the selection.


micahstubbs:
@dropletzz thanks for the contribution, this looks nice.  can you provide an example of this improvement in action or point us to an existing example that demonstrates the new behavior nicely?

Issue #3907, opened 2018-04-03
Network: Not hover node when move cursor from the edge 


eshlykov:
Hi,

If I hover a node moving cursor along an edge, hover will still be on the edge instead of the node. You can see it using the [example page](http://visjs.org/examples/network/nodeStyles/colors.html).

I use Chrome Version 65.0.3325.146 (Official Build) (64-bit) on Windows 10 or Ubuntu 16.04.

jobec:
I'm hitting the same bug...
Would be great if @Gelio 's patch could be merged...

PR #4027, opened 2018-07-05
Fix node hover when mouse moves from an edge

Gelio:
When the mouse enters a node while hovering over an edge previously, the node hover event is not emitted.

Here are the screenshots of this unintended behavior. I have marked the mouse position with a black dot.
screenshot from 2018-07-05 17-00-12
screenshot from 2018-07-05 17-00-14

This PR fixes that issue. It has been also mentioned in #3907, which this PR also fixes.

### DIFF
--- a/lib/network/modules/SelectionHandler.js
+++ b/lib/network/modules/SelectionHandler.js
@@ -14,8 +14,8 @@ class SelectionHandler {
   constructor(body, canvas) {
     this.body = body;
     this.canvas = canvas;
-    this.selectionObj = {nodes: [], edges: []};
-    this.hoverObj = {nodes:{},edges:{}};
+    this.selectionObj = { nodes: [], edges: [] };
+    this.hoverObj = { nodes: {}, edges: {} };
 
     this.options = {};
     this.defaultOptions = {
@@ -38,8 +38,8 @@ class SelectionHandler {
    */
   setOptions(options) {
     if (options !== undefined) {
-      let fields = ['multiselect','hoverConnectedEdges','selectable','selectConnectedEdges'];
-      util.selectiveDeepExtend(fields,this.options, options);
+      let fields = ['multiselect', 'hoverConnectedEdges', 'selectable', 'selectConnectedEdges'];
+      util.selectiveDeepExtend(fields, this.options, options);
     }
   }
 
@@ -104,7 +104,7 @@ class SelectionHandler {
     let properties = {};
 
     properties['pointer'] = {
-      DOM: {x: pointer.x, y: pointer.y},
+      DOM: { x: pointer.x, y: pointer.y },
       canvas: this.canvas.DOMtoCanvas(pointer)
     };
     properties['event'] = event;
@@ -213,9 +213,9 @@ class SelectionHandler {
   _pointerToPositionObject(pointer) {
     let canvasPos = this.canvas.DOMtoCanvas(pointer);
     return {
-      left:   canvasPos.x - 1,
-      top:    canvasPos.y + 1,
-      right:  canvasPos.x + 1,
+      left: canvasPos.x - 1,
+      top: canvasPos.y + 1,
+      right: canvasPos.x + 1,
       bottom: canvasPos.y - 1
     };
   }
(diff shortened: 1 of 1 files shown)


lib/network/modules/SelectionHandler.js
```diff
@@ -653,7 +653,14 @@ class SelectionHandler {
     }
 
     if (object !== undefined) {
-      hoverChanged = hoverChanged || this.emitHoverEvent(event, pointer, object);
+      const hoveredEdgesCound = Object.keys(this.hoverObj.edges).length;
```
Tooa: `hoveredEdgesCound` -> `hoveredEdgesCount`

Gelio: Good catch, thank :+1: I have fixed the typo

micahstubbs:
@yotamberk could you tag this on as network when you have a moment?

micahstubbs:
@Gelio thanks for the sending this PR with the fix!

Curious, could you provide a new example that we can use to verify the fix, or point us to an existing example that shows the bug behavior (and would also show the fix)?

Gelio:
@micahstubbs Sure thing

[events/interactionEvents.html](http://visjs.org/examples/network/events/interactionEvents.html)

1. Open the console
2. Set the filter to `hoverNode` (so there are no unnecessary logs)
3. Hover over an edge
4. Move the mouse towards a node (while hovering an edge)
5. Enter the node, still hovering an edge

The `hoverNode` event will not be emitted. It is emitted when the mouse enters the node from _empty canvas space_, not from an edge

Gelio:
@micahstubbs It is not the *extra* `hoverNode` event. It is the lack of it :smile: Notice that there is no `hoverNode` event when you enter the node while following the edge

micahstubbs:
@Gelio acknowledged 👌 

I notice this desired behavior even before I apply the changes in this PR.  this makes me wonder if it is already fixed somehow 🤔

micahstubbs:
@Gelio thanks for patiently explaining - I think I was confused before and perhaps understand now 💡 

does this sound correct to you?

1) the desired behavior is to emit a `hoverNode` event anytime a node is hovered over. (this means if the node is entered from the background or from an edge)

2) the bug behavior is that when a node is hovered over after the cursor was hovering an edge, the expected `hoverNode` event is missing.

Gelio:
@micahstubbs No problem :smile: I am glad I can help

Yup, that sounds just about right. Both 1 & 2 are correct.

I have a custom popup (as a regular HTML component overlayed on the canvas) and it is shown upon `hoverNode` event. When the user hovers a node having been previously hovering an edge, the popup does not appear because of the lack of the `hoverNode` event. That is why I thought of it as a bug. Would you agree :slightly_smiling_face:?

micahstubbs:
ok, now that I understand the bug, let's take another look at the fix 😄

micahstubbs:
ok, after I remove the linter/code style changes, this is the substantive change that remains:

screen shot 2018-09-05 at 10 27 13 am

micahstubbs:
here's another look at the clean diff https://github.com/visjs-community/visjs-network/pull/29/files

screen shot 2018-09-05 at 10 32 40 am

micahstubbs:
✅ looks good to me.  I'm able to see the desired behavior after applying the changes in this PR. 
 when I hover over an edge, and then hover over a node, the `hoverNode` event is emitted.  nice work @Gelio 👍 

http://127.0.0.1:8080/examples/network/events/interactionEvents.html
vis-pr-4027-after

Gelio:
@micahstubbs Awesome! Thank you for testing my work so thoroughly :+1:

micahstubbs:
I've merged this in the community standalone network module project, and published a patch version with the fix up to npm 🎉 

https://github.com/visjs-community/visjs-network/pull/29
https://www.npmjs.com/package/visjs-network

mojoaxel:
:love_letter: Thanks @Gelio for your contribution!
This pull-request has been merge into https://github.com/visjs/vis-network/pull/22

Issue #2475, opened 2016-12-18
Is there a way to "fitVertical()" the Timeline?


pauljayd:
I would like to be able to execute a function to casue the Timeline to expand such that no items are out-of-view (vertically) at the current date. IS there any way to do this?

Thanks

yotamberk:
What do you mean by expand? The height becomes bigger?

pauljayd:
THanks for the quick response -- sorry I wasn't clear:
I would like to be able to execute a function such that given the current view (height & width) and centered date, the zoom would be set such that no items are out of view vertically (which would require either dragging up/down or scrolling). 
If no item is then visible, perhaps an option would be to move forward/backward to the nearest date with an item.
Does this make sense?
Thanks.

yotamberk:
Sorry but I'm still not clear on what you mean... Do you mean assuming you have many groups and there are items only on the center groups, that the timeline will focus on them?
The `fitVertical` option doesn't make sense because there is no vertical scaling done. Maybe you mean `verticalFocus`. It still isn't clear to me the purpose of it.

pauljayd:
Sorry again that I'm not making myself clear.
Given a Timeline with a fixed height and zoom level, some items may be vertically out-of-view or cut (see image 1):
![image1](https://cloud.githubusercontent.com/assets/5870432/21298137/80c93d3e-c55a-11e6-9762-7ad01983141f.png)

Notice the blue box-shadow at top, where item 2 is cut off.

At this point, I would like to execute a function, say "fitVertical()" such that the
zoom level is adjusted so that all out-of-view vertical items are brought into view (see image2):
![image2](https://cloud.githubusercontent.com/assets/5870432/21298149/a0d24fa8-c55a-11e6-9cfe-e8159bb0e9b7.png)

Thus, the box shadow has disappeared, and of course the zoom has changed.
I want this to be able to present the timeline without forcing the user to scroll or drag (at
least in the initial view).
Does this make sense?

Paul

yotamberk:
much clearer! Thanks! No way to do this yet. I'll add this as a feature request!

yotamberk:
I've now branched out to a new fork of my own including only Timeline.
I've applied a fix for this here: https://github.com/yotamberk/timeline-plus
and can be used via npm here: https://www.npmjs.com/package/timeline-plus

PR #4038, opened 2018-07-17
Timeline .focus() should also zoom in as well as zoom out. Fix vertical scroll when not using groups or rtl = true

### LINKED ISSUE #3389: Timeline - Select and focus are not working properly with vertical scroll
There is an option to select and focus an item in timeline:
http://visjs.org/examples/timeline/interaction/animateWindow.html

**Problem 1:** If there is a vertical scroll(Have many resources). I am expecting the timeline to not just horizontally scroll to the element but also vertically scroll to the element position.

**Problem 2:** If the item is very zoomed-out(very small in size) and there are many other items around it. I can't easily find it. I am expecting it to zoom-in little bit so that the item will be having some minimum width which makes it more visible to the user.

Is there any work around or any small change to plugin I can make to achieve this **two functionalities.**

Need: I am displaying many items of smaller in size and to many come one after another making this select and focus functionalities useless for navigation in timeline.

nathanboktae:
Fixes #3389, #2475

### DIFF
--- a/lib/timeline/Timeline.js
+++ b/lib/timeline/Timeline.js
@@ -463,7 +463,7 @@ Timeline.prototype.focus = function(id, options) {
         // The redraw shifted elements, so reset the animation to correct
         initialVerticalScroll = verticalScroll;
         startPos = me._getScrollTop() * -1;
-      }      
+      }
 
       var from = startPos;
       var to = initialVerticalScroll.scrollOffset;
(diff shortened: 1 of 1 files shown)


lib/timeline/Timeline.js
```diff
@@ -492,13 +492,13 @@ Timeline.prototype.focus = function(id, options) {
...
 
-      // Let the redraw settle and finalize the position.      
+      // Let the redraw settle and finalize the position.
       setTimeout(setFinalVerticalPosition, 100);
     };
 
     // calculate the new middle and interval for the window
     var middle = (start + end) / 2;
-    var interval = Math.max(this.range.end - this.range.start, (end - start) * 1.1);
+    var interval = (end - start) * 1.1;
```
nathanboktae: `Math.max` ensured a zoom out, but never a zoom in, which is odd. make the behavior more consistent by just snapping to the item with 5% buffer on each side.

yotamberk:
Can you resubmit this PR implemented in the new fork I've moved to here: 
https://github.com/yotamberk/timeline-plus
this fork is more updated and gets released with weekly updates

yotamberk:

PR #4044, opened 2018-07-21
Add option to maintain selection order of selected nodes and edges

softwareCobbler:
**Thank you for contributing to vis.js!!**

Please make sure to check the following requirements before creating a pull request:

* [ ] Update the documentation if you introduced new behavior or changed existing behavior.
* [ ] Provide an additional or update an example to demonstrate your changes or new features.

### DIFF
--- a/lib/network/modules/SelectionHandler.js
+++ b/lib/network/modules/SelectionHandler.js
@@ -14,12 +14,14 @@ class SelectionHandler {
   constructor(body, canvas) {
     this.body = body;
     this.canvas = canvas;
-    this.selectionObj = {nodes: [], edges: []};
+    this.insertionOrderSelectionObj = {nodes: [], edges: []};
+    this.selectionObj = {nodes: {}, edges: {}};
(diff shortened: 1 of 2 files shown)


softwareCobbler:
Hello, I am new to git, but I had to change a few lines of Vis.js to get it to stop sorting the nodes I clicked on, and figured I would try to learn some more git by publishing these minor changes.  
The problem I was trying to solve was on my rendered DAGs (heirarchical tree layout), the order of selected nodes was not preserved--for example, click node 2 and then node 1 would result in an ordered list of [node 1, node 2]. I needed to preserve their selected order, so here we are.

I see I didn't pass the "codeclimate" tests but I don't think I touched 99.9% of the code it is unhappy with.

Thanks for your time, this is a neat project.

micahstubbs:
@yotamberk could you tag this one as `network` when you have a moment? thanks!

micahstubbs:
@softwareCobbler thanks for the contribution!

Issue #4079, opened 2018-08-13
Network module LayoutEngine re-seeds random for each node adding


soinio:
When adding nodes to network (via DataSet.add) LayoutEngine.positionInitially is used to give them initial positions (if they are not fixed etc.). On version 4.21.0 positionInitially method re-sets the random seed every time. The end result is, that if the same number of nodes are added several times, the nodes always end up on top of each other. E.g. if adding one node at time, they will all end up in the exact same position.

I suppose that the "initialRandomSeed" is only supposed to be used at engine construction/configuration time, not every time something is added to the graph.

I have a proposed fix ready.

PR #4080, opened 2018-08-13
Proposed fix for bug 4079. Seed LayoutEngine random only at the beginning.

soinio:
Proposed fix for bug #4079. Makes LayoutEngine seed it's random only at construction or configuration time, not every time nodes are added.


micahstubbs:
@yotamberk could you tag this on as `network` when you have a moment?

micahstubbs:
@soinio thanks for the sending this PR with the fix!

PR #4098, opened 2018-08-29
GWT interoperability / usage of typeof instead instanceof

daRoof:
When using the timeline from GWT, type determination for provided functions via the instanceof operator does not work.
Changed setOptions method to use typeof instead instanceof.

### DIFF
--- a/lib/timeline/component/ItemSet.js
+++ b/lib/timeline/component/ItemSet.js
@@ -447,7 +447,7 @@ ItemSet.prototype.setOptions = function(options) {
     var addCallback = (function (name) {
       var fn = options[name];
       if (fn) {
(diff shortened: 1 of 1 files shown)


yotamberk:
Thank @daRoof !
This is not the only use-case of using `instanceof`.
Is there any reason for changing only this function to use `typeof` instead of `instanceof`?

In any case:

daRoof:
I am writing a GWT wrapper for the timeline component. Until now, there has only been this one occurance of instanceof that I had problems with.

As far as I understand, the problem with instanceof is caused by complex objects (a function in this case) passed from GWT code to the timeline. I think this happens only when setting options on the timeline.
But I have not included all features of the timeline, yet. So there may be further pull requests from me.

PR #4110, opened 2018-09-05
fix(timeline): destroy hammer bindings of DragCenter and DeleteButton

ShynRou:
Hey,

Selecting an item in the timeline currently adds 6 listeners each time to the window element but never removes them. Even a few dozen have a huge effect on performance.

I added the destruction of the hammer.js bindings, which fixes the issue.

Best Regards,
Dustin


yotamberk:

PR #4117, opened 2018-09-09
Allow setting title of customtime in constructor

kbasten:
This allows setting of custom time bar title in the constructor, allowing calls like the following to correctly display the title.

`graph.addCustomTime(new Date(1531135747 * 1000), 'custom_id', { title: 'custom title' });`


yotamberk:

PR #4122, opened 2018-09-12
Graph3d rotating axis labels

eanar:
This feature is to help improve the readability of long axis label names. The option to use rotating axis labels could benefit some visualizations where long axis label names would be preferred when presenting the data.  An example and small data set is provided.

[rotateAxisLabels](https://youtu.be/SLQffJNhNHw)


eanar:
I see the typos: '_msg2' and '_msg' in Graph3d.js should both be 'msg'.  Should I fix this locally and submit another PR?

PR #4143, opened 2018-09-30
[Timeline] Addittional fix for page scrolling over timeline

Justabug:
One more improvement (I hope so) for #3734.
After this fix, Timeline only does preventDefault on a vertical scroll event if the timeline has actually been scrolled. AFAIK this behaviour would be consistent with what all popular GUI systems do.


yotamberk:
Can you please submit this change to https://github.com/yotamberk/timeline ?
I am only maintaining timeline changes there.

Justabug:
OK, I'll do it a bit later

hypery2k:
if anyone just needs this scroll bugfix, i published a fork on npm: https://github.com/holisticon/vis

Issue #3906, opened 2018-04-03
Network: Keyboard navigation blocks useful keyborad shortcuts


eshlykov:
Hi,

If I activate keyboard navigation (`options.interaction.keyboad = true`), some keyboard shortcuts won't work. You can see it using the [example page](http://visjs.org/examples/network/other/navigation.html).

Some of blocked shortcuts:
- Alt + ←
- Alt + →
- Ctrl + Shift + Page Up for 
- Ctrl + Shift + Page Down

I use Chrome Version 65.0.3325.146 (Official Build) (64-bit) on Windows 10 or Ubuntu 16.04.

PR #4176, opened 2018-10-28
handle preventDefault per-keybind #3906

adamschachne:
By default, all keybinds are having their defaults prevented. This PR only calls preventDefault on keys that are not combined with ctrl, shift, or alt, because those are modifiers for normal browser hotkeys.

The same should probably be done for scroll events, considering that ctrl+scrollwheel will simultaneously zoom in/out on the canvas as well as the window. Perhaps there should be an interaction option that allows which keybinds should propogate?

Looking forward to feedback.

### DIFF
--- a/lib/network/modules/components/NavigationHandler.js
+++ b/lib/network/modules/components/NavigationHandler.js
@@ -120,26 +120,50 @@ class NavigationHandler {
   }
 
   /**
-   *
-   * @param {string} action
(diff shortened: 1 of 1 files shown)


YakovL:
FYI, Vivaldi supports one-key shortcuts, so I'm not sure this is a proper suggestion. Does this default prevention cause any actual problems?

PR #4183, opened 2018-11-04
Added option hierarchical.layout.userControlsFreeAxis (revision 2)

softwareCobbler:
Sorry for polluting your pull requests, this is a revised version of my pull request from yesterday containing only the relevant commit.

This commit solves a question I asked on stackoverflow: https://stackoverflow.com/questions/53127971/how-can-i-set-node-positions-in-a-hierarchical-layout

Here we override the `LayoutEngine` logic if a user has BOTH specified this option and supplied a "free axis" value in any of their nodes within the `Dataset` they've bound to `Network`. The "free axis" is `x` if `hierarchical.direction` is `"DU"` / `"UD"`, and `y` if it's `"LR"` / `"RL"`. Undefined values for the free axis leave `LayoutEngine` doing what it did prior to this commit.

In addition I added some documentation for this in './docs/network/layout.html'.

* [ ] Provide an additional or update an example to demonstrate your changes or new features.
* [x] Update the documentation if you introduced new behavior or changed existing behavior.
* [x] Expect review comments and change requests by reviewer.

### DIFF
--- a/docs/network/layout.html
+++ b/docs/network/layout.html
@@ -66,7 +66,8 @@ var options = {
       edgeMinimization: true,
       parentCentralization: true,
       direction: 'UD',        // UD, DU, LR, RL
-      sortMethod: 'hubsize'   // hubsize, directed
+      sortMethod: 'hubsize',   // hubsize, directed
+      userControlsFreeAxis: false
     }
   }
 }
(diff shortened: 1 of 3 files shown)


softwareCobbler:
I am considering changing the test to check if `targetValue` is not `undefined` from a strict (`!==`) to a loose (`!=`), to allow the expression to return `true` for both values `undefined` and `null`. Is there a standard pattern I should use here?

ref line 796 in lib/network/modules/LayoutEngine.js (1788c67a8925a2495e19f1da5f7f23c5b1b8bf72)

lib/network/Network.js
```diff
@@ -170,6 +170,21 @@ Network.prototype.setOptions = function (options) {
 
     // the hierarchical system can adapt the edges and the physics to it's own options because not all combinations work with the hierarichical system.
     options = this.layoutEngine.setOptions(options.layout, options);
+    console.log(this.layoutEngine);
+    //
+    // this clears "userControlsFreeAxis" when "hierarchical" is disabled
+    // without doing this, and performing the following steps:
```
softwareCobbler: May still be a good idea to clear the `userControlsFreeAxis` option if `hierarchical` layout is disabled, but the bug I was attempting to fix is probably #4228.

Issue #4227, opened 2019-01-17
Timeline rolling mode not working


Superlukec:
The demo for rolling mode functionality is not working:
http://visjs.org/examples/timeline/interaction/rollingMode.html

I've checked the source and the problem is that the root element (.vis-timeline) is always hidden as soon as you add option for rollingMode. If you open DevTools and set element to visible it works normally.

After looking for possibilites, I've tried the previous version (4.20.1) with the same code and it works.

PR #4231, opened 2019-01-23
Fix for #4227

softwareCobbler:
Attempted fix for #4227

### DIFF
--- a/lib/timeline/Timeline.js
+++ b/lib/timeline/Timeline.js
@@ -75,6 +75,9 @@ function Timeline (container, items, groups, options) {
       domNode = domNode.parentElement;
     }
     this.options.rtl = (directionFromDom && (directionFromDom.toLowerCase() == "rtl"));
+
+    // restore element visibility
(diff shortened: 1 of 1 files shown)


jgorene:
Hello there,
I'm following some pull requests to see where I can help to move forward on v4.22.

After work around for a while on this problem (it's yet a huge thing to understand... vis... for me;), I ended up looking at the patch applied to timelineplus by @yotamberk  as suggested by @avrahamcool :wink:

In the end, it works well too (there was no reason for it not to work), but it still requires testing of course.

In fact, you just have to change the code between lines 164 to line 192 on Timeline.js and that's enough (a priori).
But I'm still struggling with git and do not know how to proceed so as not to have to make a new PR...

asiby:
Hi guys. Not offence, but I think it would help to just grab the fix from the timeline-plus project that got forked from this one. It is at https://github.com/yotamberk/timeline-plus

They have the rollingMode working just fine.

Issue #4233, opened 2019-01-24
[Network] Mouse events not working


rafaelsaback:
Hi all,

I wanted to use mouseDown and mouseUp events in the Network module's canvas, but apparently those events are being caught by Hammer.js.

I noticed that by removing the flag **preventDefault: 'mouse'** the events occur normally. But I guess there's a real reason for it to be set the way it's.

In the documentation, I didn't see anything about mousedown and mouseup events, so I assume they are simply not implemented for this module.

I've seen that [another PR](https://github.com/almende/vis/pull/3059) has implemented this in the Timeline module. I tried to implement those events for the network module as well, but no success so far. You can see what I've done here in this [personal PR](https://github.com/rafaelsaback/vis/pull/1).

Any help/feedback is highly appreciated.

jgorene:
Hello,
To find out a little more, do you have an example of use.
Just to see how I could use this event in the canvas of the Network module?
Thank you very much

rafaelsaback:
Hi @jgorene,

I want to implement an area selection functionality inside the network canvas, in that way by clicking and moving the mouse I'd be able to select multiple nodes. For that, I need to listen to mousedown, mousemove and mouseup events. The mousemove event is working, but mousedown and mouseup are not.

To emulate the issue with mousedown, I have created a basic [JSFiddle using the original Vis library](https://jsfiddle.net/obhj2cdL/1/). At the end of the JavaScript window, you'll see that I define 2 events, one for onpointerdown, which works, and one for onmousedown, that does not work. You can comment/uncomment the events to test them. 

I also created a [2nd JSFiddle](https://jsfiddle.net/nem6gfq3/) that uses my modified version of Vis library, where I tried to implement the mousedown event.

Thanks!

softwareCobbler:
You want to be able to click and drag a rectangle, and upon mouseup select nodes and edges within the rectangle? Like on a Windows desktop or within file explorer? I envision the user would be required to press ctrl or something in order to override canvas dragging?

jgorene:
Of course, I also think that this would be an interesting feature :wink:
However, I'm just wondering if it's worth adding this feature given the work that needs to be done to manage all the impacts on the module as it was designed, it seems to me in any case.

That being said, I tried to see what we could already do with the existing features...
Take a look at these 2 examples to illustrate what we can already do with some existing network interaction options and the difficulties we may encounter, not to mention some unwanted (or not) "side effects" that should not be ignored.

[...]

PR #4242, opened 2019-02-03
Selection box

softwareCobbler:
This implements a "selectionBox" feature in Network's interaction module. With this option enabled, the user can press CTRL and left click and then drag a box around the canvas. Let go and it selects the nodes and/or edges within the box (there are option flags the user can set to have it select nodes / edges).    
Open to feedback!  
This was briefly discussed in #4233.  
Thanks!
sbox


softwareCobbler:
[Quick video example of feature](https://vimeo.com/315103295)

jgorene:
A priori, I didn't encounter any bugs at least after a few tests.
Subject to further testing, this seems ready to be implemented at the latest on v4.22.
Good work !

softwareCobbler:
considering the line width option - since `CanvasRenderingContext2D.lineWidth` is in coordinate space, I'm not certain how to provide a sane default. The value the user provides will change in apparent thickness as the zoom scale changes. Recalculating upon camera scale changes is the obvious solution, but I don't know quite what units the user expects to see.  
Should a value of `1.0` should always be 1 screen pixel? Or 1% of the total canvas area?

jgorene:
@softwareCobbler Indeed and it's already very good like that, no problem :relieved:
I still have a lot to learn and it will take me a little more time to properly analyze the new option , anyway.

@mojoaxel Just by the way and nothing to do with this new interaction option but it seems that "zoomSpeed" option is obsolete... but maybe I missed something?

softwareCobbler:
Got the `lineWidth` option up and running for this:
+ It is a value in screen pixels, the canvas zoom should not affect the pixel count.  
+ `lineWidth` is kept greater than or equal to zero (negative values are forced to 0), and while the user can pass in real numbers, the passed-in values are rounded down to the nearest integer.  
+ a zero value means we draw no border, and just fill in the appropriate `fillRect`
+ canvas bounds clamping has been adjusted to deal with different sized `lineWidth`'s

lib/timeline/component/ItemSet.js
```diff
@@ -541,6 +541,7 @@ ItemSet.prototype.show = function() {
 
 /**
  * Activates the popup timer to show the given popup after a fixed time.
+ * @param {any} popup
```
mojoaxel: :+1:

softwareCobbler: cherry-picked this into #4256 to provide a more logical separation of concern (it's a bit out of place  in this PR)

mojoaxel:
@softwareCobbler Please consider moving this pull-request over to [vis-network](https://github.com/visjs/vis-network)! Also we are always looking for people that might be interesting in helping maintaining :wink:

softwareCobbler:
Haha thought it was DOA--when I can scrape together some time I'd be happy
to.

Thanks,
Dave R.

PR #4253, opened 2019-02-14
Fix for #3562

justinharrell:
Adds forEach to DataView, based on how map was implemented, just passes through but applies the dataview filter. See #3562

### DIFF
--- a/lib/DataView.js
+++ b/lib/DataView.js
@@ -240,7 +240,43 @@ DataView.prototype.getIds = function (options) {
 };
 
 /**
- * Map every item in the dataset.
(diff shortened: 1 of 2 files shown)


justinharrell:
Seems to be failing lint because of 9de41d6f916c49fa280939611f726a96e2461c4d

softwareCobbler:
> Seems to be failing lint because of [9de41d6](https://github.com/almende/vis/commit/9de41d6f916c49fa280939611f726a96e2461c4d)

true; easy fix, too: #4256

justinharrell:
Should be good to go now after rebase

PR #4274, opened 2019-03-15
Show tooltip on dot as well

kelvinau:
When `showTooltips` is enabled, items with titles will display a tooltip on mouseover. When there's no end time, a dot appears on the timeline. Use would intuitively look for the tooltip on the dot, but it is not shown when mouseover. This PR adds `timeline-item` to the dot DOM and binds `verticalBackground` with the mouse events, so the mouseover logic also applies to the dot.

Below is the effect after the changes:
Screenshot from 2019-03-15 14-03-08


yotamberk:
The vis.js project is no longer being maintained. if You'd like this feature to be added, please reapply this PR to https://github.com/yotamberk/timeline-plus and I'll review it there.

PR #4285, opened 2019-04-17
Added `pointer-events: none` for read-only timeline items

e1ectr0cut1e:
Without this, read-only items accept mouse events. If an item is dynamically removed on horizontal drag, dragging stops and the crossed arrows cursor remains.


yotamberk:
Please re-apply this in https://github.com/yotamberk/timeline-plus to be merged

